PR #71, opened 2019-12-08
Fix typo and format/status in API reference

curquiza:


references/search.md
```diff
@@ -39,20 +39,41 @@ curl \
...
+        "release_date": "1989-02-24"
+      }
+    },
+    ...
+  ],
+  "offset": 0,
+  "limit": 20,
+  "processingTimeMs": 32,
+  "query": "american ninja 5"
+}
```
bidoubiwa: Is the `_formatted` necessary to add? This subchild should only be present in the case of a formatting parameter in the query.  It's a bug that should be fixed by meiliSearch. In either case, I can see why you want to report the exact truth in the response example but don't you think this massive response is confusing and not readable?

references/settings.md
```diff
@@ -4,14 +4,14 @@
 
 <RouteHighlighter method="GET" route="/indexes/:uid/settings" />
 
-Get the index settings.
+Get settings for a given index.
```
bidoubiwa: I think it should be more straight to the point. Not sure 
```suggestion
Get settings of an index
```

references/synonyms.md
```diff
@@ -1,17 +1,17 @@
 # Synonyms
 
-## List one synonym
+## List synonyms of one sequence
```
bidoubiwa: Should sequence become a word in our terminology? I used it in the documentation because I liked it, but maybe `string` could be better.

references/search.md
curquiza: I didn't know for the `_formatted` bug, I will remove it.
For the readable point, I wanted to display the **real** response returns by MeiliSearch and not a kind of one, because I know it confuses me when I read some docs. It gives me the feeling it's not up-to-date, and sometimes, if there are other points unclear, I feel lost. That's why, For me, it's important to display the `hits` and all the other fields.
But if others do not agree with me, I take your side of course :)

references/settings.md
curquiza: I choose this kind of sentence ('for a given index') because it's used all along our documentation ^^'
But I agree.

references/synonyms.md
curquiza: I read it because I needed to understand synonyms and I did not understand 😂 That's why I changed. For me, we don't *list* one thing. Really confusing in a title for me.
But we have to discuss because I'm not sure at all what is a sequence, even after reading the docs.

PR #89, opened 2019-12-11
Update settings in API references

curquiza:


references/settings.md
```diff
@@ -1,6 +1,6 @@
 # Settings
 
-## Get the index settings
+## Get index settings
```
Kerollmops:
```suggestion
## Get the settings related to an index
```

references/settings.md
```diff
@@ -44,11 +44,11 @@ List the settings.
 }
 ```
 
-## Add or update settings
+## Add or replace index settings
```
Kerollmops:
```suggestion
## Add or replace the settings related to an index
```

references/settings.md
```diff
@@ -78,17 +78,19 @@ An objet containing document attributes as keys and  `asc` ascending or `dsc` de
...
+None of the 3 settings parameters are mandatory
 :::
 
-### Example
+### Examples
+
+#### Add settings
 
 ```bash
 curl \
```
Kerollmops: Why is there `"distinctField": "",` ? That's probably invalid.

references/settings.md
```diff
@@ -110,6 +112,22 @@ curl \
...
+  --request GET 'http://localhost:7700/indexes/12345678/settings' \
+  --data '{
+  "rankingOrder": null,
+  "distinctField": null,
+  "rankingRules": null
+}'
+```
+
+::: danger
+You must set the fields to `null` and not to empty.</br>
```
Kerollmops:
```suggestion
You must set the fields to `null` to reset them and not to the empty value.</br>
```

references/settings.md
curquiza: Isn't that too long for a title (visible in the sidebar)?
Plus, for all the routes related to an index (documents, stop-word, synonyms, updates) we hadn't put that precision in the titles.
To be consistent, I can remove it from the title if you prefer? :)

references/settings.md
curquiza: I don't know, it was there before.
I don't know what to write, to be honest, I have never used this setting before, and the current example does not help me ^^
Can you give me an example to replace that? :)

curquiza: (if by invalid you think it will not be processed, it is)

references/settings.md
bidoubiwa:
```suggestion
## Get settings of an index
```

references/settings.md
bidoubiwa:
```suggestion
## Add or replace settings of an index
```

Kerollmops: It's ok to keep the title you wrote, I forget that it was visible in the sidebar.

references/settings.md
Kerollmops: By invalid I mean that it the `"distinctField"` paramter must not be set to empty but even `null` or not be specified at all.

references/settings.md
```diff
@@ -44,11 +44,11 @@ List the settings.
 }
 ```
 
-## Add or update settings
+## Add or replace settings of an index
```
Kerollmops:
```suggestion
## Add or replace the settings of an index
```

This is not a correct English sentence however, but I understand its too long for the sidebar.

references/settings.md
```diff
@@ -1,6 +1,6 @@
 # Settings
 
-## Get the index settings
+## Get settings of an index
```
Kerollmops:
```suggestion
## Get the settings of an index
```

references/settings.md
curquiza: Oula, I don't understand at all.
must not be set to empty but even null  -> by default the value of `distinctField` is `null`
or not be specified at all -> why can we change it in the body of the request then?

curquiza:
I change the titles by all your suggestions (@Kerollmops  and @bidoubiwa).
It creates a new line in the sidebar, quite ugly.
There are titles, why can't we do them short?

references/settings.md
Kerollmops: In this cURL example the `"DistinctField"` is set to empty (`""`), as yo wrote it in the paragraph above, setting it to empty is invalid, the engine will try to distinct documents on an empty, non-existent field which is invalid.

The API accept two things:
  - Set the `"DistinctField"` to `null` to reset its value to the default one
  - Set the `"DistinctField"` to a valid field, this way the engine will distinct documents on the specified existing field.
  - Don't write it in the body at all, this way the API will ignore and keep the current value associated to it.

Kerollmops:
We must write valid English short title, not just short titles.
I forgot that those titles were visible in the sidebar, keeping the first ones (i.e. "Get index settings") is probably the best solution at the moment.

references/settings.md
curquiza: Yes, I well understood that the current example is not a great and consistent example, and that's why I asked you first to give me a valid field (your second point), so that I can put it as example documentation :)

Kerollmops: Removing the `"distinctField"` from the body should do the job.

curquiza:
I haven't mentioned that I wanted to disregard English in our titles, I just want short and correct english titles.
Ok, I keep my first choice then.

Kerollmops:
Sorry for that, it's hard to see the results of the modification when you cannot see the front page live :)

PR #122, opened 2020-01-24
Fix some mini mistakes

### LINKED ISSUE #123: getting_started folder
What is the purpose of the "getting_started" folder? Not visible in the documentation via links, but still accessible with the url.
For example:
https://docs.meilisearch.com/getting_started/download.html

curquiza:
Fixes little mistakes.
And closes #123.


guides/README.md
```diff
@@ -17,7 +17,7 @@ Contents :
 
 First of all, you must have access to a running instance of MeiliSearch.
 
-There are [several download possibilities](/guides/advanced_guides/binary.md#download-and-launch).
+There are several download possibilities.
```
bidoubiwa: Did you remove it because I copy pasted the possibilities underneath it ?
Because I thought about it, and even if it is redundant, the objective of the getting started page is to link to the other pages of the documentation. In this context, i'd prefer to leave the link. What do you think ?

curquiza: I think there are already too much things to click everywhere. And when you click you have exactly the same tabs than below. Quite frustrating. It looks like we forgot to remove it.
But it's as you want. I really don't like "there are a lot of links throughout the docs" when I read documentation, but I'm not everybody :) I really thought you forgot to remove it.

curquiza:
@bidoubiwa Did you see my issue? #123 
It's not an emergency but if the answer is just "we forgot to remove this folder" I can do it in this PR :)

guides/README.md
bidoubiwa: I get your point but I would prefer keeping it. Thanks for the changes though !

curquiza:
In the getting started: the same words are linked to the same reference of the doc.
For example, in the "Add Documents" part, the word "document" is link three times to the same part of the doc.
It is not consistent. Either we link ALL the "document" words in this section, or we link one only. For me, the first one in the part is enough. The link on a word is really visible thanks to vuepress.

guides/advanced_guides/distinct.md
```diff
@@ -34,13 +34,13 @@ Lets use the following documents sample with 3 jacket of **different `colors`**
...
 
 ```bash
-curl
+ $ curl
   -X POST 'http://localhost:7700/indexes/jackets/settings' \
   --data '{ "distinctField": "skuid" }'
 ```
 
-With this setted, document request made to MeiliSearch **will never return two or more jacket with the same `skuid`**.
+With this settings, document request made to MeiliSearch **will never return two or more jackets with the same `skuid`**.
```
tpayet:
```suggestion
With this setting, search requests **will never return two or more jackets with the same `skuid`**.
```

PR #143, opened 2020-02-14
Schemaless

bidoubiwa:


references/documents.md
```diff
@@ -92,27 +86,26 @@ $ curl \
 
 <RouteHighlighter method="POST" route="/indexes/:uid/documents"/>
 
-Insert a list of documents or replace them if they already exist based on [their unique identifiers](/guides/main_concepts/indexes.md#schema-definition).
+Add a list of documents or replace them if they already exist.
+
+If you send an already existing document (same [identifier](/guides/main_concepts/documents.md#identifier)) **the whole old document**  will be override by the new document. Thus, removing any fields previously in the document not present in the new document.
```
curquiza:
```suggestion
If you send an already existing document (same [identifier](/guides/main_concepts/documents.md#identifier)) **the whole old document** will be overridden by the new document. Thus, removing any fields previously in the document not present in the new document.
```

references/documents.md
```diff
@@ -153,27 +146,25 @@ This `updateId` allows you to [track the current update](/references/updates.md)
 
 <RouteHighlighter method="PUT" route="/indexes/:uid/documents"/>
 
-Insert a list of documents or update them if they already exist based on [their unique identifiers](/guides/main_concepts/indexes.md#schema-definition).
+Add a list of documents or update them if they already.
```
curquiza:
```suggestion
Add a list of documents or update them if they already exist.
```

references/documents.md
```diff
@@ -153,27 +146,25 @@ This `updateId` allows you to [track the current update](/references/updates.md)
...
 
-In case of an update, the old document will be only partially updated according to the fields in the request body. It will not be overwritten entirely.</br>
-To completely overwrite a document, check out the [add and replace documents route](/references/documents.md#add-or-replace-documents).
+If you send an already existing document (same [identifier](/guides/main_concepts/documents.md#identifier)) **only the field of the new documents** will be changed in the old document. Thus, any fields not present in the new document remained unchanged.
```
curquiza:
```suggestion
If you send an already existing document (same [identifier](/guides/main_concepts/documents.md#identifier)) **only the fields of this new document** will be changed in the old document. Thus, any fields not present in the new document remain unchanged.
```

curquiza: I dont really understand the senteces. I guess that you try to describe the partial updating without mention it because I know MeiliSearch API (and because I can see it on the git diff).
I really think that the word "partial update" should stay, to be understood by most of the people. Anyone who used an API before know what a partial update is.
For those who don't know, indeed, it's ok to explain, and your explanation is welcome. But I would use more "field in the request body" than "fields of new document". I find it confusing.
Another point of view is needed :) @tpayet

references/documents.md
```diff
@@ -153,27 +146,25 @@ This `updateId` allows you to [track the current update](/references/updates.md)
...
+
+To completely overwrite a document, check out the [add and replace documents](/references/documents.md#add-or-replace-documents).
```
curquiza:
```suggestion
To completely overwrite a document, check out the [add or replace documents](/references/documents.md#add-or-replace-documents).
```

references/indexes.md
```diff
@@ -44,146 +51,107 @@ Get the index relative information.
...
 <RouteHighlighter method="POST" route="/indexes"/>
 
 Create an index.
 
-The [schema](/guides/main_concepts/indexes.md) definition is optionally send through the body.
-If no [schema](/guides/main_concepts/indexes.md) has been defined when the first document is sent it will be [infered based on that document](/guides/main_concepts/indexes.md#schema-definition).
-
-A randomly generated UID will be returned. It's associated to the new index. This UID will be essential to make all request over the created index.
-
+Takes as parameter a unique `uid` and **optionnaly** the name of the <glossary word="document identifier"/>: `identifier`.
```
curquiza:
```suggestion
This route takes as parameter an unique `uid` and **optionally** the name of the <glossary word="document identifier"/>: `identifier`.
```

references/indexes.md
```diff
@@ -44,146 +51,107 @@ Get the index relative information.
...
 
-The body take at least an index name.
 
-The body accepts an optional [schema](/guides/main_concepts/indexes.md) definition of your documents.
+| Variable  | Description           |
+|-----------|-----------------------|
+| **uid** | The index unique identifier |
+| **identifier** | the document <glossary word="attribute" /> of his unique identifier |
+
+The body take at least an index uid. The document identifier is optionnal.
```
curquiza:
```suggestion
The body takes at least an index uid. The document identifier is optional.
```

references/indexes.md
```diff
@@ -44,146 +51,107 @@ Get the index relative information.
...
 
 #### Body
 
-The body take at least an index name.
 
-The body accepts an optional [schema](/guides/main_concepts/indexes.md) definition of your documents.
+| Variable  | Description           |
+|-----------|-----------------------|
+| **uid** | The index unique identifier |
+| **identifier** | the document <glossary word="attribute" /> of his unique identifier |
```
curquiza:
```suggestion
| **identifier** | The document <glossary word="attribute" /> of its unique identifier |
```

curquiza: And if you chose to use the glossary, instead of the link, you might change it everywhere.

references/documents.md
bidoubiwa: Do you like this more : 
`
If you send an already existing document (same [id](/guides/main_concepts/documents.md#identifier)) the old document will be only partially updated according to the fields of the new document. Thus, any fields not present in the new document are kept and remained unchanged.
`

guides/main_concepts/indexes.md
```diff
@@ -1,153 +1,74 @@
...
+An index is defined by an `uid` and contains the following information:
+- One document identifier <glossary word="attribute"/>
+- A set of relevancy rules (based on presets and customization)
+- A list of Synonyms and stop-words
+- Rules for each field of a document
+- Settings
 
-## Index UID and name
+#### Example
+An index will typically be the collection of a certain type of data. For example, a `movie` index with documents containing each information about a movie. On the same server, you could have another index containing all the movie reviews, where each document contains information about a review.
```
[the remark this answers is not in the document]
bidoubiwa: Accepted but changed typically with usually

guides/main_concepts/indexes.md
```diff
@@ -1,153 +1,74 @@
...
 
-The `uid` of an index is its unique identifier. It is the `:uid` parameter on every `indexes/:uid` route.
+Each of the indexes will have information about the fields found in the documents. What should be done with each field, and their order of importance. Different synonyms, relevancy rules, stop-words, could be set on both indexes based on the context.
```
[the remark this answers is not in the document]
bidoubiwa: It means that the index has the information about all those things. I changed the sentence with this : 
``` 
Each of the indexes has information about the fields found in the documents, what should be done with each field, and their order of importance. An index also has its own synonyms, relevancy rules and stop-words. The information of one index never acts on another index.

This means, you can create synonyms for a `movie` index and different synonyms for a `clothes` index on the same MeiliSearch server.
```

references/documents.md
bidoubiwa: applied !

references/documents.md
bidoubiwa: changed with `overwritten`

bidoubiwa: Changed with `If you send an already existing document (same [id](/guides/main_concepts/documents.md#identifier)) the **whole existing document** will be overwritten by the new document. Fields previously in the document not present in the new document are removed.`

PR #196, opened 2020-03-14
Additional review

qdequele:


guides/advanced_guides/prefix.md
```diff
@@ -16,19 +16,18 @@ Given a set of words in a dataset:
...
-
+query: `s`:
+response:
 - `show`
 - `shine`
 
-but not
-
-- `movies`
-- `musical`
```
bidoubiwa: Why did you remove this?

bidoubiwa:
Everything else is great. Thanks a lot.

guides/advanced_guides/prefix.md
qdequele: Look at the entire file, this part seems obvious to me. What do you think about it?

bidoubiwa: I agree that from our perspective this is obvious. The reason why I put it is that you generally look at examples either to reassure your knowledge either because you did not understand after reading the theoretical part. So I thought by adding this very-obvious example people will 100% be sure to understand the idea.

PR #201, opened 2020-03-17
Review api reference

qdequele:


qdequele:
The second commit is just a proposal, tell me what you're thinking about.

bidoubiwa:
I need @curquiza opinion. Potentially it could be not compatible with her scrapper.

curquiza:
I don't know this kind of situation with the scrapper. But I would say, it should be ok. If I realize it's not, I will fix it soon :)

PR #203, opened 2020-03-17
Stop words wording

bidoubiwa:


references/stop_words.md
```diff
@@ -1,6 +1,6 @@
 # Stop-words
 
-Stop-words are a list of words that will be ignored in search queries. So if you add `the` as a stop word and you make a search on `the mask` you will only have matching documents with `mask`.
+Stop-words is a list of words that will be ignored in search queries. So if you add `the` as a stop word and you make a search on `the mask` you will only have matching documents with `mask`.
```
curquiza: Weird, even if for you it's clear, it's weird to read.
The reader does not know about our convention.
`Stop-words list is` or `Stop-words are`.

bidoubiwa:
```suggestion
The stop-words route lets you add a list of words that will be ignored in search queries. So if you add `the` as a stop word and you make a search on `the mask` you will only have matching documents with `mask`.
```

PR #220, opened 2020-03-24
WIP: Removed tooltip package to fix the building bug

bidoubiwa:
There was a conflict between the `vuepress-plugin-element-tabs` and `vue-directive-tools`. 

This conflict resulted in the getting_started guide to be not accessible via the documentation landing page.

By completely removing `vue-directive-tools` and replacing it with a custom made tooltip that uses the base package "popperJs" the problem is fixed.


.vuepress/components/glossary.vue
```diff
@@ -25,8 +34,74 @@ export default {
...
+            glossary,
+            content: ""
         }
+    },
+    created () {
+        this.content = glossary[this.word] + "<div id='arrow' data-popper-arrow></div>"
+    },
+    methods: {
+        showTooltip() {
+            console.log(" hello ");
```
curquiza:
```suggestion
```

.vuepress/components/glossary.vue
```diff
@@ -25,8 +34,74 @@ export default {
...
+        },
+        hideTooltip() {
+            this.$el.querySelector('.tooltip-content').classList.add("tooltip-hide");
+        }
+    },
+    mounted() {
+        const text = this.$el.querySelector('.tooltip-text');
+        const tooltip = this.$el.querySelector('.tooltip-content');
+
+        console.log(text, tooltip) // I'm text inside the component.
```
curquiza:
```suggestion
```

.vuepress/components/glossary.vue
```diff
@@ -25,8 +34,74 @@ export default {
...
+        createPopper(text, tooltip, {
+            placement: 'top',
+            modifiers: [
+                {
+                name: 'offset',
+                options: {
+                    offset: [0, 8],
+                },
```
curquiza: Bad indentation

.vuepress/components/glossary.vue
```diff
@@ -25,8 +34,74 @@ export default {
...
+        font-family: inherit !important;
+        background: rgba(0, 0, 0, 0.8);
+        color: white;
+        padding: 4px 8px;
+        font-size: 13px;
+        border-radius: 4px;
+        z-index: 100;
+        max-width: 500px;
+    }
+
```
curquiza:
```suggestion
```

bidoubiwa:
So the problem came from the component. It was considered as a SSR component, so it was build as one. I suppose this conflicted with the vuepress-plugin-element-tabs because the problem resolved itself when I removed that package.
The problem was fixed by adding <ClientOnly> around the glossary component which has custom `onmouse` over directives.
[More information here: ](https://vuepress.vuejs.org/guide/using-vue.html#browser-api-access-restrictions)

Kerollmops:
Isn't it possible to add the `<ClientOnly>` tag inside of the template of the component instead of duplicating this everywhere? It is highly possible that we forget to add that tag the next time we declare a tooltip.

bidoubiwa:
I tried to add it inside the component but then the DOM was considered undefined in my `mounted()` method. Since I need to reach the DOM on component creation to set up the tooltip and that it is the last method in the VUE component lifecycle (before : beforeUpdate and update) I have to do it in mounted().

So I thought of creating an in-between component while only `<clientOnly><glossary></ClientOnly>`. But because I mostly wanted the bug to be fixed online I preferred to try that tomorrow.

bidoubiwa:
@Kerollmops  #223

PR #216, opened 2020-03-24
Rename README.md into index.md to have a proper repository README

tpayet:
closes #188

PR #224, opened 2020-03-25
Adding linter and prettier fixes and tests

bidoubiwa:
Because of many styling mistakes and missing a common styling guide I added a linter and a styler.

A test was added that will check - without fixing - the style of any `.js` `.vue` and `.md` files.
- `js` and `vue` files are checked by `prettier` and `es-lint`
- `md` files are checked by `prettier` only

During contribution it will be expected from all contributors to use the following script before submitting a PR :
```bash
yarn style:fix
```
This script will fix all styling and linting errors that are fixable.
All warning and errors that this script will raise is expected to be fixed by the contributor.
This information will be added to the README.md on a later PR (because a PR is made on the same file for the moment #216 ). 

### Note
- Two `vue` files are being ignored voluntarily by the linter because they will be fixed by @curquiza.
- There is a problem with the URL `https://en.wikipedia.org/wiki/Atomicity_(database_systems)` in the documents guide. Prettier does not like the parentheses in the URL. This is why you will find a comment in front of it stating : `<!-- prettier-ignore -->` In order to tell prettier to ignore this link.
- (Also little styling update on the tooltip that had a bad color. )

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,14 +9,6 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
-[**.js]
-indent_style = space
-indent_size = 2
-
-[**.{css,scss}]
-indent_style = space
-indent_size = 2
-
-[**.html]
+[**.{js,vue,css,scss,html}]
 indent_style = space
 indent_size = 2
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,4 @@
+!.vuepress
+.vuepress/dist/
+.vuepress/theme/components/Navbar.vue
+.vuepress/theme/components/MeiliSearchBox.vue
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,4 @@
+node_modules/
+.yarn/
+.github/
+public/
--- a/.vuepress/components/glossary.vue
+++ b/.vuepress/components/glossary.vue
@@ -1,102 +1,119 @@
 <template>
-    <span>
-        <span
-            class="tooltip-text"
-            v-on:mouseover="showTooltip"
-            v-on:mouseleave="hideTooltip"
-        >{{ word }}</span>
-        <div class="tooltip-content" v-html="content"></div>
+  <span>
+    <span
+      class="tooltip-text"
+      @mouseover="showTooltip"
+      @mouseleave="hideTooltip"
+    >
+      {{ word }}
     </span>
+    <div class="tooltip-content" v-html="content" />
+  </span>
 </template>
 
 <script>
-import { createPopper } from '@popperjs/core';
+import { createPopper } from "@popperjs/core";
 
 const glossary = {
-    'field': "A field is composed of an <b>attribute</b> and its associated data. <br><br> Ex: <code>attribute: 'data'</code>",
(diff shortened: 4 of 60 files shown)


.editorconfig
```diff
@@ -13,6 +13,10 @@ insert_final_newline = true
 indent_style = space
 indent_size = 2
 
+[**.vue]
+indent_style = space
+indent_size = 2
+
```
qdequele: You can merge js rules and VueJS rules.

.prettierignore
```diff
@@ -0,0 +1,5 @@
+.vuepress/
```
qdequele: Do you ignore vuepress folder from prettier?

.vuepress/components/glossary.vue
```diff
@@ -1,102 +1,119 @@
...
-    'attribute': "The key associated to some data in a field. <br><br> Ex:  <code>title: 'batman'</code> <br> title is the attribute in this example.",
-    'ranking rules': "Rules that are used by MeiliSearch to determine the relevancy of a document. <br><br> For example, the number of typos or the number of times the matching query is found in a document.",
-    'primary key': 'The attribute in a document of its unique identifier field. <br><br> Used by MeiliSearch to store the document. <br><br> Example: `movie_id` is the primary key of a movie document.',
-    'schemaless': 'This means you don\'t need to define or describe the structure of your data before adding data to an index. <br><br> For example, SQL\'s tables need schemas but mongodb\'s collections does not require it.',
-    'searchable': 'The data is used to determine the relevancy of a document when doing a search query.',
-    'displayed': 'The field is present in the document returned upon search.'
-}
+  field:
+    "A field is composed of an <b>attribute</b> and its associated data. <br><br> Ex: <code>attribute: 'data'</code>",
```
qdequele: Why the attribute is on one line and the value on another? Is it the linter?

.vuepress/components/mermaid.vue
```diff
@@ -1,18 +1,18 @@
...
-    import("mermaid/dist/mermaid").then(m => {
+    import("mermaid/dist/mermaid").then((m) => {
       m.initialize({
-        startOnLoad: true
+        startOnLoad: true,
       });
       m.init();
     });
-  }
+  },
```
qdequele: What is this? is it normal? That's the solution you told us about for the infinite loading.

.prettierignore
bidoubiwa: Yes this is an error, it was ignored by eslint so it still fixed the style. I remove this.

.vuepress/components/mermaid.vue
bidoubiwa: No, this is an asynchronous import to get the mermaid package that generates schemas (like the one in the asynchronous guide). It is not new.

.vuepress/components/glossary.vue
bidoubiwa: Yes, it is the linter because of the length of the line 19.

.eslintignore
```diff
@@ -0,0 +1,5 @@
+!.vuepress
+.vuepress/dist/
+.vuepress/theme/components/Navbar.vue
+.vuepress/theme/components/MeiliSearchBox.vue
+
```
curquiza:
```suggestion
```

bidoubiwa: of course  😭

PR #240, opened 2020-03-27
Doc word cropping

MarinPostma:
documents the fact that word are now cropped to match word boundaries, as implemented in [this PR (543)](https://github.com/meilisearch/MeiliSearch/pull/543)


bidoubiwa:
Thank you for keeping the documentation up to date. Could you add this information here as well: 
`/guides/advanced_guides/search_parameters.html#crop-length`
And also, are those features included in the v0.9 release?

PR #263, opened 2020-04-03
Scraper GitHub action

curquiza:
I confirmed with @tpayet, the secrets have been added.


curquiza:
@bidoubiwa pretty sure it will be removed with the rebase. Wait! I'm doing it.

PR #276, opened 2020-04-06
Rephrase distinctAttribute

matboivin:


guides/advanced_guides/distinct.md
```diff
@@ -1,45 +1,50 @@
-# Distinct
+# Distinct attribute
```
bidoubiwa, line 1: As a general rule in the documentation we use Title Case for titles.
Here for me information about it: https://www.grammar-monster.com/lessons/capital_letters_title_case.htm
```suggestion
# Distinct Attribute
```

matboivin:
Thanks!

PR #292, opened 2020-04-10
Fix links for scraper

curquiza:


bidoubiwa:
There is another bad link here:
in `/references/search` on line 32
```
[our dedicated page](/guides/advanced_guides/filtering) (/Users/charlottevermandel/meili/documentation/references/search.md:32:76)
```

PR #303, opened 2020-04-15
Trailing slash checker in the paths of vuepress config file

bidoubiwa:
# Trailing slash checker

This PR will check for trailing slashes in the path found in `config.js`.

This will raise an error because in the path key, path must always finish with a trailing slash in order for the scrapper to work.
```javascript
{
    title: 'All Settings',
    path: '/references/settings',
}
```
The error will output like this:
Screenshot 2020-04-15 at 10 33 47

The output will inform you on which line and which column you have to add the missing slash.


.vuepress/config-path-checker/utils.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  isArray: function (element) {
+    if (Array.isArray(element)) return true
+    return false
+  },
```
qdequele, line 5: Why did you not use directly `Array.isArray()`? I understand the syntax sugar for the `isObject`, but less for `isArray`

bidoubiwa: I think I was annoyed by the length of `Array.isArray()`. But yes, it's unnecessary complexity.

.vuepress/config-path-checker/result-report.js
```diff
@@ -0,0 +1,107 @@
...
+    console.info(chalk.green('Checking passed\n'))
+  }
+}
+
+function getLineAndColumn(path) {
+  const found = configFile.reduce((acc, curr, index) => {
+    // regex to match word path and its associated path on the same line
+    const regex = new RegExp('path.*' + path + '[^/].*', 'g')
+    const found = curr.match(regex)
+    if (!(found == null)) {
```
curquiza, line 77: Is there another way to write this condition?
Like `if(found)`?

If I understand right, with `!(found == null)` (I tested it on the console):
- if `found` is `null` -> the condition is `false`, because `!true -> false`
- if `found` is not `null` -> the condition is `true`

With my condition:
- if `found` is `null` -> the condition is `false`
- if `found` is not `null` -> the condition is `true`

bidoubiwa: There is in fact other ways and yours is really good. 
The reason I wrote it like this is more for code comprehension than for code compactness.

The `str.match()` function will return a null if a match is not found
```javascript
const paragraph = 'The quick brown fox jumps over the lazy dog. It barked.';
const regex = /notinthestring/g;
const found = paragraph.match(regex);

console.log(found);
```
```bash
> null
```
So I wrote the if in a way that the reader of the code understand that match will return a `null`

curquiza: Ok, if it's for the code comprehension, I took me multiples minutes to understand the condition because of the combination of `null` + `!`.
But if you tell me that is a common way to do it a condition in JS, I'm ok with that and I hope I'll get used to it one day 🤞😂

PR #305, opened 2020-04-15
Clarify partial update of settings via POST request

eskombro:
It doesn't seem clear enough in the documentation that a `POST` request on `/indexes/:index_id/settings` lets the user **partially** update an index settings 

This paragraph proposes an alternative to using specific routes, which seems to be an architectural choice of MeiliSearch and should be documented :) Let me know what you think!

[to my understanding, usually a POST request requires the whole body and won't accept just a few parameters]

### DIFF
--- a/references/settings.md
+++ b/references/settings.md
@@ -96,7 +96,6 @@ Update the settings of an index.
 | **displayedAttributes**  | [Strings] | Fields displayed in the returned documents                                       | All attributes found in the documents                                                             |
 | **acceptNewFields**      | Boolean   | Defines whether new fields should be searchable and displayed or not             | `true`                                                                                            |
 
-
 ### Examples
 
 #### Add settings
--- a/references/settings.md
+++ b/references/settings.md
@@ -18,6 +18,8 @@ These are the reference pages for the dedicated routes:
 Updating the settings means overwriting the default settings of MeiliSearch. You can reset to default values using the `DELETE` routes.
 :::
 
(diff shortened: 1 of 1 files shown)


references/settings.md
```diff
@@ -18,6 +18,8 @@ These are the reference pages for the dedicated routes:
 Updating the settings means overwriting the default settings of MeiliSearch. You can reset to default values using the `DELETE` routes.
 :::
 
+It is also possible to partially update your settings using a POST request to `Settings`, containing only the parameters that you want to modify. Parameters that are not present in the request body will keep their current configuration and WONT be modified.
```
bidoubiwa, line 21:
```suggestion
Updates in the settings route are **partial**. This means that any parameters not provided in the body will be left unchanged.
```

bidoubiwa: Could you add this one after the explanation in ## Update settings

references/settings.md
```diff
@@ -94,8 +96,6 @@ Update the settings of an index.
 | **displayedAttributes**  | [Strings] | Fields displayed in the returned documents                                       | All attributes found in the documents                                                             |
 | **acceptNewFields**      | Boolean   | Defines whether new fields should be searchable and displayed or not             | `true`                                                                                            |
```
bidoubiwa, line 98:
```suggestion

Any parameters not provided will be left unchanged.

```

eskombro: Added the `warning`

bidoubiwa:
Could you add this text 
```
Updates in the settings route are **partial**. This means that any parameters not provided in the body will be left unchanged.
```
here: 
Screenshot 2020-04-16 at 09 21 40

eskombro:
Done!

PR #317, opened 2020-04-21
Use vuepress-plugin-meilisearch

curquiza:
Uses this fresh new and beautiful [vuepress-plugin-meilisearch](https://www.npmjs.com/package/vuepress-plugin-meilisearch) 🎉

Bonus: renames the config file for the scraper as you wanted @bidoubiwa 😬


tpayet:
Screenshot 2020-05-07 at 16 10 18
Neat!

.github/workflows/gh-pages-scraping.yml
```diff
@@ -40,7 +40,7 @@ jobs:
       env:
         HOST_URL: ${{ secrets.MEILISEARCH_HOST_URL }}
         API_KEY: ${{ secrets.MEILISEARCH_API_KEY }}
-        CONFIG_FILE_PATH: ${{ github.workspace }}/.vuepress/scraper/config.json
+        CONFIG_FILE_PATH: ${{ github.workspace }}/.vuepress/scraper/scraper.config.json
```
bidoubiwa, line 43: Is it possible to give it a name that is more linked to the project that uses it?
For example: 
`doc-scraper.config.json`
That way contributors will know that this is the config file for: `doc-scraper` and not any other possible scraper.

I know I was not as precise when I suggested scraper.config.json but after some observation config files tend to have their dependency in their name.
```suggestion
        CONFIG_FILE_PATH: ${{ github.workspace }}/.vuepress/scraper/doc-scraper.config.json
```
ps: I cannot change the name of the file remotely otherwise I would have proposed it :(

curquiza: No worries, I will do it manually anyway because I would rather `docs-scraper` than `doc-scraper` 😉
Changing that!

curquiza: > That way contributors will know that this is the config file for: doc-scraper and not any other possible scraper.

In the same idea, I also renamed the folder. Tell me if you agree!

bidoubiwa: Perfect!

Issue #228, opened 2020-03-25
Table of contents for FAQ


curquiza:
Currently, the FAQ questions are visible on the left in the sidebar. They can be confused with titles and the users might not see them.

We should find a way to generate a table of contents at the beginning of the FAQ so that the user can select the question he/she wants.

Issue #227, opened 2020-03-25
FAQ button in home page


curquiza:
Find a better place for the FAQ button on the home page. Or a better way to display it.
(I have no solution right now).

Warning: the Getting Started button should stay more visible than the FAQ one.

Issue #218, opened 2020-03-24
Move tuturials menu into resources menu


bidoubiwa:
Move all tutorials and its sub-menus into the resources menu.

PR #321, opened 2020-04-23
Add FAQ table of content

matboivin:
Closes: #228

Closes: #218
Closes: #227

I had to rename `faq.md` into `README.md` (path : `faq/README.md`).


matboivin:
I wanted to show you I moved the FAQ in PR #323 and it impacts the display. It would look like this without any TOC:

move

(Screenshot of preview from #323)

bidoubiwa:
I find it still hard to read. Those are long questions, which makes you less inclined to read them, but even more when they are packed like that.

Issue #290, opened 2020-04-10
Add better example in the search parameters for the new filters in v0.10


bidoubiwa:
The new version of MeiliSearch brings a new guide on filters and thus changes to the search parameters guide.

This example is missing some deepness and should be more precise:

bidoubiwa:
+ Move the filtering menu title under Search parameters

bidoubiwa:
And typo on Jordan Peele

PR #324, opened 2020-04-23
Filtering examples

matboivin:
Closes: #290


guides/advanced_guides/filtering.md
```diff
@@ -57,7 +57,11 @@ An arbitrary number of expressions can be connected together thanks to logical c
 - `AND` performs a logical 'and' between two expressions, _e.g._ `title = Dumbo AND title = "Tim Burton"`
 
 ::: note
-`NOT` has the highest precedence, this means that `NOT title = Scream OR title = "The Avengers"` is effectively evaluated `(NOT title = Scream) OR title = "The Avengers"`. `AND` precedence is lower than `NOT` and higher than `OR`.
+`NOT` has the highest precedence.<br/>
+`NOT` precedence is higher than `AND`.<br/>
+`AND` precedence is higher than `OR`.<br/>
+`OR` has the lowest precedence.<br/>
+This means that `NOT title = Scream OR title = "The Avengers"` is effectively evaluated `(NOT title = Scream) OR title = "The Avengers"`.
```
MarinPostma, line 64: `NOT` is a unary operator, which is kind of a specific case, so I'd add an example with a binary operator too, for the sake of completion
```suggestion
`NOT` has the highest precedence.<br/>
`NOT` precedence is higher than `AND`.<br/>
`AND` precedence is higher than `OR`.<br/>
`OR` has the lowest precedence.<br/>
This means that `NOT title = Scream OR title = "The Avengers"` is effectively evaluated `(NOT title = Scream) OR title = "The Avengers"`, and `title = Scream OR title = "The Avengers" AND release_date > 795484800` is effectively evaluated `title = Scream OR (title = "The Avengers" AND release_date > 795484800)`
```

guides/advanced_guides/filtering.md
```diff
@@ -122,14 +126,50 @@ $ curl --get 'http://localhost:7700/indexes/movies/search' \
     --data-urlencode 'filters=release_date > 795484800'
 ```
 
-Now imagine that we want only the movies released after the 18 of march 1995, and directed by either Jordan Peel or Tim Burton, then you would use this filter:
+Now imagine that we want only the Batman movies released after the 18 of march 1995, and directed by either Tim Burton or Christopher Nolan, then you would use this filter:
```
MarinPostma, line 129:
```suggestion
Now imagine that we only want Batman movies released after the 18 of march 1995, and directed by either Tim Burton or Christopher Nolan, then you would use this filter:
```

guides/advanced_guides/filtering.md
```diff
@@ -122,14 +126,50 @@ $ curl --get 'http://localhost:7700/indexes/movies/search' \
...
+
+```bash
+$ curl --get 'http://localhost:7700/indexes/movies/search' \
+    --data-urlencode 'q=Batman' \
+    --data-urlencode 'filters=release_date > 795484800 AND (director = "Tim Burton" OR director = "Christopher Nolan"'
 ```
 
 Note that filtering on string is case insensitive. Here, the parentheses are mandatory, as `AND` has a higher precedence.
 
+If we want to select only horror movies of our favorite director:
```
MarinPostma, line 145:
```suggestion
If we want to select only horror movies from our favorite director:
```

guides/advanced_guides/filtering.md
```diff
@@ -122,14 +126,50 @@ $ curl --get 'http://localhost:7700/indexes/movies/search' \
...
+
+Querying on `horror`, the above example results in the following CURL command:
+
+```bash
+$ curl --get 'http://localhost:7700/indexes/movies/search' \
+    --data-urlencode 'q=horror' \
+    --data-urlencode 'filters=director = "Jordan Peele"'
+```
+
+If we want to find Planet of the Apes movies which have been well rated but we want to ignore the one directed by Tim Burton:
```
MarinPostma, line 159:
```suggestion
If we want to find the `Planet of the Apes` movies that have been well rated but we want to ignore the one directed by Tim Burton:
```

guides/advanced_guides/filtering.md
matboivin: Thanks a lot!

Issue #291, opened 2020-04-10
Change example of attributes to crop in search parameters guide


bidoubiwa:
For the moment the example is not easy to read and does not explain much about the explained search parameter : `Croplength` and `attributesToCrop`
```json
{
  "id": "50393",
  "title": "Kung Fu Panda Holiday",
  "poster": "https://image.tmdb.org/t/p/w1280/gp18R42TbSUlw9VnXFqyecm52lq.jpg",
  "overview": "The Winter Feast is Po's favorite holiday. Every year he and his father hang decorations, cook together, and serve noodle soup to the villagers. But this year Shifu informs Po that as Dragon Warrior, it is his duty to host the formal Winter Feast at the Jade Palace. Po is caught between his obligations as the Dragon Warrior and his family traditions: between Shifu and Mr. Ping.",
  "release_date": 1290729600,
  "_formatted": {
    "id": "50393",
    "title": "Kung Fu Panda Holiday",
    "poster": "https://image.tmdb.org/t/p/w1280/gp18R42TbSUlw9VnXFqyecm52lq.jpg",
    "overview": "d his father hang decorations, cook together, and serve noodle soup to the villagers. But this year Shifu informs Po that as Dragon Warrior, it is his duty to host the formal Winter Feast at the Jade ",
    "release_date": 1290729600
  }
}
```

An example should be made on a smaller cropLength, like 5, so that it can be explanatory at sight.

PR #333, opened 2020-04-27
Update cropLength example

matboivin:
Closes:  #291

If put 10 instead of 5 to display a word after the matching term. When cropLength is set to 5, we have:
```
"overview": "year Shifu",
```

There is no word after Shifu since Shifu is 5 characters long. I thought it was less demonstrative. What do you think?


guides/advanced_guides/search_parameters.md
```diff
@@ -139,7 +139,7 @@ You will get the following response with the **cropped version in the \_formatte
     "id": "50393",
     "title": "Kung Fu Panda Holiday",
     "poster": "https://image.tmdb.org/t/p/w1280/gp18R42TbSUlw9VnXFqyecm52lq.jpg",
-    "overview": "his father hang decorations, cook together, and serve noodle soup to the villagers. But this year Shifu informs Po that as Dragon Warrior, it is his duty to host the formal Winter Feast at the Jade",
+    "overview": "this year Shifu info",
```
bidoubiwa, line 142: I tried the command and this was my result
```suggestion
    "overview": "this year Shifu informs",
```

matboivin: Thank you!

PR #341, opened 2020-04-30
Markdown extention

bidoubiwa:
In this PR

## Remove prettier added markdownlint

Until now prettier plugin for markdown was used to lint and prettify our markdown files. This was unsatisfactory since no errors hints were given on a style error. 
All errors were fixed using the `yarn style:fix` script. 

This has been replaced with [markdownlint](https://github.com/DavidAnson/markdownlint) that has 1.9k stars and +11 contributors.

Some rules were removed because they do not apply to our use case: 
- `MD013`:  line-length Line length [Expected: 80] 
- `MD001`:  heading-increment/header-increment Heading levels should only increment by one level at a time
- `MD026`:  no-trailing-punctuation Trailing punctuation in heading 
- `MD014`:  commands-show-output Dollar signs used before commands without showing output
- `MD024`: no-duplicate-heading/no-duplicate-header
- `MD033`: no-inline-html Inline HTML

MD024 should be addressed but no easy solution exists for the moment.

[Here is the list of all rules](https://github.com/DavidAnson/markdownlint#rules--aliases)

## Added customization of our markdown parser

Two parser plugin were added to our markdown parser:
```
breaks:       true,        // Convert '\n' in paragraphs into <br>
linkify:      true,        // Autoconvert URL-like text to links
```
Since we write a lot of doc we need the possibility of doing a new line without having to right `<br>`.

## Remove quickstart guide
This guide is a smaller version than the getting started guide that is not kept updated properly and that conflict with the getting started guide in the search bar.


package.json
```diff
@@ -11,13 +11,14 @@
     "dev": "vuepress dev",
     "check-links": "yarn check-md && yarn check-config-paths",
     "check-md": "vuepress check-md .",
-    "check-style": "yarn lint && yarn prettier-md-check",
+    "check-style": "yarn style",
```
curquiza, line 14: Why does not use "style" directly? (We need to change the README and the CI)

Or simply removing `style` rule and just having:
```
"check-style": "yarn lint && yarn marklint"
```

bidoubiwa: Yes this is only to match the syntax of `check-md`. 
I thought about removing it but I was not sure. If you don't like it either I remove it

PR #347, opened 2020-05-05
Specify values must be numbers

matboivin:
Apply an ascending or descending sorting rule on a given attribute whose values must be numbers.

### DIFF
--- a/guides/main_concepts/relevancy.md
+++ b/guides/main_concepts/relevancy.md
@@ -116,7 +116,7 @@ Depending on your needs, you might want to change this order of importance. To d
 
 New rules can be added to the existing list at any time and placed anywhere in the sequence.
 
-A custom rule allows you to create an ascending or descending sorting rule on a given attribute. The attribute **must** have a numeric value in the documents. If any value is not a numeric type, the sorting rule won't be applied. Only numbers can be arranged in ascending or descending order.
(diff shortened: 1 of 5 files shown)


guides/main_concepts/relevancy.md
```diff
@@ -126,6 +126,12 @@ To add your own ranking rule, you have to communicate either `asc` for ascending
 
 Add this rule to the existing list of ranking rules using the [settings route](/references/ranking_rules.md#update-ranking-rules).
 
+::: warning
+
+If you apply an ascending or descending sorting rule on an attribute, the values of the latter **must be numbers**. If any value is a string, it won't be converted into a numeric type.
```
bidoubiwa, line 131: `it won't be converted into a numeric type.` What does this implies? This is confusing. 
` If any value is a string, it won't be converted into a numeric type.` If it is a string, but what if it is a boolean ? What if it is a null type? Do you mean `not a number` ?

bidoubiwa:
### Add the information in the settings guide

https://docs.meilisearch.com/guides/advanced_guides/settings.html#ranking-rules

Warning should not be used if we can just add the information in plain text.

Could you not add it using a warning ? For example: 

```
### Custom ranking rule

You can add a custom ranking rule anywhere in the list of ranking rules. A custom ranking rule is composed of an attribute and an ascending or descending order. The attribute **must** have a numeric value in the documents.
```

matboivin:
Yes, you're right, I am going to remove the warning. I will also update the settings page.

Issue #330, opened 2020-04-24
Improve the Asynchronous Update page


Kerollmops:
Could we add a note about the fact that the user updates are deleted after they are processed, the only thing that is kept is the number of documents that were processed and the update id.

https://meilicommunity.slack.com/archives/CP9DVS1RQ/p1587730019257300?thread_ts=1587723005.255800&cid=CP9DVS1RQ

PR #351, opened 2020-05-07
Add deletion information

matboivin:
Closes: #330

I added the information under `Async flow`. They may belong somewhere else. Let me know.


guides/advanced_guides/asynchronous_updates.md
```diff
@@ -9,6 +9,7 @@ Some operations are put in a queue and will be executed in turn (asynchronously)
 - When making a write request (_create/update/delete_) against the search engine, it stores the operation received in a queue and returns an `updateId`. With this id, the operation update is trackable.
 - Each update received is treated following the order it has been received.
 - You can get the update status on the [`/updates`](/references/updates.md) route.
+- Once an update has been processed, its user content is removed. The only data that is kept includes the number of documents that were processed and the `updateId`.
```
bidoubiwa, line 12: What do you mean by `user` content?

matboivin: The content related to the user's update but your question means I should rephrase it and my explanations weren't clear enough! Let me know if you have any suggestion.

curquiza: I bugged on `user content` too 😉 But sorry I have no suggestion to improve that...
(I don't know if saying "I bugged" is a real word, sorry 😇)

matboivin: I like "bugged". ^^
I post you 2 screenshots. I understood my explanation was totally wrong. They are not really deleted but kept and marked as processed.

before

After:
after

matboivin:
Also added an "s" to update, here: https://deploy-preview-351--dreamy-wiles-8fd85b.netlify.app/references/updates.html

guides/advanced_guides/asynchronous_updates.md
bidoubiwa: I think you did not bug but maybe this example does not showcase the problem. Maybe @Kerollmops could enlighten us.

Kerollmops: The data that was sent by the user, the update content, is kept until it is processed by the database (it is asynchronous). Once processed the only information that is kept by the database is, for example, the number of documents that were processed in this update.

PR #366, opened 2020-05-14
Removed bad public dir and changed all links

bidoubiwa:
There were two public folders. 

One at the root of the documentation, on a the root of the .vuepress folder.
The one at the root of the documentation is not recognized by vuepress as a public folder. Thus any image that was stored inside that directory is accessed using the complete relative path. Which was often something like this:

```
../../public/image
```

I have moved them to the right `public` folder and changed all links.


guides/main_concepts/relevancy.md
```diff
@@ -46,7 +46,7 @@ Results are sorted by **the similarity of the matched words with the query words
 
 ::: tab Typo
 
-![Image from alias](../../public/image/vogli3.png)
+![Image from alias](/image/vogli3.png)
```
curquiza, line 49: Should we keep the name `image` since we almost only have images in `.vuepress/public`?

eskombro: I think it makes sense. We may have something different than images at some point, and also it makes really clear what's the content of the folder, so why not?

bidoubiwa: I will change the name of the folder to something more explanatory

eskombro: I think it makes sense. We may have something different than images, and also it makes really clear what's the content of the folder, so I would be ok with leaving it there

bidoubiwa: So i have changed image to ranking-rules and react-guide and put the appropriate images inside the appropriate folders

eskombro: No that I want to add to your todo list, but what if we do a `references` directoy, a `guides` dir and a `ressources` dir, reflecting the docs structure. (This was inspired by a conversation with @curquiza but I'm not sure thats exactly what she has in mind)

bidoubiwa: This could be done in the long run but I don't see the necessity right now

PR #396, opened 2020-06-08
Settings docs scraper

curquiza:
The new [release](https://github.com/meilisearch/docs-scraper/releases/tag/v0.9.1) and [Docker image](https://hub.docker.com/repository/docker/getmeili/docs-scraper) of `docs-scraper` contain the customization of the settings. This feature allows us to add synonyms like the one I had:
`relevant <-> relevancy`

Tell me @bidoubiwa and @erlend-sh if you need more!


erlend-sh:
Should we have one for `relevance` as well? Or does `relevancy` cover it since it's just one letter difference?

curquiza:
> Should we have one for relevance as well? Or does relevancy cover it since it's just one letter difference?

I would say it's enough. Maybe @Kerollmops can confirm :)

Kerollmops:
If you set a synonym the two words are considered equal (if this is a one word to one word synonym) which is not the same as doing a typo. I would recommend setting a synonym if you want documents containing both words to be considered equal. :)

curquiza:
@bidoubiwa Changed!

PR #401, opened 2020-06-10
New fields and Known fields

bidoubiwa:
Introduces the difference between new fields and known fields and how they will behave.

Based on this issue: meilisearch/MeiliSearch#761

### DIFF
--- a/guides/advanced_guides/field_properties.md
+++ b/guides/advanced_guides/field_properties.md
@@ -60,7 +60,7 @@ For example:
 
 1. Add `size` to the `attributesForFacetting` setting before adding documents.
 
-2. Add documents containing the `size` field
+2. Add documents containing the `size` field.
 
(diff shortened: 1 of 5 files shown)


guides/advanced_guides/field_properties.md
```diff
@@ -36,6 +36,56 @@ Suppose you manage a database that contains information about movies. By adding
...
+
+A **new field** is a field that is not yet known by MeiliSearch. As soon as a field is present in the settings, it becomes a **known field**.
+
+Depending on its state, a different behavior will be applied.
+
+### New Fields Behavior
+
+A new field is a field that is **present in a new document and not present in the settings**. To know if a field is in the setting, you can [get the settings list](/references/settings.md#get-settings) to ensure the presence of your field.
+
+A new field is automatically added to the `searchableAttributes` and `displayedAttributes` lists.
```
Kerollmops, line 49: Only if the `acceptNewFields` flags is set to true (which is the default).

PR #422, opened 2020-06-22
uid is not needed

Nothing-Works:
`uid` is not needed and adding uid will cause `UnhandledPromiseRejectionWarning`.

error detail

`UnhandledPromiseRejectionWarning: MeiliSearchApiError: Invalid JSON: invalid type: map, expected a string at line 1 column 7`

### DIFF
--- a/guides/introduction/quick_start_guide.md
+++ b/guides/introduction/quick_start_guide.md
@@ -135,7 +135,7 @@ const MeiliSearch = require('meilisearch')
 
 var client = new MeiliSearch({ host: 'http://127.0.0.1:7700' })
 const index = client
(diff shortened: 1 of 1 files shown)


Nothing-Works:
Hi @curquiza does someone still need to review it?

curquiza:
More than reviewing, someone needs to merge it 😂 I don't have the permission to do that
Be sure that @bidoubiwa or @tpayet will do it soon 😉 They are reactive!

Nothing-Works:
@curquiza All good! thanks for the info.

bidoubiwa:
Thanks a lot!

PR #424, opened 2020-06-23
should fix getting started

bidoubiwa:
A bug only present on master crashes when going directly to the getting_started guide.
-> [https://docs.meilisearch.com/guides/introduction/quick_start_guide.html](https://docs.meilisearch.com/guides/introduction/quick_start_guide.html)

I could not reproduce this neither on local neither on netlify. I suppose it comes from the latest plugin which is the `lazy load`.

Only indication I have is `warning An error was encountered in plugin "img-lazy"` on build. 
I suppose this is at the origin of the bug. I will investigate more later but fixing the doc is the most important.

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -143,7 +143,6 @@ module.exports = {
     [require('./config-path-checker')],
     [require('./custom-markdown-rules')],
     [require('./code-samples')],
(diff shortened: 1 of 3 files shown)


eskombro:
Thanks!

PR #427, opened 2020-06-25
Refactor getting started and fix bug

bidoubiwa:
- Removed installation from getting started (buggy)
- Removed hard-coded examples 
- Replaced them with code-sample plugin


curquiza:
Should I conclude we need to fill the ruby, PHP and go docs soon, at least for the getting started?

PR #433, opened 2020-06-29
Check if code sample is not empty

bidoubiwa:


tpayet:
Should we try it after merging [this PR](https://github.com/meilisearch/meilisearch-ruby/pull/61)

curquiza:
What will be the displayed result? 🙂An empty tab with no code inside, or a no tab at all?

bidoubiwa:
No tab at all. 
At first it was an empty tab but that is deceiving.

PR #434, opened 2020-06-29
Add limitations guide

### LINKED ISSUE #327: Create a page: MeiliSearch limits
Limites are:

- 10Mb of body JSON on AddDocuments and UpdateDocuments.
- 1000 words per attribute max indexed.
- 65536 words per documents
- 65536 max fields given to MeiliSearch during all life of the Index.
- ?? indexes
- 100Gb Database size (Could be increased)

-> Redirect to "Best practices of indexing"

@Kerollmops could you confirm this information and tell us if there are more restrictions?

tpayet:
closes #327

### DIFF
--- a/guides/advanced_guides/known_limitations.md
+++ b/guides/advanced_guides/known_limitations.md
@@ -1,0 +1,30 @@
+# Known limitations
+
+MeiliSearch has known limitations, some of which may be addressed in future versions.
+
+Some of these limitations have been made by MeiliSearch developers for relevacy and/or performance while others limitations are enforced by LMDB, the key-value store used by MeiliSearch.
+
+## Design limitations
+
+### Database size
+
+MeiliSearch uses two databases: the first one for storage and the second one for updates.
+On launch LMDB needs to know the size that it can allocate on disk. This space will be reserved on disk for LMDB, thus MeiliSearch. This space will also be allocated as virtual memory.
+
+The maxime database size is by default __100GiB__ for each databases. This size can me modified using the options `--max-mdb-size` & `--max-udb-size` as described in the [configuration guide](/guides/advanced_guides/configuration.md#max-mdb-size).
+
+### Number of indexes
+
+You can create __up to 200 indexes__ in MeiliSearch. This limit has been hard set for performance reasons.
+
+### Maximum words per attribute
+
+A maximum of __1000 words par attribute__ can be indexed. This means that if an attribute contains more than 1000 words, only the first 1000 words will be indexed and the rest will be silently ignored.
+
+This limit is enforced for relevancy reasons. The more words there are in a given attribute, the less relevant the search queries will be.
+
+## Other limitations
+
+### Payload size
+
+The default limit for the payload size is around __10MB__. [This limit can be modified](/guides/advanced_guides/configuration.md#payload-limit-size).
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -63,6 +63,7 @@ module.exports = {
             '/guides/advanced_guides/prefix',
             '/guides/advanced_guides/concat',
             '/guides/advanced_guides/bucket_sort',
+            '/guides/advanced_guides/known_limitations',
           ],
         },
       ],


tpayet:
@Kerollmops can you remind me why we are limited in number of indexes & maximum words per attribute?

Kerollmops:
> [...] can you remind me why we are limited in number of indexes [...]

This is because when we open LMDB we must specify the maximum number of databases (named key-value stores in a database), and it is good to keep this number low for performance reason.

> [...] maximum words per attribute?

This is not for technical but relevancy reasons, the more you have words in a document the more often you will return it in the results. It is better to keep documents small and accurate on the subject they describe to help MeiliSearch in its relevancy.

qdequele:
It could be great if this one could be finished soon. @tpayet @bidoubiwa

tpayet:
@MarinPostma for context: I started to write this doc a while ago and I am back on it since today. I request your review because I would like your input on what I wrote and if it's still relevant. I miss some information about these limitations so, if you can add more context I would be glad :)

guides/advanced_guides/known_limitations.md
```diff
@@ -0,0 +1,26 @@
...
+
+### Database size
+
+Default database maximum size is 100GiB.
+MeiliSearch uses LMDB under the hood as the storage engine. On launch, LMDB need to know the size that it can allocate on disk. This space will be reserved on disk for LMDB, thus MeiliSearch. This space will also be allocated as virtual memory, which is different from the real memory. You can read more [about storage](/resources/about_storage.md).
+You can modify this default limit using the options `--max-mdb-size` & `--max-udb-size` as described in the [configuration guide](/guides/advanced_guides/configuration.md#max-mdb-size).
+
+### Number of indexes
+
+You can create up to 200 indexes in MeiliSearch. When using LMDB, MeiliSearch have to specify a number of databases that must stay low for performance reasons.
```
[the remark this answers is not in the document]
Kerollmops: The total number of LMDB databases as been hard set to 3000 and an index takes something like 11 databases (IIRC) so `3000 / 11 ~= 272`.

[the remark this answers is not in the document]
tpayet: Can one of you confirm that we are using 11 databases in LMDB for one index?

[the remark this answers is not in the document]
Kerollmops: [It does seems like we use 11 databases](https://github.com/meilisearch/MeiliSearch/blob/cb267b68edc2481a1b4985cbd2277cb6cab3b6c9/meilisearch-core/src/store/mod.rs#L210-L221), indeed.

guides/advanced_guides/known_limitations.md
```diff
@@ -0,0 +1,27 @@
...
+### Database size
+
+MeiliSearch uses two database: one for storage and the second one for updates.
+Default database maximum size is 100GiB.
+MeiliSearch uses LMDB under the hood as the storage engine. On launch, LMDB need to know the size that it can allocate on disk. This space will be reserved on disk for LMDB, thus MeiliSearch. This space will also be allocated as virtual memory, which is different from the real memory. You can read more [about storage](/resources/about_storage.md).
+You can modify this default limit using the options `--max-mdb-size` & `--max-udb-size` as described in the [configuration guide](/guides/advanced_guides/configuration.md#max-mdb-size).
+
+### Number of indexes
+
+You can create up to 200 indexes in MeiliSearch. When using LMDB, MeiliSearch have to specify a number of databases that must stay low for performance reasons.
```
bidoubiwa, line 16:
```suggestion
You can create up to 200 indexes in MeiliSearch. When using LMDB, MeiliSearch has to specify a number of databases that must stay low for performance reasons.
```

bidoubiwa: >  When using LMDB, MeiliSearch has to specify a number of databases that must stay low for performance reasons. 

I do not understand this sentence :s

guides/advanced_guides/known_limitations.md
```diff
@@ -0,0 +1,27 @@
...
+
+### Maximum words per attribute
+
+A maximum of 1000 words par attribute can be indexed. This limit is enforced for relevancy reasons. The more words there are in a given attribute, the less relevant the search queries will be.
```
bidoubiwa, line 20: What do you mean by attribute ?
Screenshot 2020-08-18 at 11 04 15

Do you mean the `key` or the value of a field?

guides/advanced_guides/known_limitations.md
```diff
@@ -0,0 +1,27 @@
...
+If you put more than 1000 words in one attribute, only the first 1000 will be indexed.
+
+## Other limitations
+
+### Payload size
+
+The default limit for the payload size is around 10MB. [This limit can be modified](/guides/advanced_guides/configuration.md#options).
```
bidoubiwa, line 27:
```suggestion
The default limit for the payload size is around 10MB. [This limit can be modified](/guides/main_concepts/documents.html#upload).

When trying to upload a body that is more than 10MB it will raise an error unless you have raised the the limit.

\```bash
$ ./meilisearch --http-payload-size-limit=100000000
\```
> The payload limit is now +-100MB instead of 10MB
  
```

Backslashes should be removed but where in conflict with githubs markdown

guides/advanced_guides/known_limitations.md
```diff
@@ -0,0 +1,27 @@
+# Known limitations
+
+MeiliSearch has known limitations, some of which may be addressed in future versions
```
matboivin, line 3:
```suggestion
MeiliSearch has known limitations, some of which may be addressed in future versions.
```

guides/advanced_guides/known_limitations.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Design limitations
+
+### Database size
+
+MeiliSearch uses two database: one for storage and the second one for updates.
```
matboivin, line 9:
```suggestion
MeiliSearch uses two databases: the first one for storage and the second one for updates.
```

guides/advanced_guides/known_limitations.md
```diff
@@ -0,0 +1,27 @@
...
+Default database maximum size is 100GiB.
```
matboivin, line 10:
```suggestion
The maximum size of the database is by default 100GiB.
```

guides/advanced_guides/known_limitations.md
```diff
@@ -0,0 +1,27 @@
...
+MeiliSearch uses LMDB under the hood as the storage engine. On launch, LMDB need to know the size that it can allocate on disk. This space will be reserved on disk for LMDB, thus MeiliSearch. This space will also be allocated as virtual memory, which is different from the real memory. You can read more [about storage](/resources/about_storage.md).
```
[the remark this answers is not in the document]
matboivin:
```suggestion
MeiliSearch uses LMDB under the hood as the storage engine. On launch, LMDB needs to know the size that it can allocate on disk. This space will be reserved on disk for LMDB, thus MeiliSearch. This space will also be allocated as virtual memory, which is different from real memory. You can read more [about storage](/resources/about_storage.md).
```

guides/advanced_guides/known_limitations.md
matboivin: Me neither. Does it mean to impose a limit of X GiB on the size of these databases or on the number of indexes?

guides/advanced_guides/known_limitations.md
tpayet: I am not sure this is the place we want to explain MeiliSearch's behavior

guides/advanced_guides/known_limitations.md
tpayet: For clarity I think `1000 words per attribute` is more understandable. Or `1000 word per key`? I won't write `1000 words for a field's value` 😅
What do you think?

guides/advanced_guides/known_limitations.md
tpayet: I re-wrote the whole paragraph

guides/advanced_guides/known_limitations.md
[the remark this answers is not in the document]
tpayet: 👍

guides/advanced_guides/known_limitations.md
bidoubiwa: I would suggest we change this part with the following suggestion

```
A field can contain a maximum of 1000 words. If you put more than 1000 words in one attribute, only the first 1000 will be indexed.

This limit is enforced for relevancy reasons as the more words there are in a given attribute, the less relevant the search queries will be.
```

Also, is this only for relevancy reasons or also for perfomance reasons ?

guides/advanced_guides/known_limitations.md
```diff
@@ -0,0 +1,29 @@
...
+The maxime database size is by default __100GiB__ for each databases. This size can me modified using the options `--max-mdb-size` & `--max-udb-size` as described in the [configuration guide](/guides/advanced_guides/configuration.md#max-mdb-size).
+
+### Number of indexes
+
+You can create __up to 200 indexes__ in MeiliSearch. This limit has been hard set for performance reasons.
+
+### Maximum words per attribute
+
+A maximum of __1000 words par attribute__ can be indexed. This limit is enforced for relevancy reasons. The more words there are in a given attribute, the less relevant the search queries will be.
+If you put more than 1000 words in one attribute, only the first 1000 will be indexed.
```
bidoubiwa, line 23:
```suggestion
A maximum of __1000 words par attribute__ can be indexed. This means that if an attribute contains more than 1000 words, only the first 1000 words will be indexed and the rest will be silently ignored.

This limit is enforced for relevancy reasons. The more words there are in a given attribute, the less relevant the search queries will be.

```

bidoubiwa (changes requested):

guides/advanced_guides/known_limitations.md
bidoubiwa: I understand your point, I  would like @matboivin opinion on it. But, could you answer on the newer suggestion?

guides/advanced_guides/known_limitations.md
```diff
@@ -0,0 +1,31 @@
...
+
+The maxime database size is by default __100GiB__ for each databases. This size can me modified using the options `--max-mdb-size` & `--max-udb-size` as described in the [configuration guide](/guides/advanced_guides/configuration.md#max-mdb-size).
+
+### Number of indexes
+
+You can create __up to 200 indexes__ in MeiliSearch. This limit has been hard set for performance reasons.
+
+### Maximum words per attribute
+
+A maximum of __1000 words par attribute__ can be indexed. This means that if an attribute contains more than 1000 words, only the first 1000 words will be indexed and the rest will be silently ignored.
```
matboivin, line 22: I would keep key/value terms.
```suggestion
A maximum of __1000 values attached to one attribute__ is set. This means that if an attribute contains more than 1000 values, only the first 1000 values will be indexed and the rest will be silently ignored.
```

guides/advanced_guides/known_limitations.md
```diff
@@ -0,0 +1,31 @@
...
+MeiliSearch uses two databases: the first one for storage and the second one for updates.
+On launch LMDB needs to know the size that it can allocate on disk. This space will be reserved on disk for LMDB, thus MeiliSearch. This space will also be allocated as virtual memory.
+
+The maxime database size is by default __100GiB__ for each databases. This size can me modified using the options `--max-mdb-size` & `--max-udb-size` as described in the [configuration guide](/guides/advanced_guides/configuration.md#max-mdb-size).
+
+### Number of indexes
+
+You can create __up to 200 indexes__ in MeiliSearch. This limit has been hard set for performance reasons.
+
+### Maximum words per attribute
```
matboivin, line 20: (change if you agree if the previous suggestion)
```suggestion
### Maximum values per attribute
```

guides/advanced_guides/known_limitations.md
```diff
@@ -0,0 +1,31 @@
...
+
+A maximum of __1000 words par attribute__ can be indexed. This means that if an attribute contains more than 1000 words, only the first 1000 words will be indexed and the rest will be silently ignored.
+
+This limit is enforced for relevancy reasons. The more words there are in a given attribute, the less relevant the search queries will be.
```
matboivin, line 24: (change if you agree if the previous suggestion)
```suggestion
This limit is enforced for relevancy reasons. The more values are linked to a given attribute, the less relevant the search queries will be.
```

guides/advanced_guides/known_limitations.md
bidoubiwa: I think when thomas uses the word "indexes", he does not say "you can only have 1000 words per attribute" but "MeiliSearch can only **index** 1000 words per attributes". 
After which we explain what this imply. 

Also, there is a difference between values and word. A word is a group of chars inside a string separated by a white space: 
`wordOne wordtwo ..`

Whereas a value is the whole data linked to an attributes: 

```json
{
 "attribute": value
}
```

So, 1000 values for one attribute is technically not possible because an attribute can only have one value.

guides/advanced_guides/known_limitations.md
bidoubiwa: Per my answer on the previous suggestion I disagree with this suggestion

guides/advanced_guides/known_limitations.md
bidoubiwa: Per my answer on the previous suggestion I disagree with this suggestion

guides/advanced_guides/known_limitations.md
matboivin: Oh I see!

matboivin: I updated my suggestion. However, a longer explanation would be welcome.

bidoubiwa: Do you mean I should explain the problem better or this part in the article should be explained more in depth?

I think all those limitation should be added to the relative guides. As for example this information should be added to the `documents` guide. Where in the context of the guide it makes more sense. 

Also, we could add a sort of definition in the glossary for the word "indexation".
What do you think of that?

```
In meilisearch, documents are __indexed__. Indexing is the process of saving data in a database in a prefered way to fit your purpose. In MeilISearch's case, the purpose is to optimise word-based search. Because of that, in MeiliSearch the indexation is centered arround words. 
```

matboivin: I agree about adding limitations to each relative guide. And yes it was about clarifying "indexation" term so that would be a good idea to add it to the glossary.

bidoubiwa:
Could you rebase and `yarn style:fix` ?

PR #457, opened 2020-07-20
[System] Warning for system informations

Guikingone:
Hi everyone 👋 

As discussed [here](https://github.com/meilisearch/meilisearch-php/pull/67), the system informations has been deprecated so I think it could be a good idea to add a warning here 🙂 

Thanks for the feedback and have a great day 🙂

### DIFF
--- a/references/sys-info.md
+++ b/references/sys-info.md
@@ -1,5 +1,11 @@
 # System information
 
+::: warning
+
+The system informations endpoint has been deprecated in `0.13.0`.
+
+:::
+
 System information offers extended information about hardware usage by MeiliSearch (i.e, _processor_, _memory_).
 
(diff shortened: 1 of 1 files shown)


bidoubiwa:
Hey,

Thanks a lot for your PR. This will be merged once the v.0.13 is released.

bidoubiwa:
Hey @Guikingone,
After discussing it with the team we have decided that we are too early access to keep depreceted information. Thus, I would prefer if you remove the whole route instead of adding a warning. 

I would still like to thank you for your input to this documentation :)

bidoubiwa:
Hey, we have completely removed the route in another PR.
Thanks again for contributing

PR #516, opened 2020-09-21
Revised guides/main_concepts/documents.md

react-learner:


bidoubiwa (changes requested):
This is an amazing rework. Much more clear and easy to read. 

Could you add somewhere [this information](https://docs.meilisearch.com/guides/advanced_guides/known_limitations.html#maximum-words-per-attribute) about the maximum nbr of words in a value?

guides/main_concepts/documents.md
```diff
@@ -1,68 +1,90 @@
...
-You can also apply <clientGlossary word="ranking rules" /> to some fields. For example, you may decide recent movies should be more relevant than older ones.
+At this time, there are two field properties: [<clientGlossary word="searchable" />](/guides/advanced_guides/field_properties.md#searchable-fields) and [<clientGlossary word="displayed" />](/guides/advanced_guides/field_properties.md#displayed-fields). A field can have one, both, or neither of these properties. **By default, all fields in a document are both displayed and searchable.**
+
+To clarify, a field may be:
+* searchable but not displayed
+* displayed but not searchable
+* both displayed and searchable (default)
+* neither displayed nor searchable
+
+In the latter case, the field will be completely ignored when a search is performed. However, it will still be [stored](/guides/advanced_guides/field_properties.md#data-storing) in the document.
```
bidoubiwa: I think "attributeForFaceting" is also a field property. But contrary to the two other, attributeForFaceting is by default empty. 
Maybe we should simplify this part and add information about attributeForFaceting.
What do you think ?

bidoubiwa:
Style check is failing, could you `yarn style:fix` ?

bidoubiwa (approved):
LGTM

PR #518, opened 2020-09-22
Fix small typos in "Known Limitation"

Rauno56:
Per my research "Maxime" would technically be English, but pushed me off balance while reading it.


tpayet (approved):
Ahaha, lgtm! Thank you :D

PR #543, opened 2020-10-11
Fix endpoint of delete-batch example

redsolver:
The documentation shows the invalid path (404)`http://localhost:7700/indexes/movies/delete-batch` as example endpoint for the `delete-batch` operation. 

The correct endpoint should be `http://localhost:7700/indexes/movies/documents/delete-batch`


bidoubiwa (approved):
Perfect thanks a lot :)

PR #552, opened 2020-10-15
Snapshot and Dumps redesign

LegendreM:
Documentation related to:
- [Consider changing the extension name of dump and snapshot](https://github.com/meilisearch/MeiliSearch/issues/1010)
- [Snapshot redesign](https://github.com/meilisearch/MeiliSearch/issues/1009)
- [Update of the dump status body](https://github.com/meilisearch/MeiliSearch/issues/998)
- [Rename MEILI_DUMPS_FOLDER environment variable](https://github.com/meilisearch/MeiliSearch/issues/988)

### DIFF
--- a/guides/advanced_guides/configuration.md
+++ b/guides/advanced_guides/configuration.md
@@ -37,16 +37,17 @@ Server is listening on: http://127.0.0.1:7700
 
 - [Analytics](/guides/advanced_guides/configuration.md#analytics)
 - [Payload Limit Size](/guides/advanced_guides/configuration.md#payload-limit-size)
-- [Dumps](/guides/advanced_guides/configuration.md#dumps-folder)
-  - [Dumps folder](/guides/advanced_guides/configuration.md#dumps-folder)
-  - [Import dump](/guides/advanced_guides/configuration.md#import-dump)
-  - [Dump batch size](/guides/advanced_guides/configuration.md#dump-batch-size)
 - [Snapshoting](/guides/advanced_guides/configuration.md#schedule-snapshot-creation):
   - [Schedule Snapchot Creation](/guides/advanced_guides/configuration.md#schedule-snapshot-creation)
+  - [Snapshot Destination](/guides/advanced_guides/configuration.md#snapshot-destination)
   - [Snapshot Interval](/guides/advanced_guides/configuration.md#snapshot-interval)
-  - [Load From Snapshot](/guides/advanced_guides/configuration.md#load-from-snapshot)
+  - [Load From Snapshot](/guides/advanced_guides/configuration.md#import-snapshot)
   - [Ignore Missing Snapshot](/guides/advanced_guides/configuration.md#ignore-missing-snapshot)
-  - [Ignore Snapshot if Database Already Exists](/guides/advanced_guides/configuration.md#ignore-snapshot-if-db-exists)
+  - [Ignore Snapshot If DB Exists](/guides/advanced_guides/configuration.md#ignore-snapshot-if-db-exists)
+- [Dumps](/guides/advanced_guides/configuration.md#dumps-dir)
+  - [Dumps Destination](/guides/advanced_guides/configuration.md#dumps-destination)
+  - [Import Dump](/guides/advanced_guides/configuration.md#import-dump)
+  - [Dump Batch Size](/guides/advanced_guides/configuration.md#dump-batch-size)
 - [Max MDB Size](/guides/advanced_guides/configuration.md#max-mdb-size)
 - [Max UDB Size](/guides/advanced_guides/configuration.md#max-udb-size)
 - [SSL Configuration](/guides/advanced_guides/configuration.md#ssl-authentication-path):
@@ -253,14 +254,23 @@ We use [Sentry](https://sentry.io) to get bug reports and diagnostics, and impro
 
 ### Schedule Snapshot Creation
 
-**Environment variable**: `MEILI_SNAPSHOT_PATH`
-**CLI option**: `--snapshot-path`
+**Environment variable**: `MEILI_SCHEDULE_SNAPSHOT`
+**CLI option**: `--schedule-snapshot`
 
 The directory path where MeiliSearch will create snapshots.
 
 If this command is not called, snapshotting is deactivated.
 
 [Read more about snapshots](/guides/advanced_guides/snapshots_and_dumps.md#snapshots).
+
+### Snapshot Destination
+
+**Environment variable**: `MEILI_SNAPSHOT_DIR`
+**CLI option**: `--snapshot-dir`
+
+The directory path where MeiliSearch will create snapshots.
+
+**Default value**: `snapshots/`
 
 ### Snapshot interval
 
@@ -269,14 +279,12 @@ If this command is not called, snapshotting is deactivated.
 
 Defines the time gap in seconds between each snapshot creation.
 
-Requires `--snapshot-path` to be defined.
-
 **Default value**: `86400` (1 day)
 
-### Load from snapshot
+### import snapshot
 
-**Environment variable**: `MEILI_LOAD_FROM_SNAPSHOT`
-**CLI option**: `--load-from-snapshot`
+**Environment variable**: `MEILI_IMPORT_SNAPSHOT`
+**CLI option**: `--import-snapshot`
 
 The path of the snapshot file to import.
 
@@ -294,6 +302,8 @@ If this command is not called, no snapshot will be imported.
 
 The engine ignores missing snapshots and does not throw an error in this case.
 
+Requires `--import-snapshot` to be defined.
+
 ### Ignore snapshot if db exists
 
 **Environment variable**: `MEILI_IGNORE_SNAPSHOT_IF_DB_EXISTS`
@@ -301,12 +311,14 @@ The engine ignores missing snapshots and does not throw an error in this case.
 
 The engine skips snapshot importation if a database already exists. No error is thrown in this case.
 
-### Dumps folder
+Requires `--import-snapshot` to be defined.
 
-**Environment variable**: `MEILI_DUMPS_FOLDER`
-**CLI option**: `--dumps-folder`
+### Dumps destination
 
-Path of the folder where dumps will be created if the [dump route](/references/dump.md#create-a-dump) is called.
+**Environment variable**: `MEILI_DUMPS_DIR`
+**CLI option**: `--dumps-dir`
+
+Path of the directory where dumps will be created if the [dump route](/references/dump.md#create-a-dump) is called.
 
 **Default value**: `dumps/`
 
@@ -315,7 +327,7 @@ Path of the folder where dumps will be created if the [dump route](/references/d
 **Environment variable**: `MEILI_IMPORT_DUMP`
 **CLI option**: `--import-dump`
 
-Import a dump from the specified path. Must be a `.tar.gz` file.
+Import a dump from the specified path. Must be a `.dump` file.
 
 As the data contained in the dump needs to be indexed, the process will take an amount of time corresponding to the size of the dump. Only when the import is complete and successful will the MeiliSearch server start.
 
--- a/guides/advanced_guides/configuration.md
+++ b/guides/advanced_guides/configuration.md
@@ -283,7 +283,6 @@ Defines the time gap in seconds between each snapshot creation.
 
 ### Import Snapshot
 
-**Environment variable**: `MEILI_IMPORT_SNAPSHOT`
 **CLI option**: `--import-snapshot`
 
 The path of the snapshot file to import.
@@ -297,7 +296,6 @@ If this command is not called, no snapshot will be imported.
 
 ### Ignore Missing Snapshot
 
-**Environment variable**: `MEILI_IGNORE_MISSING_SNAPSHOT`
 **CLI option**: `--ignore-missing-snapshot`
 
 The engine ignores missing snapshots and does not throw an error in this case.
@@ -306,7 +304,6 @@ Requires `--import-snapshot` to be defined.
 
 ### Ignore Snapshot if DB Exists
 
-**Environment variable**: `MEILI_IGNORE_SNAPSHOT_IF_DB_EXISTS`
 **CLI option**: `--ignore-snapshot-if-db-exists`
 
 The engine skips snapshot importation if a database already exists. No error is thrown in this case.
@@ -324,7 +321,6 @@ Path of the directory where dumps will be created if the [dump route](/reference
 
 ### Import Dump
 
-**Environment variable**: `MEILI_IMPORT_DUMP`
 **CLI option**: `--import-dump`
 
 Import a dump from the specified path. Must be a `.dump` file.
--- a/guides/advanced_guides/configuration.md
+++ b/guides/advanced_guides/configuration.md
@@ -37,7 +37,7 @@ Server is listening on: http://127.0.0.1:7700
 
 - [Analytics](/guides/advanced_guides/configuration.md#analytics)
 - [Payload Limit Size](/guides/advanced_guides/configuration.md#payload-limit-size)
-- [Snapshoting](/guides/advanced_guides/configuration.md#schedule-snapshot-creation):
+- [Snapshots](/guides/advanced_guides/configuration.md#schedule-snapshot-creation):
   - [Schedule Snapchot Creation](/guides/advanced_guides/configuration.md#schedule-snapshot-creation)
   - [Snapshot Destination](/guides/advanced_guides/configuration.md#snapshot-destination)
   - [Snapshot Interval](/guides/advanced_guides/configuration.md#snapshot-interval)
--- a/guides/advanced_guides/configuration.md
+++ b/guides/advanced_guides/configuration.md
@@ -44,7 +44,7 @@ Server is listening on: http://127.0.0.1:7700
   - [Load From Snapshot](/guides/advanced_guides/configuration.md#import-snapshot)
   - [Ignore Missing Snapshot](/guides/advanced_guides/configuration.md#ignore-missing-snapshot)
   - [Ignore Snapshot If DB Exists](/guides/advanced_guides/configuration.md#ignore-snapshot-if-db-exists)
-- [Dumps](/guides/advanced_guides/configuration.md#dumps-dir)
+- [Dumps](/guides/advanced_guides/configuration.md#dumps-destination)
   - [Dumps Destination](/guides/advanced_guides/configuration.md#dumps-destination)
   - [Import Dump](/guides/advanced_guides/configuration.md#import-dump)
   - [Dump Batch Size](/guides/advanced_guides/configuration.md#dump-batch-size)
--- a/guides/advanced_guides/configuration.md
+++ b/guides/advanced_guides/configuration.md
@@ -306,7 +306,7 @@ Requires `--import-snapshot` to be defined.
 
 **CLI option**: `--ignore-snapshot-if-db-exists`
 
-The engine skips snapshot importation if a database already exists. No error is thrown in this case.
+If a database already exists, MeiliSearch will attempt to launch using that database instead of importing a snapshot. No error is thrown in this case.
 
 Requires `--import-snapshot` to be defined.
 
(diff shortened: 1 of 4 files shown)


guides/advanced_guides/configuration.md
```diff
@@ -253,30 +254,37 @@ We use [Sentry](https://sentry.io) to get bug reports and diagnostics, and impro
...
 ### Schedule Snapshot Creation
 
-**Environment variable**: `MEILI_SNAPSHOT_PATH`
-**CLI option**: `--snapshot-path`
+**Environment variable**: `MEILI_SCHEDULE_SNAPSHOT`
+**CLI option**: `--schedule-snapshot`
 
 The directory path where MeiliSearch will create snapshots.
 
 If this command is not called, snapshoting is deactivated.
```
bidoubiwa, line 262:
```suggestion
Activates snapshots scheduling. 

If this command is not added or its value is `false` snapshotting is deactivated.
```

guides/advanced_guides/configuration.md
```diff
@@ -37,10 +37,10 @@ Server is listening on: http://127.0.0.1:7700
 
 - [Analytics](/guides/advanced_guides/configuration.md#analytics)
 - [Payload Limit Size](/guides/advanced_guides/configuration.md#payload-limit-size)
-- [Dumps](/guides/advanced_guides/configuration.md#dumps-folder)
-  - [Dumps folder](/guides/advanced_guides/configuration.md#dumps-folder)
-  - [Import dump](/guides/advanced_guides/configuration.md#import-dump)
-  - [Dump batch size](/guides/advanced_guides/configuration.md#dump-batch-size)
+- [Dumps](/guides/advanced_guides/configuration.md#dumps-destination)
+  - [Dumps Destination](/guides/advanced_guides/configuration.md#dumps-destination)
```
bidoubiwa, line 41: I like this rename a lot

guides/advanced_guides/configuration.md
```diff
@@ -287,39 +295,43 @@ This command will stop the process if:
...
-### Ignore missing snapshot
+### Ignore Missing Snapshot
 
 **Environment variable**: `MEILI_IGNORE_MISSING_SNAPSHOT`
 **CLI option**: `--ignore-missing-snapshot`
 
 The engine ignores missing snapshots and does not throw an error in this case.
 
-### Ignore snapshot if db exists
+Requires `--import-snapshot` to be defined.
```
bidoubiwa, line 305: Does this mean an error is raised if the flag is given but `--import-snapshot` is not?

bidoubiwa (changes requested):

guides/advanced_guides/configuration.md
LegendreM: yep

bidoubiwa: Is this not considered like `--snapshot-path` as a setting that can be pre-added as an ENV variable but will only work once we add `--schedule-snapshot`? What is the idea behind this decision?

LegendreM: good point 🤔 

I think there is no reason to persist env_var for theses options:
- `--import-snapshot`
- `--import-dump`
- `--ignore-missing-snapshot`
- `--ignore-snapshot-if-db-exists`

 no?

guides/advanced_guides/snapshots_and_dumps.md
```diff
@@ -17,18 +17,28 @@ With it, it is possible to create and restore MeiliSearch from snapshots and sch
...
 By default, MeiliSearch schedules a snapshot creation every day. When the instance starts, it will wait 24 hours until the first snapshot. This means that if you have launched MeiliSearch on Tuesday at 4 pm, the first snapshot will be created on Wednesday at 4 pm, and so on every day.
 
 The schedule can be modified with the `--snapshot-interval-sec` flag.
 
 ```bash
-$ meilisearch --snapshot-path mySnapShots/ --snapshot-interval-sec 3600
+$ meilisearch --schedule-snapshot=true --snapshot-interval-sec 3600
 ```
 
 Now a snapshot will be created every hour.
```
bidoubiwa, line 44:
```suggestion
Now a snapshot is created every hour.
```

guides/advanced_guides/snapshots_and_dumps.md
```diff
@@ -41,17 +51,17 @@ During snapshot creation, snapshots are **overwritten**. Meaning, only the last
 
 Because snapshots are zipped copies of your database that did not go through any processing, starting a MeiliSearch instance from one of them is significantly faster.
```
bidoubiwa, line 52:
```suggestion
Because snapshots are hard copies of your database that did not go through any processing, starting a MeiliSearch instance from one of them is significantly faster than dumps.
```

Or should we say `as fast as starting MeiliSearch without a snapshot`?

guides/advanced_guides/snapshots_and_dumps.md
```diff
@@ -41,17 +51,17 @@ During snapshot creation, snapshots are **overwritten**. Meaning, only the last
...
 
-Using the global environment `MEILI_LOAD_FROM_SNAPSHOT` or the CLI flag `--load-from-snapshot` , MeiliSearch will start the server using the provided snapshot.
+Using the global environment `MEILI_IMPORT_SNAPSHOT` or the CLI flag `--import-snapshot` , MeiliSearch will start the server using the provided snapshot.
 
 ```bash
-$ meilisearch --load-from-snapshot mySnapShots/data.ms.tar.gz
+$ meilisearch --import-snapshot mySnapShots/data.ms.snapshot
 ```
 
 If a database already exists, or if no snapshot is found at the given path, MeiliSearch will **stop processing and throw an error**. This is configurable.
```
bidoubiwa, line 60:
```suggestion
If a database already exists, or if no snapshot is found at the given path, MeiliSearch will **stop its process and throw an error**. This is configurable.
```

guides/advanced_guides/snapshots_and_dumps.md
```diff
@@ -17,18 +17,28 @@ With it, it is possible to create and restore MeiliSearch from snapshots and sch
...
+
+The destination can be modified with the `--snapshot-dir` flag.
+
+```bash
+$ meilisearch --schedule-snapshot=true --snapshot-dir mySnapShots/
+```
+
+Now a snapshot will be created in `mySnapShots/` directory.
+
 By default, MeiliSearch schedules a snapshot creation every day. When the instance starts, it will wait 24 hours until the first snapshot. This means that if you have launched MeiliSearch on Tuesday at 4 pm, the first snapshot will be created on Wednesday at 4 pm, and so on every day.
```
LegendreM, line 36: now, the first snapshot is created when meilisearch is started

bidoubiwa (changes requested):

guides/advanced_guides/snapshots_and_dumps.md
```diff
@@ -17,41 +17,51 @@ With it, it is possible to create and restore MeiliSearch from snapshots and sch
...
+
+The destination can be modified with the `--snapshot-dir` flag.
+
+```bash
+$ meilisearch --schedule-snapshot=true --snapshot-dir mySnapShots/
+```
+
+Now snapshots are created in `mySnapShots/` directory.
+
 By default, MeiliSearch schedules a snapshot creation every day. When the instance starts, it will wait 24 hours until the first snapshot. This means that if you have launched MeiliSearch on Tuesday at 4 pm, the first snapshot will be created on Wednesday at 4 pm, and so on every day.
```
bidoubiwa, line 36:
```suggestion
The first snapshot is created right at the launch of MeiliSearch with the `--schedule-snapshot=true` option.
After that, the next snapshot is created by default every 24h hours. This goes on until you deactivate snapshots.
```

bidoubiwa (approved):

guides/advanced_guides/configuration.md
bidoubiwa: Mmmh, I see this more in a option file context. Where I would only have to add `--import-snapshot` when you need to quickly (or not) load a snapshot, and not all 5 options. But I also understand your point of view. @curquiza @eskombro @MarinPostma any input?

curquiza: > I think there is no reason to persist env_var for theses options:

>    --import-snapshot
>    --import-dump
>    --ignore-missing-snapshot
>    --ignore-snapshot-if-db-exists

> no?

I don't know what to think. I have kind of the same feeling, but I'm not sure at all, I'm not used to creating devops files with a lot of env var...etc
I a

guides/advanced_guides/configuration.md
```diff
@@ -287,39 +295,43 @@ This command will stop the process if:
 
 If this command is not called, no snapshot will be imported.
 
-### Ignore missing snapshot
+### Ignore Missing Snapshot
 
 **Environment variable**: `MEILI_IGNORE_MISSING_SNAPSHOT`
 **CLI option**: `--ignore-missing-snapshot`
 
 The engine ignores missing snapshots and does not throw an error in this case.
```
bidoubiwa, line 303:
```suggestion
The engine ignores missing snapshots in the case of a `--snapshot-import`, and does not throw an error in this case.
```

guides/advanced_guides/configuration.md
```diff
@@ -287,39 +295,43 @@ This command will stop the process if:
...
 
-### Ignore snapshot if db exists
+Requires `--import-snapshot` to be defined.
+
+### Ignore Snapshot If DB Exists
 
 **Environment variable**: `MEILI_IGNORE_SNAPSHOT_IF_DB_EXISTS`
 **CLI option**: `--ignore-snapshot-if-db-exists`
 
 The engine skips snapshot importation if a database already exists. No error is thrown in this case.
```
bidoubiwa, line 312:
```suggestion
The engine skips snapshot importation if a database already exists, in the case of a `--snapshot-import`. No error is thrown in this case.
```

bidoubiwa (changes requested):

guides/advanced_guides/configuration.md
LegendreM:
```suggestion
The engine skips snapshot importation if a database already exists, in the case of a `--import-snapshot`. No error is thrown in this case.
```

?

guides/advanced_guides/configuration.md
LegendreM:
```suggestion
The engine ignores missing snapshots in the case of a `--import-snapshot`, and does not throw an error in this case.
```

bidoubiwa: I want to make it clear that this variable has to do with the import snapshot process.

guides/advanced_guides/configuration.md
bidoubiwa: I want to make it clear that this variable has to do with the import snapshot process. You don't like it ?

guides/advanced_guides/configuration.md
LegendreM: just show you the real option, because it's less natural in reading with this version but "more true" in the usage

bidoubiwa: Ah yes! Ok I had to read it 200 times before seeing the difference. My bad!

guides/advanced_guides/configuration.md
eskombro: I think it's a good choice if two variables/options are meant to be used together, and you use just one, this shouldn't be just ignored without telling the user, and with this in mind, I agree with the requirement of `--import-snapshot` if you use `--ignore-missing-snapshot`. In the other hand, tehre are other cases where options are ignored, so just doing it here doesn't feel totally consistent.


I dont understand what you mean exactly @LegendreM by `I think there is no reason to persist env_var` so sorry if my answer is not pertnent.

bidoubiwa: Could we consider a warning instead of an error?

eskombro: I like the warning idea 👍

guides/advanced_guides/configuration.md
```diff
@@ -37,16 +37,17 @@ Server is listening on: http://127.0.0.1:7700
 
 - [Analytics](/guides/advanced_guides/configuration.md#analytics)
 - [Payload Limit Size](/guides/advanced_guides/configuration.md#payload-limit-size)
-- [Dumps](/guides/advanced_guides/configuration.md#dumps-folder)
-  - [Dumps folder](/guides/advanced_guides/configuration.md#dumps-folder)
-  - [Import dump](/guides/advanced_guides/configuration.md#import-dump)
-  - [Dump batch size](/guides/advanced_guides/configuration.md#dump-batch-size)
 - [Snapshoting](/guides/advanced_guides/configuration.md#schedule-snapshot-creation):
```
react-learner, line 40:
```suggestion
- [Snapshots](/guides/advanced_guides/configuration.md#schedule-snapshot-creation):
```

guides/advanced_guides/configuration.md
```diff
@@ -37,16 +37,17 @@ Server is listening on: http://127.0.0.1:7700
...
   - [Schedule Snapchot Creation](/guides/advanced_guides/configuration.md#schedule-snapshot-creation)
+  - [Snapshot Destination](/guides/advanced_guides/configuration.md#snapshot-destination)
   - [Snapshot Interval](/guides/advanced_guides/configuration.md#snapshot-interval)
-  - [Load From Snapshot](/guides/advanced_guides/configuration.md#load-from-snapshot)
+  - [Import Snapshot](/guides/advanced_guides/configuration.md#import-snapshot)
   - [Ignore Missing Snapshot](/guides/advanced_guides/configuration.md#ignore-missing-snapshot)
-  - [Ignore Snapshot if Database Already Exists](/guides/advanced_guides/configuration.md#ignore-snapshot-if-db-exists)
+  - [Ignore Snapshot If DB Exists](/guides/advanced_guides/configuration.md#ignore-snapshot-if-db-exists)
```
react-learner, line 46:
```suggestion
  - [Ignore Snapshot if DB Exists](/guides/advanced_guides/configuration.md#ignore-snapshot-if-db-exists)
```

guides/advanced_guides/configuration.md
```diff
@@ -253,30 +254,37 @@ We use [Sentry](https://sentry.io) to get bug reports and diagnostics, and impro
 
 ### Schedule Snapshot Creation
 
-**Environment variable**: `MEILI_SNAPSHOT_PATH`
-**CLI option**: `--snapshot-path`
+**Environment variable**: `MEILI_SCHEDULE_SNAPSHOT`
+**CLI option**: `--schedule-snapshot`
 
-The directory path where MeiliSearch will create snapshots.
+Activates snapshots scheduling.
```
react-learner, line 260:
```suggestion
Activates scheduled snapshots.
```

guides/advanced_guides/configuration.md
```diff
@@ -287,39 +295,43 @@ This command will stop the process if:
...
 
 **Environment variable**: `MEILI_IGNORE_MISSING_SNAPSHOT`
 **CLI option**: `--ignore-missing-snapshot`
 
 The engine ignores missing snapshots and does not throw an error in this case.
 
-### Ignore snapshot if db exists
+Requires `--import-snapshot` to be defined.
+
+### Ignore Snapshot If DB Exists
```
react-learner, line 307:
```suggestion
### Ignore Snapshot if DB Exists
```

guides/advanced_guides/configuration.md
react-learner, line 312:
```suggestion
The engine skips importing the given snapshot if a database already exists, and instead loads the existing database. No error is thrown in this case.
```

guides/advanced_guides/snapshots_and_dumps.md
```diff
@@ -17,21 +17,31 @@ Using this feature, it is possible to schedule snapshot creation at custom inter
...
+
+```bash
+$ meilisearch --schedule-snapshot=true --snapshot-dir mySnapShots/
+```
+
+Now snapshots are created in `mySnapShots/` directory.
+
+The first snapshot is created right at the launch of MeiliSearch. After that, the next snapshot is created by default every 24h hours. This goes on until you deactivate snapshots.
 
 The amount of time between each new snapshot can be modified with the `--snapshot-interval-sec` flag.
```
react-learner, line 38:
```suggestion
The first snapshot is created on launching MeiliSearch. After that, snapshots are created routinely on a set interval until you deactivate snapshots or end the MeiliSearch instance. By default, one snapshot is taken every 24 hours.

The amount of time between each new snapshot can be modified with the `--snapshot-interval-sec` flag.
```

guides/advanced_guides/snapshots_and_dumps.md
```diff
@@ -96,7 +106,7 @@ After your dump creation process is done, the dump file is created and added in
 Once you have exported a dump, which is a `.dump` file, you are now able to use that dump to launch MeiliSearch. As the data contained in the dump needs to be indexed, the process will take some time to complete. Only when the dump has been fully imported will the MeiliSearch server start, after which you can begin searching through your data.
 
 ```bash
-./meilisearch --import-dump /dumps/12345678.dump
+./meilisearch --import-dump /dumps/20200813-042312213.dump
 ```
 
 Because the indexation is the same process as when the documents are initially added to MeiliSearch, it is still a good practice to index documents in batches if the dataset is to big. The size of the batches are based on your dataset size and on your memory capacity.
```
react-learner, line 112:
```suggestion
Because importing a dump is the same process as when documents are initially indexed by MeiliSearch, it can require some time and memory. If your dataset is very large, it is good practice to index documents in larger batches. This will speed up the indexing process at the cost of memory.
```

react-learner (changes requested):

guides/advanced_guides/configuration.md
MarinPostma: I have been thinking about this. `--ignore-missing-snapshot` is an opt-in flag, I think that having an environement variable that sets this variable to always true is error prone. We should not design an API that leads our users to commit eventual error (it's very easy to forget that a variable in the environment has be set). Also, this flag makes not sense on it's own, and only makes sense at startup (unlike `--snapshot-dir`). The user shouldn't be allowed to launch a program with an invalid configuration (even if it appears like it does nothing bad). For this reason, I think this should be a hard error, and there should't be an env variable for this one, so previous configuration does not interfere with the current one.

eskombro: Agree with @MarinPostma POV!

So IMO we should keep the hard error as he says, and remove the env var as @LegendreM proposed 👌

guides/advanced_guides/configuration.md
LegendreM: Sorry @react-learner , English is probably better but it's harder for me to understand your version. I needed to read it 4 times to understand it. Have you another sentence to submit?

guides/advanced_guides/configuration.md
LegendreM: removed `ENV_VAR` for:

-    --import-snapshot
-    --import-dump
-   --ignore-missing-snapshot
-    --ignore-snapshot-if-db-exists

guides/advanced_guides/configuration.md
react-learner: @LegendreM You're right, it's confusing. How about this instead?

```suggestion
If a database already exists, MeiliSearch will attempt to launch using that database instead of importing a snapshot. No error is thrown in this case.
```

guides/advanced_guides/configuration.md
bidoubiwa: What does this imply? I have a hard time understanding.

bidoubiwa: @LegendreM explained it to me. So now all our options have an env variable alternative except those 4? That is not really consistent...

LegendreM: I think that `ENV_VAR` should not describe a starting behavior and should not be conflictual:
- if you both set `MEILI_IMPORT_DUMP` and `MEILI_IMPORT_SNAPSHOT`, it will conflict
- if you  set `MEILI_IMPORT_DUMP` you'll risk to delete data at the next start forgetting that you set this `ENV_VAR`

bidoubiwa: Oke, I understand. Apparently I was the one that did not understand what Marin said and everyone seems to agree. So lets go for that.

But this does not answer my original question: 
`where if you use --ignore-missing-snapshot without --import-snapshot it should not raise an error but a warning.`

Do we agree on this ?

LegendreM: I don't like warnings

bidoubiwa: This discussion continues here: 
https://github.com/meilisearch/MeiliSearch/issues/1033

I tried to gather all opinions.

PR #564, opened 2020-10-20
Prepare documentation for release 11/2020 of MeiliSearch

bidoubiwa:
## Prepare documentation for release 11/2020
- ~[ ] URL encode search in web UI meilisearch/meilisearch#989 & `&` Breaks Search on Dashboard meilisearch/meilisearch#986~
- ~[ ] Add code coverage meilisearch/meilisearch#978~
- ~[ ] Sort displayedAttributes field meilisearch/meilisearch#946~
- ~[ ] Add dependabot with a monthly update meilisearch/meilisearch#979~
- [x] Lazily create an index meilisearch/meilisearch#918 lazily & create an index on documents push meilisearch/meilisearch#914
- [x] Refactor `/health` route meilisearch/meilisearch#940
    - Remove update routes in references
    - Remove `maintenance` errorCode
    - Disclaimer in references that health will only return 204
- ~[ ] Some Settings seem to be overwritten when documents are pushed after updating settings meilisearch/meilisearch#942~
- [x] Dump info meilisearch/meilisearch#1008
   - Update dump and snapshot doc
- ~[ ] Remove not maintained csv movies dataset meilisearch/meilisearch#1032~
- [x] Change all genre fields to genres meilisearch/meilisearch#1031
    - Check in SDK's code-samples if this is correct
- ~[ ] Synonym unidecode meilisearch/meilisearch#1037 & Revert "Merge #1037" meilisearch/meilisearch#1045 & Synonyms case meilisearch/meilisearch#1044~


curquiza:
Is it worth removing the maintenance error code as it is not removed from the codebase? See https://github.com/meilisearch/MeiliSearch/blob/4398f2c0234cc6c25910572a8eea59db66c3bb47/meilisearch-http/src/error.rs

Should we open an issue to ask the core-team to remove it?
Or is it ok for everyone to have the errors documentation not really synchronized with the errors in the code base?

LegendreM:
I agree with @curquiza the health route don't return `maintenance` anymore but the error code still exists in meilisearch. If we want to remove this error code from documentation, it's only because the error is explicitly removed from meilisearch.

Which is not the case

bidoubiwa:
I agree that we should stay in sync, I think the error code should be removed from MeiliSearch and once it is done it will be removed here.

PR #581, opened 2020-11-03
Env needs to be lowercase

bb:
I typed the wrong value `Production` from this place in the documentation and got the following error message:

```
error: 'Production' isn't a valid value for '--env <env>'

[possible values: development, production]

Did you mean 'production'?

USAGE:

meilisearch --db-path <db-path> --dump-batch-size <dump-batch-size> --dumps-dir <dumps-dir> --env <env> --http-addr <http-addr> --http-payload-size-limit <http-payload-size-limit> --master-key <master-key> --max-mdb-size <max-mdb-size> --max-udb-size <max-udb-size> --no-analytics <no-analytics> --sentry-dsn <sentry-dsn> --snapshot-dir <snapshot-dir>


For more information try --help
```

### DIFF
--- a/guides/advanced_guides/configuration.md
+++ b/guides/advanced_guides/configuration.md
@@ -116,8 +116,8 @@ Analytics allow us to know how many users are using MeiliSearch and the followin
 
 By default, MeiliSearch runs in `development` mode.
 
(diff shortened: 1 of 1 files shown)


bidoubiwa:
Thanks a lot! Good find :)

bidoubiwa (approved):

PR #582, opened 2020-11-03
Change yarn installation cmd

bidoubiwa:


tpayet:
Is this the standard way to use yarn? Aren't we hiding what is happening behind the scene? 🎬

bidoubiwa:
There is no difference between both. These are allias. I suggested this because it seemed like it was less buggy on your side. But it could be non related.

bidoubiwa:
Closed because no apparent reason to do this change

PR #595, opened 2020-11-10
Clarify ranking rules behaviour in placeholder search

CaroFG:


guides/main_concepts/search.md
```diff
@@ -64,7 +64,7 @@ Meilisearch allows you to define [filters](/guides/advanced_guides/filtering.md)
 
 ### Placeholder Search
 
-Placeholder returns documents sorted by the [ranking rules](/guides/advanced_guides/settings.md#ranking-rules) specified by the user, without inputting any query words. It is compatible with faceting and filtering.
+Placeholder returns documents without inputting any query words. The returned documents will be sorted according to Meilisearch internal logic. To impact that order, [custom ranking rules](/guides/advanced_guides/settings.md#ranking-rules) need to be set. It is compatible with faceting and filtering.
```
bidoubiwa, line 67:
```suggestion
Placeholder returns documents without inputting any query words. The returned documents will be sorted according to MeiliSearch internal logic. To impact that order, [custom ranking rules](/guides/advanced_guides/settings.md#ranking-rules) need to be set. It is compatible with faceting and filtering.
```
>  It is compatible with faceting and filtering.

What is `it` in this context ?

bidoubiwa (changes requested):

guides/main_concepts/search.md
CaroFG: Good point.

guides/main_concepts/search.md
```diff
@@ -64,7 +64,7 @@ Meilisearch allows you to define [filters](/guides/advanced_guides/filtering.md)
 
 ### Placeholder Search
 
-Placeholder returns documents sorted by the [ranking rules](/guides/advanced_guides/settings.md#ranking-rules) specified by the user, without inputting any query words. It is compatible with faceting and filtering.
+Placeholder returns documents without inputting any query words. The returned documents will be sorted according to Meilisearch internal logic. To impact that order, [custom ranking rules](/guides/advanced_guides/settings.md#ranking-rules) need to be set. Placeholder search is compatible with faceting and filtering.
```
bidoubiwa, line 67:
```suggestion
Placeholder returns documents without inputting any query words. The returned documents will be sorted according to MeiliSearch internal logic. To impact that order, [custom ranking rules](/guides/advanced_guides/settings.md#ranking-rules) need to be set. Placeholder search is compatible with faceting and filtering.
```

bidoubiwa (changes requested):

guides/main_concepts/search.md
```diff
@@ -64,7 +64,7 @@ Meilisearch allows you to define [filters](/guides/advanced_guides/filtering.md)
 
 ### Placeholder Search
 
-Placeholder returns documents sorted by the [ranking rules](/guides/advanced_guides/settings.md#ranking-rules) specified by the user, without inputting any query words. It is compatible with faceting and filtering.
+Placeholder returns documents without inputting any query words. The returned documents will be sorted according to MeiliSearch internal logic. To impact that order, [custom ranking rules](/guides/advanced_guides/settings.md#ranking-rules) need to be set. Placeholder search is compatible with faceting and filtering.
 
 To perform a placeholder search, simply omit the query parameter.
```
react-learner, line 69:
```suggestion
Placeholder search makes it possible to view all documents in a given index sorted only by that index's [ranking rules](guides/main_concepts/relevancy.md#ranking-rules). MeiliSearch supplies default ranking rules, which can be [customized as needed](/guides/advanced_guides/settings.md#ranking-rules).

Placeholder search is compatible with faceting and filtering. To activate placeholder search, simply make a search without inputting any query words. 
```

react-learner (changes requested):

guides/main_concepts/search.md
bidoubiwa: With your suggestion @react-learner we fall within the same problem that this PR initially addressed. That is: 
The default ranking rules do not impact the order of the documents if there is no query term provided. 

Meaning, if no query is given and no custom ranking rule are made, the "default" order of the documents will be based on MeiliSearch internal logic. 
If you create a custom ranking rule, lets say `desc(release_date)`. When you'll be doing placeholder search (read: search without a query word) , documents will be ordered by the most recent movie to the oldest.

About `MeiliSearch internal logic`. Here is how it works. 
Imagine you have an index with two documents
```json
[
 { 
   "id": 1, 
   "title": "Batman"
},
 { 
   "id": 2, 
   "title": "Minions"
}

]
```

MeiliSearch will gather all documents id's: `1` and `2` apply a hash on them, that will transform them into a 64int: 
`1` => hash => `938..`(..64int)
`2`=> hash => `127..`(..64int)
Based on this hash, documents will be ordered in the database. because the hash of document id `2` is smaller in value (`127..`) as the hash from document id `1` (`938..`) they will be internally ordered like this:
`[127.., 938..]` 
Thus, one placeholder search they will be returned like this: 
```json
[
 { 
   "id": 2, 
   "title": "Minions"
},
 { 
   "id": 1, 
   "title": "Batman"
}

]
```

If we had made a custom ranking rule on id: `asc(id)` they would have been returned with `batman` first and then `Minions`.

---
So what we try to explain in this small paragraph is that the ordering of the placeholder search will not be affected by the **default** ranking rules but only the **custom ranking rules**. And when there are no **custom ranking rule** then it will be ordered based upon what I explained above.

react-learner: Hmm, I see the problem. How about this?
```suggestion
Placeholder search makes it possible to view all documents in a given index sorted by that index's [ranking rules](guides/main_concepts/relevancy.md#ranking-rules). To activate placeholder search, simply make a search without inputting any query words.

MeiliSearch's default ranking rules do not affect placeholder search—only custom rules [added by the user](/guides/advanced_guides/settings.md#ranking-rules). If no custom rules have been set, the results are displayed in an order that reflects their internal position in the database.

Placeholder search is compatible with faceting and filtering.
```

bidoubiwa: In your first paragraph, you are explaining a feature that orders documents based on ranking rules. Which is the same as what the search route already does with query terms. This is confusing, I think we should tell user as fast as possible in this paragraph that placeholder search is a search without any query words.

react-learner: Is this any better? 
```suggestion
If you make a search without inputting any query words, MeiliSearch will return all the documents in your index sorted by your [custom ranking rules](guides/advanced_guides/settings.html#custom-ranking-rule). This feature is called **placeholder search**. It is particularly effective when used with other features such as [faceting or filtering](guides/advanced_guides/faceted_search.html#filters-or-facets), which allow users to narrow their searches and browse by category.

Placeholder search is not affected by MeiliSearch's [default ranking rules](guides/advanced_guides/settings.html#ranking-rules)—only custom rules added by a user. If no custom rules have been set, the results are displayed in the order of their internal database position.
```

CaroFG: @bidoubiwa what do you think?

bidoubiwa: Good idea for the gif. Could you create an issue so that we do not forget about it? 
Your suggestion looks good to me. @CaroFG if you like it as well you can commit it :)

bidoubiwa:
You still have 3 dead links
```
  Should use .md instead of .html: [custom ranking rules](/guides/advanced_guides/settings.html#custom-ranking-rule) (/home/runner/work/documentation/documentation/guides/main_concepts/search.md:67:127)
  Should use .md instead of .html: [faceting or filtering](/guides/advanced_guides/faceted_search.html#filters-or-facets) (/home/runner/work/documentation/documentation/guides/main_concepts/search.md:67:324)
  Should use .md instead of .html: [default ranking rules](/guides/advanced_guides/settings.html#ranking-rules) (/home/runner/work/documentation/documentation/guides/main_concepts/search.md:69:53)
```

You can check them out by doing `yarn check-links` at the root of the directory

PR #615, opened 2020-12-01
Update Postman collection

curquiza:
It seems to have a lot of changes, but there are not: it's because I reorganized the orders of different requests. But actually, only the name of the collection changed.


tpayet (approved):
👍

Issue #580, opened 2020-11-03
Clarify documentation on searchableAttributes


bidoubiwa:
It is really not clear what `searchableAttributes` can do and how to use it in the documentation. It should be updated in the documentation in a way where it becomes clear for every user. 

### Fields to search in

Searchable attribute represents the lists of keys in which MeiliSearch will search when search queries are made. By default its value is `"*"`. This symbol means that all keys are added automatically. They are added the following way: 
- In the order they appeared in the first document added
- Every new key discovered in other documents are appended to the end of this list when they are discovered.

For example, if I add the following two documents 
```json
[
  {
    "title": "Batman",
    "id": 1
  },
   {
    "id": 2,
    "title": "Superman",
    "description": "a guy with a red underwear"
  }
]
```

In the first added document, the keys are ordered the following way, first `title` then `id`. This means that in searchableAttribute, this "*" symbol means this behind the scene: `["title", "id"]`. When the second document is added, the `description` key is discovered and added to the end of the list. Searchable Attributes now looks like this: `["title", "id", "description"]`.

### The fields order of importance

The order in which these keys are added is really important for the 4th ranking rule](https://docs.meilisearch.com/guides/main_concepts/relevancy.html#ranking-rules): `Attribute`. 

> 4. Attribute
Results are sorted according to the order of importance of the attributes: find documents that contain query terms in more important attributes first.

The order can not be known by the user unless he changes manually `"*"` with an array of keys `["title", "description"]`

To change it manually, one must know what keys are present in the documents and then decide how to order them. Once they have chosen the right order, let's say `["title", "description"]`, the need to update the `searchableAttribute` settings using the [update settings endpoint](https://docs.meilisearch.com/references/searchable_attributes.html). From now on, new keys present in the newly added documents will not be automatically added to the searchable attribute list. They will have to be added manually. 

Another important information is that when the `searchableAttributes` are reset, or if the whole settings are reset, the order that you have previously chosen will **not be reset**. This is not noticeable as the `searchableAttribute` as again this symbol `"*"`. 

### Inspirations

A good way to handle this explanation is to take inspiration from Algolia since they have the same setting: 
https://www.algolia.com/doc/guides/managing-results/must-do/searchable-attributes/.

as per requested here: https://github.com/meilisearch/specifications/pull/3#issuecomment-729940601

react-learner:
Fixed by #652

PR #652, opened 2020-12-23
Searchable attributes

react-learner:
Attempt to improve the documentation around searchable attributes as requested in #580 . 

"Attribute Importance" renamed to "Attribute Ranking Order". Information removed from API References and Main Concepts and consolidated in /guides/advanced_guides/field_properties.md.

I'm still unclear on the meaning of the last paragraph of #580 . Does this mean that if you manually reorder your searchableAttributes list, then reset it, the reset list will maintain the order you set, adding in all other attributes (new or previously non-searchable) to the end of the list?

If so, then wouldn't "reset" be a misnomer? It would actually be keeping your order while restoring the behavior of automatically adding all attributes.

### DIFF
--- a/guides/advanced_guides/field_properties.md
+++ b/guides/advanced_guides/field_properties.md
@@ -16,7 +16,7 @@ When you perform a search, all searchable fields are searched for matching query
 Non-searchable fields are most useful for internal information that's not relevant to the search experience, such as URLs, sales numbers, or ratings used exclusively for sorting results.
 
 ::: tip
-Even if you make a field non-searchable, it will remain on the server and can be made searchable again at a later time.
+Even if you make a field non-searchable, it will remain stored in the database and can be made searchable again at a later time.
 :::
 
 ### The Searchable Attributes List
--- a/guides/advanced_guides/field_properties.md
+++ b/guides/advanced_guides/field_properties.md
@@ -1,16 +1,50 @@
 # Field Properties
 
-By default, every field in a document is **searchable** and **displayed**. These properties can be modified in the [settings](/references/settings.md).
+By default, whenever a document is added to MeiliSearch, all new attributes found in it are automatically added to two lists:
+
+- **The [searchable attributes list](/guides/advanced_guides/field_properties.md#the-searchable-attributes-list)**: Attributes whose fields are searched for matching query words.
+- **The [displayed attributes list](/guides/advanced_guides/field_properties.md#displayed-fields)**: Attributes whose fields are displayed in documents.
+
+This means that by default, every field in a document is **searchable** and **displayed**. These properties can be modified in the [settings](/references/settings.md).
 
 ## Searchable Fields
 
-The **values** of the fields whose attributes are added to the [searchable-attributes list](/references/searchable_attributes.md) are **searched for matching query words**.
+A field can either be **searchable** or **non-searchable**.
 
-Their content is used by MeiliSearch to assess the relevancy of a document.
+When you perform a search, all searchable fields are searched for matching query words and used to assess document relevancy, while non-searchable fields are ignored entirely. **By default, all fields are searchable.**
 
-**By default, all field attributes are set as searchable**.
+Non-searchable fields are most useful for internal information that's not relevant to the search experience, such as URLs, sales numbers, or ratings used exclusively for sorting results.
 
-Therefore, if a field attribute is not in the searchable-attributes list, the field will be ignored from searches but will remain stored in the server. This list can be restricted to a selected set of attributes in the settings.
+::: tip
+Even if you make a field non-searchable, it will remain on the server and can be made searchable again at a later time.
+:::
+
+### The Searchable Attributes List
+
+MeiliSearch uses an ordered list to determine which attributes are searchable. The order in which attributes appear in this list also determines their [impact on relevancy](/guides/main_concepts/relevancy.md#attribute-ranking-order), from most impactful to least.
+
+In other words, the `searchableAttributes` list serves two purposes:
+
+1. It designates the fields that are searchable.
+2. It dictates the [attribute ranking order](/guides/main_concepts/relevancy.md#attribute-ranking-order).
+
+There are two possible modes for the `searchableAttributes` list.
+
+#### Default: Automatic
+
+**By default, all attributes are automatically added to the `searchableAttributes` list in their order of appearance.** This means that the initial order will be based on the order of attributes in the first document indexed, with each new attribute found in subsequent documents added at the end of this list.
+
+This default behavior is indicated by a `searchableAttributes` value of `["*"]`. To verify the current value of your `searchableAttributes` list, use the [get searchable attributes endpoint](/references/searchable_attributes.md#get-searchable-attributes).
+
+If you'd like to restore your searchable attributes list to this default behavior, simply [set `searchableAttributes` to an empty array `[]`](/references/searchable_attributes.md#update-searchable-attributes) or use the [reset searchable attributes endpoint](/references/searchable_attributes.md#reset-searchable-attributes).
+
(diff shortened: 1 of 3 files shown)


bidoubiwa:
> I'm still unclear on the meaning of the last paragraph of #580 . Does this mean that if you manually reorder your searchableAttributes list, then reset it, the reset list will maintain the order you set, adding in all other attributes (new or previously non-searchable) to the end of the list?

> If so, then wouldn't "reset" be a misnomer? It would actually be keeping your order while restoring the behavior of automatically adding all attributes.

Absolutely. This is something we are thinking on how we might improve it. The first step is to improve the documentation around it to receive feedback from users. After which, we can think (or not) about a rework. 

It could also be interesting to initiate conversations with users on slack in order to know if they realize this setting also orders the attributes by importance (@fharper).

guides/main_concepts/relevancy.md
```diff
@@ -157,31 +157,18 @@ To add a rule to the existing ranking rule, you have to add the rule to the exis
...
-
-Whenever a document is added to MeiliSearch, all new attributes found in it are automatically added to two lists:
-
-- **The [searchable attributes list](/references/searchable_attributes.md)**: Attributes of the fields in which to search for matching query words.
-- **The [displayed attributes list](/references/displayed_attributes.md)**: Attributes of the fields displayed in documents.
-
-This searchable attributes list is **ordered**, which means the order in which the attributes appear in the list determines their relevancy. Attributes are arranged from the most important attribute to the least important attribute.
-
-Place the attributes in the desired order and send this updated list using the [settings routes](/references/settings.md). Attributes will be re-ordered.
+For a more detailed look at this subject, see our reference page for [the searchable attributes list](/guides/advanced_guides/field_properties.md#the-searchable-attributes-list).
```
bidoubiwa, line 166: Should this sub-title be after `adding your rules` (like it is now) or should it be after `order of the rules` ? 
I feel like `order of the rules` and `attribute ranking order` are more similar and thus should be closer to each other. Also maybe both title could be more consistent in the wording. Lastly, it seems normal that `adding your own rule` comes last as it is the last step once you know all the theory. 

What do you think ? If you prefer it like this, it is ok for me.

guides/main_concepts/relevancy.md
```diff
@@ -157,31 +157,18 @@ To add a rule to the existing ranking rule, you have to add the rule to the exis
...
 
 #### Example
 
 ```json
-["title", "description", "director"]
+["title", "description", "release_date"]
 ```
 
-If you take a look at the above order, the matching words found in `title` will make the document more relevant than one with the same matching words found in `description` or `director`.
+With the above attribute ranking order, matching words found in the `title` field would have a higher impact on relevancy than the same words found in `description` or `release_date`. If you searched "1984", for example, results like Michael Radford's film "1984" would be ranked higher than movies released in the year 1984.
```
bidoubiwa, line 174: nice reference. The book is great, I forgot about the movie

references/searchable_attributes.md
```diff
@@ -2,8 +2,7 @@
 
 _Child route of the [settings route](/references/settings.md)._
 
-The values of the fields whose attributes are added to the searchable-attributes list are **searched for matching query words**.
-By default, all fields are considered to be `searchableAttributes`. This behavior is represented by the `["*"]` in the settings. Setting `searchableAttributes` to an empty array `[]` will reset the setting to its default value.
+The values associated with attributes in the `searchableAttributes` list are **searched for matching query words**. The order of the list also determines the [attribute ranking order](/guides/main_concepts/relevancy.md#attribute-ranking-order).
```
bidoubiwa, line 5: Should we add something like `at its default value of "*", the order of attributes are determined by their order of appearance in the first document added`? But maybe it is not the place to do so.

bidoubiwa (changes requested):
Awesome awesome job! This is made way more clear :)

guides/main_concepts/relevancy.md
```diff
@@ -157,31 +157,18 @@ To add a rule to the existing ranking rule, you have to add the rule to the exis
 ]
 ```
 
-## Importance of the attributes
+## Attribute Ranking Order
```
bidoubiwa, line 160: Should this sub-title be after adding your rules (like it is now) or should it be after order of the rules ?
I feel like order of the rules and attribute ranking order are more similar and thus should be closer to each other. Also maybe both title could be more consistent in the wording. Lastly, it seems normal that adding your own rule comes last as it is the last step once you know all the theory.

What do you think ? If you prefer it like this, it is ok for me.

references/searchable_attributes.md
react-learner: Great question. **For now**, I don't think we should add it. My goal at the moment is to make a precise distinction between references and other resources. Since this is an API reference for altering the index settings, I ideally want it to **only contain information related to that action**.

We might end up adding a line like this later. The behavior is confusing, so it might be useful to echo this explanation in as many places as possible. Personally, though, I think we should change the confusing behavior itself, instead of compensating with the docs.

guides/main_concepts/relevancy.md
react-learner: I think we should leave it like this for now so that all the info about ranking rules stays grouped together. Plus, I think we should aim to give Attribute Ranking Order its own page in the future, because it strikes me as an important feature that we don't want to get buried. <- Might do this as part of the upcoming restructure.

guides/advanced_guides/field_properties.md
```diff
@@ -1,20 +1,54 @@
...
+::: warning
+Be aware that after manually updating the `searchableAttributes` list, subsequent new attributes will no longer be automatically added unless the settings are [reset](/references/searchable_attributes.md#reset-searchable-attributes).
+:::
 
 #### Example
 
-Suppose you manage a database that contains information about movies. By adding the following settings, the fields `uid`, `movie_id`, `title`, `description`, `poster`, `release_date` and `rank` will be searched.
+Suppose you manage a database that contains information about movies. By adding the following settings, the fields `uid`, `movie_id`, `title`, `description`, `poster`, `release_date` and `rank` will be searchable, and ranked in that order.
 
 <CodeSamples id="field_properties_guide_searchable_1" />
```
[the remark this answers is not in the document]
react-learner: Agree 100%!

guides/main_concepts/relevancy.md
react-learner: I've never seen it, but I thought it was a good example! Only problem is, it might confuse people who try to follow along with the movies.json dataset 🤔

references/searchable_attributes.md
bidoubiwa: @qdequele

guides/main_concepts/relevancy.md
bidoubiwa: That's a nice tutorial or guide idea "How to order your fields by importance" something like that

guides/advanced_guides/field_properties.md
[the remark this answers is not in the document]
bidoubiwa: Since this code sample is precisely made for the guide that you have updated, could you create a new PR and change the `cURL` example in `.code-samples.meilisearch.yaml`. I will then see that there is a issue opened in every SDK directory to change their respective code-sample file.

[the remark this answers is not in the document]
react-learner: #679

bidoubiwa (dismissed):
Great great rewrite!! 🚀🔥

bidoubiwa (approved):
REALLY USEFUL GUIDE! Thanks so much. This is a key guide we absolutely needed 🔥🔥🔥🔥

react-learner:
Fixes #580

PR #656, opened 2020-12-28
add an action to update the year automatically

fharper:
trying with a manual workflow for test

### DIFF
--- a/.github/workflows/update-copyright.yml
+++ b/.github/workflows/update-copyright.yml
@@ -1,0 +1,9 @@
+name: Update copyright year(s) in license file
+
+on: workflow_dispatch
+
+jobs:
+  run:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: FantasticFiasco/action-update-license-year@v1


fharper:
Will modify to be automatic on January 1st if this action works well.

eskombro (approved):
That is an interesting GH action! Seems good! Any reason to prefer triggering it manually than using for example the CRON option?

LGTM 👌🏽

fharper:
Just want to try it first because automating it and adding it to all repos.

fharper:
Unfortunately, it doesn't work with `LICENSE.md`, just `LICENSE` file. I created an issue https://github.com/FantasticFiasco/action-update-license-year/issues/111

eskombro:
Shouldn't we actually call it `LICENSE` instead of `LICENSE.md`?

fharper:
Both are valid. I prefer Markdown so we can use markdown if needed. I also think all files should have an extension, even text based ones. Lastly, it is consistent with our other files like CONTRIBUTING.md.

It's not the end of the world for me, just sharing why I create LICENSE.md when I manually add a license to a repo.

eskombro:
Perfect :)

PR #677, opened 2021-01-06
Update facet filters section

csntr:
The docs mentioned attributesForFiltering which does not exist so I think what was meant is attributesForFaceting


react-learner (approved):
Thanks for noticing this! 🔥

PR #679, opened 2021-01-06
Field prop code sample

react-learner:
Improve the example for searchable fields in field_properties.md.
Idea comes from @bidoubiwa in https://github.com/meilisearch/documentation/pull/652#discussion_r551434483_


bidoubiwa:
@curquiza in this PR @react-learner changed a code-sample that will greatly improve the example in which it is used. 
This means we have to change it in every SDK that is showcased in the documentation. What would be the best way to do so?

bidoubiwa (dismissed):
LGTM 🚀
related https://github.com/meilisearch/integration-guides/issues/75

bidoubiwa (approved):

Issue #675, opened 2021-01-05
Need tokenizer documentation


react-learner:
Now that we are on the cusp of releasing our redone tokenizer (https://github.com/meilisearch/MeiliSearch/pull/1091), it seems like the perfect time to add more robust documentation of how this works "under the hood". I am envisioning a separate docs page dedicated to the tokenizer, which ideally would be ready before the new release.

I would greatly appreciate it if someone on the core team could write a technical description of the tokenizer and either send it to me, or open a PR into the prepare-meilisearch-2021-w1 branch. Once completed, @CaroFG would be able to link it in her v0.18 blog post.

Is this doable?

MarinPostma:
Many has some material, and there is the spec, what else do you need?

bidoubiwa:
This is the spec: https://github.com/meilisearch/specifications/blob/master/text/0001-script-based-tokenizer.md
It technically has everything you want to know, but it is maybe a bit hard to understand for non-rusters.
 
Before you arrive I read it and I called Many to ask some questions about what I did not understand (which was a lot). And he explained it to me very well. I would suggest maybe doing the same.

react-learner:
Okay, I will do my best to write a summary based on the spec (you're right, I don't understand all of it 😂) and the tracking issue (https://github.com/meilisearch/MeiliSearch/issues/624). I assume I'll get some stuff wrong, but I don't really need to understand everything—just the details that could affect users.

MarinPostma:
@react-learner call me if you need to get up to speed

bidoubiwa:
fixed by: #685

PR #685, opened 2021-01-07
New tokenizer documentation

react-learner:
See https://github.com/meilisearch/documentation/issues/675

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -64,6 +64,7 @@ module.exports = {
             '/guides/advanced_guides/concat',
             '/guides/advanced_guides/bucket_sort',
             '/guides/advanced_guides/known_limitations',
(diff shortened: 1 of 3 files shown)


react-learner:
I would like to link an issue around Russian/Cyrillic support in the last section, but the ones I found were both closed already:
https://github.com/meilisearch/MeiliSearch/issues/612
https://github.com/meilisearch/MeiliSearch/issues/552

react-learner:
I'm still not totally sure on the order of the analysis process. Does it comb through all the documents once, marking sections by language, and then go through a second time using the corresponding pipeline? Or does it analyze until the script changes and tokenize that section immediately? Can a single field feature multiple scripts? 🤔

react-learner:
Merging this so that all the documentation for the new release is ready to go tomorrow.

PR #686, opened 2021-01-09
Fix facetDistribution spelling

Nick-Mazuk:
facetDistribution should be facetsDistribution (with the "s")


react-learner (approved):
Thanks so much for spotting this! 🥇

PR #693, opened 2021-01-13
Fixes add_or_update_documents_1 code sample

react-learner:
Fix code sample for API references update document route (was POST, should be PUT).


bidoubiwa (approved):
LGTM 🔥 Also thanks @eskombro for finding that typo.

Issue #694, opened 2021-01-13
Clarify storage and memory requirements for MeiliSearch


Nick-Mazuk:
The [About Storage](https://docs.meilisearch.com/resources/about_storage.html#measured-disk-usage) page on the docs is great to learn about how MeiliSearch works with memory and storage, but I think the example at the bottom could be clearer.

For instance, when I was first choosing a Digital Ocean droplet to use with the MeiliSearch integration, I didn't have a good idea on what droplet I needed. Ideally, I'd like to see, "If you have X documents, you need Y RAM and Z disk space". I know this is not always possible, but even a clearer explanation with the example would have helped.

Specifically, this…

<img width="764" alt="Screen Shot 2021-01-13 at 1 57 53 PM" src="https://user-images.githubusercontent.com/5659171/104515208-56daa080-55a7-11eb-8b31-4b62c01960f8.png">

should be more like this…

<img width="779" alt="Screen Shot 2021-01-13 at 2 00 09 PM" src="https://user-images.githubusercontent.com/5659171/104515423-a7ea9480-55a7-11eb-9d33-14eeab17aaec.png">

Hopefully, this will more easily let people estimate the system requirements for their needs. Submitting a pull request with these changes.

PR #695, opened 2021-01-13
Clarify storage and memory requirements

Nick-Mazuk:
The example on the [About Storate](https://docs.meilisearch.com/resources/about_storage.html#measured-disk-usage) page could be much clearer in my opinion about how much RAM and disk space MeiliSearch actually uses.

This PR uses the example and explicitly says how much RAM and disk space was used for that data set, just in case someone didn't understand the difference between virtual memory size, private memory size, real memory size, LMDB, etc.

**Before merging**, please double-check that I used the correct numbers for RAM and disk space. I think they're right, but I'm not 100% sure.

Fixes https://github.com/meilisearch/documentation/issues/694


react-learner:
Looks good to me. Thanks for your contribution, @Nick-Mazuk ! 🔥

react-learner (approved):

PR #697, opened 2021-01-13
Prepare MeiliSearch for release w6 2021

react-learner:


reference/features/synonyms.md
```diff
@@ -1,69 +1,85 @@
...
 
-example:
+Use this when you want one word to be synonymous with another, but not the other way around.
 
 ```
 phone => iphone
 ```
 
-By searching `phone` you will get all results containing `iphone` with the same relevance. However, if you search for `iphone`, documents containing `phone` will not be shown in your results.
+By searching `phone`, you will get all results containing `iphone` with the same relevance. However, if you search for `iphone`, documents containing `phone` will not be returned in your results.
```
react-learner, line 33:
```
However, if you search for `iphone`, documents containing `phone` will not be returned in your results.
```
Does anybody want to test this and confirm that it's true? I'm worried that due to typo tolerance, querying `iPhone` might still return results that contain `phone`.

Can/should we think of a better example?

reference/features/known_limitations.md
```diff
@@ -1,17 +1,23 @@
-# Known limitations
+# Known Limitations
 
 MeiliSearch has known limitations, some of which may be addressed in future versions.
 
-Some of these limitations have been made by MeiliSearch developers for relevacy and/or performance while others limitations are enforced by LMDB, the key-value store used by MeiliSearch.
+Some of these limitations have been made by MeiliSearch developers for relevancy and/or performance, while others are enforced by [LMDB](/reference/under_the_hood/storage.md), the key-value store used by MeiliSearch.
```
fharper, line 5: Maybe putting `Lightning Memory-Mapped Database Manager` with LMDB in parentheses or the other way around. Or if not here, in the storage under the hood file. Even if we point to the website, it's nice to explain accronym and not expect people to know.

reference/features/synonyms.md
```diff
@@ -1,69 +1,85 @@
...
+
+```json
+{
+  "Résumé": ["CV"],
+  "CV": ["Résumé"]
+}
+```
 
-## The One-way association
+A search for "cv" would return any documents that contain the strings "Résumé", "resumé", or "resume", without being affected by case or accent marks.
```
fharper, line 23: and cv

fharper (dismissed):
I added some comments, but nothing crucial so LGTM

I will let someone from core get back to you about the iphone thing.

reference/features/synonyms.md
qdequele: The difference will be that without the synonyms documents with "phone" will be deprioritized behind documents with iPhones due to the type ranking rule. But if you add a synonym for "iPhone" => "phone", the documents with iPhone or phone will be classified without typo, so with the same relevancy.

reference/features/known_limitations.md
qdequele: I'm not sure that the real name is more understandable than the acronym, I've totally forgotten what `LMDB` means but I still know very well `LMDB`. I will prefer `LMDB` over `Lightning Memory-Mapped Database Manager`. `LMDB` talks to advanced developers but not to everyone, but I think that `Lightning Memory-Mapped Database Manager` is even less known.

reference/features/synonyms.md
```diff
@@ -1,69 +1,85 @@
...
-To create a multi-way synonym list this is the JSON that should be [added to the settings](/reference/api/synonyms.md#update-synonyms).
+Suppose you set "San Francisco" and "SF" as synonyms with a [mutual association](#mutual-association)
 
 ```json
 {
   "san francisco": ["sf"],
   "sf": ["san francisco"]
 }
 ```
+
```
qdequele, line 84: Could we add a little warning or a known limitation. Multi-Word Phrase is limited to three words. That's why `"The Lord of the Rings" => "TLOTR"` doesn't work.

reference/features/known_limitations.md
react-learner: You both make good points. I'm going to add the full name into /reference/under_the_hood/storage.md, just so that it's written somewhere and will show up if someone searches it for some reason.

bidoubiwa (approved):
LGTM! 

Small suggestion: 

Not sure about the `synonymes string are normalized..` that has been added in the api reference of synonyms. It seems like a specific bit of information that should be written together with all other specific bits of information about synonyms.

I think it would be better to let the user go to the guide to know more about synonyms than to add more text in the API. But its up to you.

react-learner:
> Not sure about the `synonymes string are normalized..` that has been added in the api reference of synonyms. It seems like a specific bit of information that should be written together with all other specific bits of information about synonyms.
> 
> I think it would be better to let the user go to the guide to know more about synonyms than to add more text in the API. But its up to you.

@bidoubiwa My logic is that most people looking at the API References are actively working and creating with MeiliSearch. If you're making or planning to make a synonym list, you may want to know how the synonyms will be processed. I don't think it needs to be explained fully on that page, so I chose to drop a link there just in case people want to know more.

fharper (approved):
LGTM

qdequele:
Ready for the merge?

PR #703, opened 2021-01-17
Remove axios vulnerability

bidoubiwa:
# Axios vulnerability

Oke! So this was a tricky one. 

## Problem analysiss

- `axios` is used by `meilisearch` < v0.14
- `docs-searchbar.js` uses `meilisearch`. 
    - Docs-searchbar >= v.1.1.8 [Upgraded to meilisearch](https://github.com/meilisearch/docs-searchbar.js/blob/v1.1.8/package.json) > v0.14
    - Before that version, because it uses meilisearch < v0.14, it had axios as a dependency of a dependency
- `vuepress-plugin-meilisearch` uses docs-searchbar.js. It uses the last version of docs-searchbar.js `v1.1.10`. Which uses meilisearch    `v0.17.0` which do NOT use `axios`.

So the 1 million question. Why with the latest version of `vuepress-plugin-meilisearch` do we still have `axios` as a dependency of dependencies?

## What is going on in yarn.lock
```
docs-searchbar.js@^1.1:
  version "1.1.7" // Why is it still 1.1.7 ?
  dependencies:
       meilisearch "^0.12.0" // Meilisearch that still uses axios

meilisearch@^0.12.0:
  version "0.12.0"
  dependencies:
    axios "^0.19.2"

vuepress-plugin-meilisearch@^0.10.6:
  version "0.10.6" // This version uses docs-searchbar@1.1.10 !
  dependencies:
    docs-searchbar.js "^1.1" // Here is the problem !
```

Because the version of docs-searchbar that changed from meilisearch v0.12 to v0.14 was a patch change `1.1.7` > `1.1.8`, and because we are giving as information `^1.1`, yarn.lock of documentation had no need to update the dependency.

## Solution

I removed and re-installed `vuepress-plugin-meilisearch` so that it would download the latests version of all dependency. 
This made the yarn.lock change to the following: 
```
docs-searchbar.js@^1.1:
  version "1.1.10"
  dependencies:
      meilisearch "^0.17.0" // Does not uses axios
```

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1745,10 +1745,10 @@ autocomplete.js@0.36.0:
   dependencies:
     immediate "^3.2.3"
 
-autocomplete.js@0.37.1:
-  version "0.37.1"
-  resolved "https://registry.yarnpkg.com/autocomplete.js/-/autocomplete.js-0.37.1.tgz#a29a048d827e7d2bf8f7df8b831766e5cc97df01"
-  integrity sha512-PgSe9fHYhZEsm/9jggbjtVsGXJkPLvd+9mC7gZJ662vVL5CRWEtm/mIrrzCx0MrNxHVwxD5d00UOn6NsmL2LUQ==
+autocomplete.js@0.38.0:
+  version "0.38.0"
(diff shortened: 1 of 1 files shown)


curquiza (approved):
👍

react-learner:
Thank you @bidoubiwa !! 🙇‍♂️

Issue #551, opened 2020-10-15
Fix search deployment pipeline


tpayet:
# Context
The documentation was initially deployed on GitHub pages using a GitHub action. Then we added the search feature with a GitHub action that is running a scraper that will feed the index after each deploy on GitHub pages.
At some point, we got some random deployment issues so we decided to move to Netlify. Netlify is integrated to GitHub and is watching when a push/commit is made to master so Netlify can build and deploy the documentation.
And the scraper action is still in place, running after the deploy on GitHub pages (which is not relevant since the URL address redirects to the Netlify deploy).

# Problem
The scraper action is not running after the deploy on Netlify, but after the deploy on GitHub pages which is not the deploy used for the documentation.

# Suggested solution
Set up the scraper action to be launched after the deployment on Netlify.
Netlify provides [notification webhook that notify when the build is done](https://docs.netlify.com/site-deploys/notifications/#outgoing-webhooks). And GitHub can trigger workflows (so the scraper action) [on manual event](https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/manually-running-a-workflow).
I did not find a GitHub action for that, it should exist but when I always find trigger from GitHub to Netlify and not the other way around 😅

fharper:
blocked by #671

bidoubiwa:
I'm not sure it should be blocked as the alternative for vuepress will not be for arriving and this fix, which I don't think would take more than a day to fix, and blocks the installation of the `bors` tool.  Also, even with another documentation, the problem would remain the same as it is not linked to the documentation itself but in the CI process and our choice of hosting service.

(Bors is really great to quickly rebase and merge branches.)

fharper:
By bad, I read doc deployment, not search deployment 🤦‍♂️

maryamsulemani97:
[fixed with the new site](https://github.com/meilisearch/documentation/blob/c16cc92ed25272689ef27510f1f030356f5a4c74/.github/workflows/scraper.yml). It uses a webhook to make sure that Vercel has finished deploying the website before running the scraper on it.

PR #728, opened 2021-01-28
Wait for netlify

react-learner:
Attempting to solve netlify deploy pipeline problems (#551).

Explanation of problem: Basically, each time there's a push to the production branch (Master), we run a scraper on our own docs site to fill a MeiliSearch instance (this is what powers the search bar on the docs site). The problem is that the scraper is running before Netlify has finished deploying and publishing the new build, so the search bar is often scraping the old content. In other words, items that were just added in the most recent push MAY NOT BE SEARCHABLE. 

This PR attempts to solve this problem using https://github.com/marketplace/actions/wait-for-netlify, a GitHub action that waits for an OK response from the Netlify site (or a certain number of seconds) before continuing subsequent actions.


react-learner:
I have my doubts that the gh action to trigger a build on Netlify is actually doing anything. I think Netlify is just watching the production branch and triggering a build on merge. Will have to investigate.

PR #755, opened 2021-02-10
Fixes code copy position

bidoubiwa:
Copy icons should not increase the size of the code blocks with this style :) 

Preview: 
Screenshot 2021-02-10 at 20 20 17


react-learner (approved):
LGTM 🔥🔥🔥

PR #757, opened 2021-02-11
Fix link in language.md

react-learner:
Fix link in language.md, and add language.md to Feature Reference table of contents.


fharper (approved):
LGTM

PR #766, opened 2021-02-15
Clarify Update fields

### LINKED ISSUE #765: get an update status and the "duration" field
Some [users seem to be confused](https://github.com/meilisearch/meilisearch-rust/issues/107) by the current documentation example of an update status -> https://docs.meilisearch.com/reference/api/updates.html#get-all-update-status

This example might indeed suggest the "duration" field is always present, (with `null` or with a value). But actually, the `duration` field is not present when the status is `enqueued` which can lead to code failure when trying to access this field.

Maybe should we add an example of an `enqueued` update, so that they know the expected return in this case? 🙂

Related to https://github.com/meilisearch/meilisearch-rust/issues/107

react-learner:
As stated in #765 , the current documentation is unclear on updates, particularly which fields may be contained in the response body when you make a call to GET updates. This PR attempts to improve that situation.

Fixes #765

### DIFF
--- a/learn/advanced/asynchronous_updates.md
+++ b/learn/advanced/asynchronous_updates.md
@@ -4,7 +4,7 @@ MeiliSearch is an **asynchronous API**. It means that the API does not behave as
 
 Some operations are put in a queue and will be executed in turn (asynchronously). In this case, the server response contains the identifier to track the execution of the operation.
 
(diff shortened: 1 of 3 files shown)


fharper (approved):
LGTM

PR #770, opened 2021-02-15
remove time from last updated

fharper:
added missing commit


fharper:
Please update your local master branch before testing this one since the package needed are already committed 🤦‍♂️

react-learner (approved):
LGTM :)

PR #774, opened 2021-02-15
hide copy button for response code box

fharper:
could have been done with JavaScript to remove the element, but I tricked it by rendering them invisible. It is working for code box without a language set or the JSON one since they are usually an example of return from the API

### DIFF
--- a/.vuepress/styles/index.styl
+++ b/.vuepress/styles/index.styl
@@ -14,3 +14,7 @@
     position: absolute;
     right: 0;
     bottom: 5px;
+.language-
+.language-json
+  div
+    visibility: hidden;


fharper:
We can change which type of code box it affect.

bidoubiwa:
This removes every copy button on every json boxes and every boxe without a proper language? I don't think it is a good idea. We may have other boxes in the doc that needs copy on those types of boxes and this also creates a behaviour that we have to explain to contributors as it is hidden in the css and not explicit during the creation of the code boxe.

bidoubiwa:
But you could replace the code box in the guide with this. You don't have to touch the css and we do not create unexpected behavior.

But the file is a bit less readable in markdown.

```
<div class="language- extra-class"><pre class="language-text code-copy-added"><code>888b     d888          d8b 888 d8b  .d8888b.                                    888
8888b   d8888          Y8P 888 Y8P d88P  Y88b                                   888
88888b.d88888              888     Y88b.                                        888
888Y88888P888  .d88b.  888 888 888  "Y888b.    .d88b.   8888b.  888d888 .d8888b 88888b.
888 Y888P 888 d8P  Y8b 888 888 888     "Y88b. d8P  Y8b     "88b 888P"  d88P"    888 "88b
888  Y8P  888 88888888 888 888 888       "888 88888888 .d888888 888    888      888  888
888   "   888 Y8b.     888 888 888 Y88b  d88P Y8b.     888  888 888    Y88b.    888  888
888       888  "Y8888  888 888 888  "Y8888P"   "Y8888  "Y888888 888     "Y8888P 888  888

Database path:       "./data.ms"
Server listening on: "127.0.0.1:7700"
</code></pre></div>
```

fharper:
> This removes every copy button on every json boxes and every boxe without a proper language?

Yes

> We may have other boxes in the doc that needs copy on those types of boxes

Which ones?


> this also creates a behaviour that we have to explain to contributors

Yes, it's fine, it's easy to explain and not complicated to understand. It's OK to give specific instructions to contributors: we will have to write some anyway to ensure a level of quality.

fharper:
> But you could replace the code box in the guide with this. You don't have to touch the css and we do not create unexpected behavior.

Let's not do this.

bidoubiwa:
> Let's not do this.

Why not? We already do that, in the `index.md` (the landing of the doc). 

Screenshot 2021-02-16 at 16 50 18

And it is still better than removing the possibility to copy json or none-types languages.

And in the API references there are a lot of places where you would want to copy the json to try it out. Example the settings: 

Screenshot 2021-02-16 at 16 51 18

> It's OK to give specific instructions to contributors: we will have to write some anyway to ensure a level of quality.

This seems like a specific thing to add for a rare use-case. In my opinion, but I could be wrong, it creates a technical debt as it is not intuitive that the CSS is the one removing the copy possibility. If it was in the plugins settings, it would be better, but I still thing that removing the possibility of copying json and non-language is a solution that will close possibilities on the assumption that we will not have a use case later.

fharper:
> Why not? We already do that, in the `index.md` (the landing of the doc).

I'll check it, we should stick to strict Markdown as much as possible in the documentation.

> In the API references there are a lot of places where you would want to copy the json to try it out. Example the settings:

Where are those lot? I'll check the one you share and update the CSS accordingly. If it doesn't make sense, we won't move forward with this.

> I still think that removing the possibility of copying json and non-language is a solution that will close possibilities on the assumption that we will not have a use case later.

If in the future it doesn't serve our needs, we change it like with anything we do code-wise, it's that simple.

We already spent too much time for something as trivial as that.

react-learner:
I agree with @bidoubiwa that JSON code samples should be copy+pastable (some are just examples, but others are genuinely useful and it's better to have the option than nothing). Could we make it so that the copy+paste button is hidden only on code samples without any specified language?

fharper:
I checked, and yes, there are some. I'll modify to include only those without a programming language. Thanks a lot @bidoubiwa for the help.

fharper:
For future reference, this is the plugin that is used for the copy button: https://github.com/znicholasbrown/vuepress-plugin-code-copy

fharper:
I'm closing this PR, let's leave the copy button like this for now as we will change platform anyway and have a better implementation.

PR #777, opened 2021-02-16
change JSON link to the raw directly

fharper:
make it faster for people wanting to curl it up right away


react-learner (approved):
Nice suggestion :)

PR #780, opened 2021-02-16
add missing line breaks

fharper:


react-learner (approved):
Big fan of this change. Now to make this page less text-heavy...

PR #772, opened 2021-02-15
Add recommended requirements answer to FAQ

react-learner:
Add information from #721 to the FAQ.

Closes #721

PR #786, opened 2021-02-16
Add AWS deploy guide to How To

react-learner:
Many thanks to @eskombro for writing this fantastic guide!


create/how_to/aws.md
```diff
@@ -0,0 +1,181 @@
...
+> We recommend prioritizing memory allocation for better MeiliSearch performance.
+> Note that the **free tier** is sufficient for tests or prototypes, but not recommended for large datasets.
+
+Once you've made your choice, click on **Next: Configure instance details** to continue.
+
+### 4. Instance details
+
+Here you can specify details of your MeiliSearch instance. Since **this section is not required to run MeiliSearch**, we won't cover it in this guide.
+
+![Instance details (optional)](/aws/04.instance-details.png)
```
eskombro, line 44:
```suggestion
![Instance details](/aws/04.instance-details.png)
```

eskombro (approved):
Thanks a lot @react-learner !!!
Amazing 🎉

create/how_to/aws.md
```diff
@@ -0,0 +1,181 @@
+
+# Deploy a MeiliSearch Instance on AWS
+
+Using our **MeiliSearch [AMI](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AMIs.html)**, MeiliSearch can be deployed on AWS in just a few minutes.
+The following `how-to guide` will walk you through every step to deploy MeiliSearch in an AWS EC2 instance.
+
+**Table of Contents**:
+
+[[toc]]
```
fharper, line 9: There is something weird, but it may not be just there, it may be on all the site. The permalink pointing to achors are working when you click them inside the page, but if you open a link on a new tab, it doesn't work. You can also see the issue if you copy a link pointing to an anchor, and paste it yourself in your browser, it will open the right page, but not scroll to the right position.

create/how_to/aws.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Part 1: Deploy an out-of-the-box MeiliSearch instance
+
+### 1. Launch an instance from the AWS console
+
+After logging into your [AWS Console](https://aws.amazon.com/console), navigate to the **Compute** service. Then go to **EC2**, and finally open your **Instances** console.
+
+![Launch an instance from AWS console](/aws/01.launch-instances.png)
```
fharper, line 17: The image quality isn't good. Not sure if it's the quality or the size. If it's the size that make it hard to read, we should link to the full size image. It's like this for all images.

create/how_to/aws.md
```diff
@@ -0,0 +1,181 @@
...
+
+Click on **Select** (right side of the screen) to confirm your choice.
+
+### 3. Size and specs
+
+Select the specifications of the server you want MeiliSearch to run on.
+
+![Size and specs](/aws/03.size-and-specs.png)
+
+> We recommend prioritizing memory allocation for better MeiliSearch performance.
```
fharper, line 35: I can't find where we mention this in the doc right now, but we should point to the section where we elaborate on minimum specifications or explanation on why memory is a possible issue. It will give more context for people that want it.

create/how_to/aws.md
```diff
@@ -0,0 +1,181 @@
...
+> Note that the **free tier** is sufficient for tests or prototypes, but not recommended for large datasets.
+
+Once you've made your choice, click on **Next: Configure instance details** to continue.
+
+### 4. Instance details
+
+Here you can specify details of your MeiliSearch instance. Since **this section is not required to run MeiliSearch**, we won't cover it in this guide.
```
fharper, line 42: Should we point to the AWS doc talking about that part? Feel like asshole than just "we won't help you suckers" 🤣

create/how_to/aws.md
```diff
@@ -0,0 +1,181 @@
...
+
+Simply click on **Next: Add Storage** to keep going.
+
+### 5. Storage
+
+Choose the storage **device** and **size** for your MeiliSearch instance.
+
+![Storage](/aws/05.storage.png)
+
+The amount of storage space required can vary drastically depending on the data you plan to index. In this example, we will use 25 GiB, which is more than enough for a small to medium dataset.
```
fharper, line 54: small to medium dataset means nothing, can we be clearer

create/how_to/aws.md
```diff
@@ -0,0 +1,181 @@
...
+
+```bash
+ssh admin@<your-ipv4-address>
+```
+
+```bash
+ssh admin@<your-domain-name>
+```
+
+Write `yes` and press `Enter` to accept the authentication process.
```
fharper, line 153: Write yes when asked what?

create/how_to/aws.md
```diff
@@ -0,0 +1,181 @@
...
+
+The server should answer with a `204 No content` status code as shown in the example below:
+
+```bash
+...
+< HTTP/1.1 204 No Content
+...
+```
+
+You're all set to use MeiliSearch in production with AWS!
```
fharper, line 181: Is our image in a repo somewhere? If not, should we put it out there so people can contribute to it? No matter about that, we should probably finish with something like: if you have any issues with our AWS image please create issues in the XYZ repo.

fharper (changes requested):
I added some comments of things that can be improve IMNSHO. Nothing is critical for merging this PR except the images quality.

create/how_to/aws.md
react-learner: I think this is a couple different behaviors.

1. If you select a header smaller than an `<h2>` the anchor link automatically changes to the parent header. I assume this is a VuePress thing but no idea if it's modifiable or not.
2.  If you open a page with lots of images (like our deploy guides), the page will scroll to the anchor BEFORE the images load. Then, when the images load in that pushes the content further down and you end up looking at the wrong place on the page.

Does that explain/cover the behavior you're seeing or is there another aspect to it? And do you see the same behavior on both the live site and the deploy previews?

create/how_to/aws.md
react-learner: Are you sure it's a problem with the image itself? On my browser the deploy previews open very zoomed-out. A couple `⌘` + `+`s solves the problem.

create/how_to/aws.md
react-learner: Totally agree. It's [still a PR](https://github.com/meilisearch/documentation/pull/772)

create/how_to/aws.md
fharper: No, it's a new one! Let me do a video :)

create/how_to/aws.md
fharper: So as we discovered, it's the size of the image that is causing problem, so I did another PR to add a zoom features. It doesn't change the fact that it's blurry, but it give the opportunity for people to see the full resolution version!

create/how_to/aws.md
fharper: oh, once @qdequele approve the PR, we can add the link. No need to wait for it for this PR, but be sure to remember to add it once it's merged.

create/how_to/aws.md
fharper: Or maybe it's the same issue, but check. The fact that VuePress cannot manage something as basic as links and anchors is another reason for switching IMNSHO

https://user-images.githubusercontent.com/1170989/108264838-06220e80-7136-11eb-961c-c88bc9ce948d.mp4

fharper: Anyway, it's a problem with VuePress, not this PR.

create/how_to/aws.md
react-learner: It's a little arbitrary, but I would say that
Small dataset: less than 10,000 documents
Medium dataset: less than 1 million documents
Large dataset: 1 million documents and above
Does that sound right to you?

react-learner: But of course it's *all* arbitrary.

I'm sure you could easily pass 25gb with a dataset that has only 10,000 documents if those documents each had 1000+ fields of 1000 or more characters. There's no good way to escape the lack of clarity here, so I think maybe just change it to:

```suggestion
The amount of storage space required can vary drastically depending on the data you plan to index. In this example, we will use 25 GiB, which is more than enough for most small datasets.
```

create/how_to/aws.md
react-learner: @eskombro

curquiza: https://github.com/meilisearch/meilisearch-aws

curquiza: Creating an issue into this repo is definitely a good way to help us if there is any trouble 👍

eskombro: Do you mean the AMI itself? If yes, it is stored in our AWS account (for MeilI orga) but the rope that @curquiza linked has all the tools that we use to build it, so it is public and open to contributions

react-learner: Thanks!

fharper (approved):

PR #789, opened 2021-02-17
Wait for Netlify

react-learner:
Wait for 200 from Netlify site (or 200 second timeout) before running the docs scraper. Also adds a small scraper test in reference/under_the_hood/README.md. Attempting to solve netlify deploy pipeline problems (#551).

**Explanation of problem:** Each time there's a push to the production branch (Master), we run a scraper on our docs site to fill a MeiliSearch instance index (this powers the search bar). The problem is that the scraper is running before Netlify has finished deploying and publishing the new build, so it ends up scraping the old outdated content. This means that items that were just added in the most recent commit MAY NOT BE SEARCHABLE until the next time build is triggered (the next commit on Master).

**Explanation of solution:** This PR uses https://github.com/marketplace/actions/wait-for-netlify, a GitHub action that waits for an OK response from a Netlify preview (or a certain number of seconds) before continuing subsequent actions. Unclear if it will work for a Netlify site to build rather than a preview.

### DIFF
--- a/.github/workflows/gh-pages-scraping.yml
+++ b/.github/workflows/gh-pages-scraping.yml
@@ -12,18 +12,16 @@ jobs:
     steps:
     - uses: actions/checkout@master
 
-    - name: Publish
-      uses: netlify/actions/build@master
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


tpayet (approved):
lgtm

PR #792, opened 2021-02-19
support Qovery as a deployment Option

evoxmusic:
Hi the Meili team,

I added docs on deploying Meilisearch to Qovery - feedback is welcome.

Romaric.


react-learner (approved):
Thanks Romaric, happy to have Qovery among our deploy options! 🚀

PR #799, opened 2021-02-22
Improve storage.md

react-learner:
Improve explanations of system requirements, virtual memory, etc.


reference/under_the_hood/storage.md
```diff
@@ -37,30 +37,31 @@ As a result, you may see that the disk occupied by LMDB and therefore by MeiliSe
...
+
+This dataset is a JSON file of 8.6 MB and has 19,553 documents. When we index this file in MeiliSearch, the amount of disk space taken by LMDB is 122MB.
 
 | Raw JSON | MeiliSearch database size on disk | Real memory size | Private memory size     | Virtual memory size |
 | -------- | --------------------------------- | ---------------- | ----------------------- | ------------------- |
-| 8.6 MB   | 122 MB                            | ≃ 6.3 MB         | 120 MB (≃ size on disk) | 204 Gb (memory map) |
+| 8.6 MB   | 122 MB                            | ≃ 6.3 MB         | 120 MB (≤ size on disk) | 204 Gb (memory map) |
 
-That means this dataset is using 6.3 MB of RAM and 122 MB of disk space.
+> This means the database is using 6.3 MB of RAM and 122 MB of disk space. Note that [virtual memory](https://www.enterprisestorageforum.com/hardware/virtual-memory/) refers only to disk space allocated by your computer for MeiliSearch—it does not mean that it's actually in use by the database. See [Memory Usage](#memory-usage) for more details.
```
fharper, line 63: Should we link to https://en.wikipedia.org/wiki/Virtual_memory instead?

reference/under_the_hood/storage.md
```diff
@@ -37,30 +37,31 @@ As a result, you may see that the disk occupied by LMDB and therefore by MeiliSe
...
 
-It is important to note that **there is no reliable way to predict the final size of a database**.
+It is important to note that **there is no reliable way to predict the final size of a database**. This is true for just about any search engine on the market—we're just the only ones saying it out loud.
```
fharper, line 65: I LOVE THAT 🤣

fharper (approved):
Just one small comment for an external link, everything else LGTM

reference/under_the_hood/storage.md
react-learner: I don't like linking to wikipedia, it's too confusing and theoretical to be helpful IMO. The link I chose was the most straightforward one I could find, but if we'd rather avoid the external link I can just link to the Memory Usage section instead.

fharper: I'm not against linking elsewhere, but I always prefer to link to Wikipedia since the content is always up to date and fixed when needed. The site won't go away either. I also find it more trustworthy for whatever reason.

Anyway, we can keep the link you had.

PR #802, opened 2021-02-23
Update configuration usage

bidoubiwa:
Configuration were not clear on the fact that a flag should always have a value associated


react-learner (changes requested):

react-learner:
Nice change, thanks :)

react-learner (approved):
LGTM! ❤️

PR #803, opened 2021-02-23
Add the JSON returned for the distinct attribute example

### LINKED ISSUE #782: Add the JSON returned for the distinct attribute example
Add the JSON returned for the [distinct attribute example](https://docs.meilisearch.com/reference/features/distinct.html#example)

image

CaroFG:
Fixes #782


react-learner (commented):
Thanks for adding this! 🙌 

Could you also add the following at the top of the file, after the first two paragraphs:
```
When multiple documents have the same value for a distinct attribute, MeiliSearch returns the first one after applying [ranking rules](/learn/core_concepts/relevancy.md#ranking-rules). If documents are equivalent in terms of ranking, MeiliSearch returns the first in terms of `internal_id`.
```

react-learner (approved):
LGTM, thanks!🕺

PR #816, opened 2021-03-01
Remove code-sample in template file

bidoubiwa:
The template file is used on the build to determine if samples are missing in the different SDK's. As this one is not needed anymore it should be removed!


react-learner (approved):
Makes sense, thanks @bidoubiwa ! 🕺

PR #843, opened 2021-03-10
Improve recommended requirements FAQ answer

react-learner:
Based on feedback from @qdequele .

### DIFF
--- a/create/faq.md
+++ b/create/faq.md
@@ -180,28 +180,26 @@ Until our first stable release (v1.0), MeiliSearch minor versions are not compat
 
 ## What are the recommended requirements for hosting a MeiliSearch instance?
 
-### Short Answer
+**The short answer: we don't know.**
 
-It's extremely difficult to accurately estimate the size and memory use of a MeiliSearch database, even knowing the size of the `.JSON` files. There's just too much complexity involved in the indexing process. For a benchmark and more thorough explanation, [go here](/reference/under_the_hood/storage.md#measured-disk-usage).
+The complexity involved in the indexing process makes it extremely difficult to accurately estimate the size and memory use of a MeiliSearch database, even knowing the size of the `.JSON` files beforehand.
 
-We recommend reading [the long answer below](#long-answer) to find the setup that works best for your needs.
+In the future, we hope to provide detailed benchmarks of different use-cases. Until then, read on for more detailed information about optimizing your MeiliSearch instance.
 
-### Long Answer
+### Memory Usage
 
-#### Memory Use
-
-When it comes to RAM, there are two things that can cause your usage to spike:
+There are two things that can cause your memory usage (RAM) to spike:
 
 1. Adding documents
 2. Updating index settings (if index contains documents)
 
-To reduce RAM usage, follow this best practice: always update index settings **before** adding your documents.
+To reduce memory use and indexing time, follow this best practice: **always update index settings before adding your documents**. This avoids unnecessary double-indexing.
 
 :::tip
-MeiliSearch's new engine (release date: sometime in 2021) will allow you to set a limit on how much RAM MeiliSearch uses.
+MeiliSearch's new engine (release date: before the end of 2021) **will allow you to set a limit on how much RAM MeiliSearch uses**.
 :::
 
-#### Disk Use
+### Disk Usage
 
 The following factors have a great impact on the size of your database (in no particular order):
 
@@ -213,10 +211,10 @@ The following factors have a great impact on the size of your database (in no pa
 - The number of different words present in the dataset
   
 :::tip
-Beware heavily multi-lingual datasets and datasets with many unique words, such as IDs or URLs. If you do have ID or URL fields, [make them non-searchable](/reference/api/searchable_attributes.md#update-searchable-attributes).
+Beware heavily multi-lingual datasets and datasets with many unique words, such as IDs or URLs, as they can slow search speed and greatly increase database size. If you do have ID or URL fields, [make them non-searchable](/reference/api/searchable_attributes.md#update-searchable-attributes) unless they're useful as search criteria.
 :::
 
-#### How does hardware affect search speed?
+### Search Speed
 
 Because MeiliSearch uses a [memory map](/reference/under_the_hood/storage.md#lmdb), **search speed is based on the ratio between RAM and database size**. In other words:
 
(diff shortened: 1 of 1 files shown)


create/faq.md
```diff
@@ -180,20 +180,43 @@ Until our first stable release (v1.0), MeiliSearch minor versions are not compat
 
 ## What are the recommended requirements for hosting a MeiliSearch instance?
 
-**Short answer**: It's impossible to say. We recommend [experimenting with RAM availability and database size](#how-does-hardware-affect-search-speed) to find the hardware + relevancy setup that works best for your dataset.
+### Short Answer
 
-**Long answer**: The disk size and RAM usage of your MeiliSearch database can vary widely. The following factors have a great effect on space and memory use:
+It's extremely difficult to accurately estimate the size and memory use of a MeiliSearch database, even knowing the size of the `.JSON` files. There's just too much complexity involved in the indexing process. For a benchmark and more thorough explanation, [go here](/reference/under_the_hood/storage.md#measured-disk-usage).
+
+We recommend reading [the long answer below](#long-answer) to find the setup that works best for your needs.
```
fharper, line 187: Do we need that sentence since right after, there is the "long answer" header.

create/faq.md
```diff
@@ -180,20 +180,43 @@ Until our first stable release (v1.0), MeiliSearch minor versions are not compat
...
+
+### Long Answer
+
+#### Memory Use
+
+When it comes to RAM, there are two things that can cause your usage to spike:
```
fharper, line 193: Wen it comes to memory usage
or
Wen it comes to memory usage (RAM)

create/faq.md
```diff
@@ -180,20 +180,43 @@ Until our first stable release (v1.0), MeiliSearch minor versions are not compat
...
+
+1. Adding documents
+2. Updating index settings (if index contains documents)
+
+To reduce RAM usage, follow this best practice: always update index settings **before** adding your documents.
```
fharper, line 198: To reduce memory usage

Some people may not know what is ram, many cloud provider just write "Memory 8Gb"

fharper, line 198: What happen if you don't? Would be nice to give more explanation on why they should do that.

create/faq.md
```diff
@@ -180,20 +180,43 @@ Until our first stable release (v1.0), MeiliSearch minor versions are not compat
...
+
+:::tip
+MeiliSearch's new engine (release date: sometime in 2021) will allow you to set a limit on how much RAM MeiliSearch uses.
```
fharper, line 201: Maybe "release date: sometime in 2021" could be "before the end of 2021"? Maybe it's just me, but "sometime in 2021" feels like we are not organized as "before the end..." is more like, we know when, but we don't make it public yet.

create/faq.md
```diff
@@ -180,20 +180,43 @@ Until our first stable release (v1.0), MeiliSearch minor versions are not compat
...
+:::
+
+#### Disk Use
+
+The following factors have a great impact on the size of your database (in no particular order):
```
fharper, line 206: I think I made the same comment elsewhere, but can we order them or it's impossible for us to order them?

create/faq.md
```diff
@@ -180,20 +180,43 @@ Until our first stable release (v1.0), MeiliSearch minor versions are not compat
...
+- The number of searchable fields
 - The number of faceted fields
 - The size of each update
-- The number of different words present in documents (beware heavily multi-lingual datasets and fields with unique words, such as IDs)
-
-All of this means that **it's almost impossible to estimate the size and memory usage of a MeiliSearch database** before creation, even based on the size of its documents. There are, however, a few [general principles](#how-does-hardware-affect-search-speed) you should know.
+- The number of different words present in the dataset
+  
+:::tip
+Beware heavily multi-lingual datasets and datasets with many unique words, such as IDs or URLs. If you do have ID or URL fields, [make them non-searchable](/reference/api/searchable_attributes.md#update-searchable-attributes).
```
fharper, line 216: make them non-searchable... if they aren't useful as a search criteria?

IDs I understand, but someone may want to let the user search URL as an example, so I don't want people to think we tell them not to make fields they need non searchable.

fharper, line 216: Maybe it's just me but I feel like the sentence isn't ended...

Beware heavily multi-lingual datasets and datasets with many unique words, such as IDs or URLs ... as they will slow down the search and make your database size bigger?

Something like that

create/faq.md
```diff
@@ -180,20 +180,43 @@ Until our first stable release (v1.0), MeiliSearch minor versions are not compat
...
+:::
 
-## How does hardware affect search speed?
+#### How does hardware affect search speed?
```
fharper, line 219: Are you sure? They are a question themselves, previous was is a H2.

Unless you want to move them under the "What are the recommended requirements for hosting a MeiliSearch instance?" question. If so, I would change the title for "Search Speed" and add the "Speeding Up MeiliSearch" under.

fharper (changes requested):

create/faq.md
react-learner: You're right, I decided that all of this information should be grouped under the same "Question" so it's easy to find, but in that case I should change the heading so it's not posing a separate question.

create/faq.md
```diff
@@ -180,20 +180,41 @@ Until our first stable release (v1.0), MeiliSearch minor versions are not compat
...
+
+There are two things that can cause your memory usage (RAM) to spike:
+
+1. Adding documents
+2. Updating index settings (if index contains documents)
+
+To reduce memory use and indexing time, follow this best practice: **always update index settings before adding your documents**. This avoids unnecessary double-indexing.
+
+:::tip
+MeiliSearch's new engine (release date: before the end of 2021) **will allow you to set a limit on how much RAM MeiliSearch uses**.
```
curquiza, line 199: The user will not be able to set the RAM usage once the new search engine is released, it will come later 🙂

react-learner: Didn't know that, thanks @curquiza ! Will delete this sentence.

bidoubiwa (approved):
🔥🔥🔥🔥

curquiza:
Hello, this is not a big deal but ff I'm not wrong this sentence was not removed as I required -> "MeiliSearch's new engine (release date: before the end of 2021) **will allow you to set a limit on how much RAM MeiliSearch uses**."

react-learner:
@curquiza See https://github.com/meilisearch/documentation/pull/921#pullrequestreview-648081961 😄

curquiza:
Thanks 😁

PR #847, opened 2021-03-10
add info about suggesting tutorials

fharper:


learn/getting_started/whats_next.md
```diff
@@ -18,4 +18,4 @@ And additional resources here:
 - [How-To Guides](/create/how_to)
 - [Postman Collection](/create/how_to/postman_collection.md): it can be tedious to re-write every route when wanting to try out an API. Try out MeiliSearch with our collection in Postman
 
-More tutorials are currently in development and will be available soon.
+More tutorials are currently in development and will be available soon. If you want to suggest a tutorial, please [create an issue](https://github.com/meilisearch/devrel/issues/new) in the [Developer Relations repository](https://github.com/meilisearch/devrel/).
```
react-learner, line 21: Wouldn't it make more sense to direct them to the documentation repo?

fharper:
Depends, I thought that we wanted to have the tutorials in the blog no? I mean, the non crucial ones.

react-learner:
@fharper Agreed, but I think it's confusing to call for them in the documentation if they won't go there, at least not without further explanation. I think we shouldn't add this here, it would make more sense in the new contribution section.

fharper:
Make sense

PR #849, opened 2021-03-10
beautify JSON

fharper:


react-learner:
Is this how we want to handle all JSON examples in the documentation? Personally I'm not a big fan. It will make the examples take up much more space on the page.

Whatever we decide, I think it's best to be consistent about it.

fharper:
Oh, I only saw this one last time. For clearer example, I would prefer that: it's the standard way to show JSON in learning mode, but I'll let you decide if you want to go ahead or not.

react-learner:
@fharper I don't have a strong opinion, so I will try to get some opinions from the next docs sync meeting and look at how other documentation sites represent nested JSON fields.

fharper:
Twillio is my reference for mostly everything, including [documentation](https://www.twilio.com/docs/api), after [DigitalOcean](https://developers.digitalocean.com/documentation/v2/) (not sure the tech writers wrote the API doc though, but here is an [article on JSON](https://www.digitalocean.com/community/tutorials/how-to-work-with-json-in-javascript)).

Since the documentation is in a learning context, the best practices of beautifying and aligned the code applied, and it's easier for someone to understand the object that way.

Actually, thinking about it, I would prefer we agree as I think it's important we move the coding examples like that for mentioned reason, what do you think?

PR #854, opened 2021-03-10
make links in code more accessible

fharper:


react-learner:
Could you explain the intention behind this change?

fharper:
If you have a link in a code block, there is no visual cue that tell there is a link. It was a mistake when the code plugin was added, they didn't ensure the CSS would follow for things like links.

image
image

react-learner (approved):
LGTM!

PR #856, opened 2021-03-10
add more generic word to data backup

fharper:
feel free to change the text or even change the text in the menu. The goal is to make it clear for someone who doesn't know what are snapshots or dump.


react-learner (approved):
Great addition, thanks!

PR #859, opened 2021-03-10
Improve wording re: guessing the primary key

react-learner:


bidoubiwa (commented):
I like it way more!

fharper (approved):
LGTM

PR #893, opened 2021-03-31
add Amplitude tracking

### LINKED ISSUE #892: Add analytics to the documentation website

fharper:
It's what we use instead of Google Analytics

Closes #892

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -396,5 +396,37 @@ module.exports = {
         defer: true,
       },
     ],
+    [
+      'script',
+      {},
+      `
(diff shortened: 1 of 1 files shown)


react-learner:
@fharper Could you add a more detailed explanation of what Amplitude will do for us, why we're taking this decision, and where the script in this PR comes from? For context, we have been using Hotjar for analytics until now.

fharper:
> @fharper Could you add a more detailed explanation of what Amplitude will do for us, why we're taking this decision, and where the script in this PR comes from? For context, we have been using Hotjar for analytics until now.

As discussed in our 1:1, we use HotJar for heat mapping, and we will use Amplitude for views tracking (à la Google Analytics).

react-learner (approved):
LGTM

PR #896, opened 2021-04-01
Remove inconsistency in snapshots documentation

### LINKED ISSUE #895: Inconsistency in snapshots documentation
[Documentation about snapshots](https://docs.meilisearch.com/reference/features/snapshots.html#starting-from-a-snapshot) states that there is an env variable (called `MEILI_IMPORT_SNAPSHOT`) that allows starting the server from the snapshot, but I didn't find this option on the [configuration page](https://docs.meilisearch.com/reference/features/configuration.html#options) or in the source code. This functionality can be only controlled via the CLI option, but the "Snapshots" page states that this can be done via the env variable (which is not true).

regzon:
Fixes #895


react-learner (approved):
LGTM! 🔥

PR #903, opened 2021-04-08
Open API - Add optional matches attributes to search hits

JacobLey:
Hi currently the search hits are defined as a generic object.

Which is true, but they could have two attibributes that have a predefined schema depending on the request. I updated the schemas to reflect that.

Relevant documentation:
https://docs.meilisearch.com/reference/features/search_parameters.html#matches
https://docs.meilisearch.com/reference/features/search_parameters.html#attributes-to-highlight


JacobLey:
Couple other cleanups as well:
* Bump version
* integer data type
* save creds in local storage

dichotommy (approved):
Looks good to me! Apologies for the delay, and thank you @JacobLey for all your contributions 🔥 🚀

PR #928, opened 2021-05-12
Update laravel in integration list

curquiza:
Following the deprecation of meilisearch-laravel-scout because we are now supported by the official laravel scout repo 🎉


react-learner (approved):
LGTM!

PR #943, opened 2021-05-27
fix the code samples for the search

irevoire:
I noticed we were using the same code samples for the GET search and the POST search. Because of that, the GET search was broken.

### DIFF
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -59,10 +59,13 @@ delete_documents_1: |-
         437035,
         363869
     ]'
-search_1: |-
+search_post_1: |-
   curl \
-    'http://localhost:7700/indexes/movies/search' \
+    -X POST 'http://localhost:7700/indexes/movies/search' \
     --data '{ "q": "american ninja" }'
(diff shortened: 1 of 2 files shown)


dichotommy:
@curquiza Does this solution create problems for the SDK code samples? I imagine it would require you to duplicate `search_1`, then rename one `search_post_1` and the other `search_get_1`. Seems unnecessary but I can't think of a better way to handle it 🤷🏻 

(It occurs to me that I don't even know which HTTP method the SDKs are using for their search() method, GET or POST?)

curquiza:
Hello @dichotommy and @irevoire!

It will indeed involve changes for the SDK code-sample, so I will check with Charlotte what is the best to do. Pretty sure it will involve changes in the SDK code sample anyway 😉
I will come back to you!

curquiza:
> (It occurs to me that I don't even know which HTTP method the SDKs are using for their search() method, GET or POST?)

FYI, the SDK `search()` method only uses the `POST` search. Only the JS SDK provides a `search_get` method to use the `GET` search.

curquiza (approved):

Issue #797, opened 2021-02-22
Copy button finds itself in the middle of code sample when horizontal scrolling


CaroFG:
https://docs.meilisearch.com/reference/features/filtering.html

The code sample after this line: "Querying on Planet of the Apes, the above example results in the following command:", when scrolling to see the end of the line, the copy button doesn't stay on the right corner.

PR #944, opened 2021-05-27
Fix the copy code button

irevoire:
I noticed a small “bug” when reading the documentation on my smol screen.

When you scroll to the right, the code copy button also scrolls:
image

It's ugly. With this fix the position of the button stays “fixed” to the bottom right of the screen:

image

---------------

This fix was made thanks to @lordteka

### DIFF
--- a/.vuepress/styles/index.styl
+++ b/.vuepress/styles/index.styl
@@ -18,9 +18,8 @@ body
     background-color #f3f5f7
     border-color #1e90ff
 .code-copy
-    position: absolute;
-    right: 0;
-    bottom: 5px;
+    position: sticky;
+    left: 0px;
 
 {$contentClass} a code
   color $accentColor


dichotommy:
This does indeed solve the problem described in #797 ! Thanks @irevoire and @lordteka . 

There is still one problem that isn't fixed. If a code sample has text in the final line, the code copy symbol overlaps slightly with the text ([numerous examples here](https://deploy-preview-944--dreamy-wiles-8fd85b.netlify.app/reference/features/filtering.html#examples)). Any thoughts on how to fix this @mdubus @guimachiavelli ?

Note that a temporary solution is perfectly acceptable in this case, since we will be leaving VuePress and migrating to Docusaurus within the next six months.

mdubus:
> There is still one problem that isn't fixed. If a code sample has text in the final line, the code copy symbol overlaps slightly with the text ([numerous examples here](https://deploy-preview-944--dreamy-wiles-8fd85b.netlify.app/reference/features/filtering.html#examples)). Any thoughts on how to fix this @mdubus @guimachiavelli ?

Yep there are multiple solutions here: 
- We can readapt the position of the icon, so that it doesn't overlap with the text
Capture d’écran 2021-05-31 à 10 32 33

- We can also add more padding bottom to the block, so that the button won't be too close to the text
Capture d’écran 2021-05-31 à 10 34 20

- We can also add padding and background-color to the icon, to match the background-color of the code block, but it doesn't seem to me to be an elegant way of dealing with this problem.

Don't hesitate to tell me what you would like for this button, and I can open a new PR to fix this 😉

dichotommy:
Thanks @mdubus ! I like the first solution. Let's merge this PR and open a new one 😄

dichotommy (approved):
LGTM 🔥

PR #950, opened 2021-05-31
Fix position of the copy button inside a code block

mdubus:
## Current problem 

The copy button inside a code block overlaps with text (see [here](https://deploy-preview-944--dreamy-wiles-8fd85b.netlify.app/reference/features/filtering.html#examples))

## Solution

Adapt the position of the button in style. Note that I had to use a `!important` because we use the external library `vuepress-plugin-code-copy` and it's the only way to override the initial style of the button.

## How to test 

Go to http://localhost:8081/reference/features/filtering.html#examples and check the position of the button

## Screenshots

Before: 
Capture d’écran 2021-05-31 à 11 33 09

After: 
Capture d’écran 2021-05-31 à 11 32 53

### DIFF
--- a/.vuepress/styles/index.styl
+++ b/.vuepress/styles/index.styl
@@ -20,6 +20,10 @@ body
 .code-copy
     position: sticky;
     left: 0px;
+    svg {
+      right: -6px;
+      bottom: -6px !important;
+    }
 
 {$contentClass} a code
   color $accentColor


dichotommy (approved):
Beautiful fix, thanks so much @mdubus 🙌🏻 🥇

Issue #963, opened 2021-06-03
Add .nvmrc to repo


guimachiavelli:
Ran into the [node versioning issue](https://github.com/meilisearch/documentation/pull/951) reported by @bidoubiwa earlier this week. 

Adding the node version requirement to the readme is great, but perhaps we should look into adding support for [NVM](https://github.com/nvm-sh/nvm)?

For those unfamiliar, NVM stands for Node Version Manager and works more or less like [RVM](https://rvm.io/): it looks for .nvmrc files specifying a node version and installs/switches to the required node version. When it works, at least.

Not sure how many contributors will have NVM installed on their machines, but it might make life easier for those who do.

bidoubiwa:
Good idea! We can also add this in the package.json

```js
"engines": {
    "node": ">=12 <15"
  }
```

based on this [doc](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#engines)

and to force the error message if the node version is not good the [following StackOverflow](https://stackoverflow.com/questions/29349684/how-can-i-specify-the-required-node-js-version-in-package-json) recommend this: 

> As ben pointed out below, creating a .npmrc file at the root of your project (the same level as your package.json file) with the text engine-strict=true will force an error during installation if the Node version is not compatible.

Thanks so much for having looked into it more

PR #966, opened 2021-06-04
Error thrown when installing documentation with wrong node version

bidoubiwa:
fixes: #963

Example of output:

```
➜ nvm use 10
Now using node v10.22.1 (npm v6.14.6)
➜ yarn
yarn install v1.22.5
[1/5] 🔍  Validating package.json...
error documentation@1.0.0: The engine "node" is incompatible with this module. Expected version ">=12 <15". Got "10.22.1"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
➜ nvm use 16
Now using node v16.2.0 (npm v7.13.0)
➜ yarn 
yarn install v1.22.5
[1/5] 🔍  Validating package.json...
error documentation@1.0.0: The engine "node" is incompatible with this module. Expected version ">=12 <15". Got "16.2.0"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
➜ nvm use 14
Now using node v14.17.0 (npm v6.14.13)
➜ yarn 
yarn install v1.22.5
[1/5] 🔍  Validating package.json...
```

Apparently netlify was still using `node 10` without any problem. But since I required in the package json the node version to be between 12 and under 15, the build first failed. 
Per [their documentation](https://docs.netlify.com/configure-builds/manage-dependencies/#node-js-and-javascript) `.nvmrc` acts as the node version netlify should use.


package.json
```diff
@@ -45,6 +45,9 @@
     "vuepress-plugin-sitemap": "^2.3.1",
     "vuepress-plugin-zooming": "^1.1.8"
   },
+  "engines": {
+    "node": ">=12 <15"
```
guimachiavelli, line 49: I think you mentioned this over slack, but why do we need the `engines` field? If I understand correctly both netlify and local give precedence to .nvmrc, no?

guimachiavelli (approved):

dichotommy:
Should we also create a `.npmrc` with `engine-strict=true`, or is that no longer necessary?

bidoubiwa:
> Should we also create a .npmrc with engine-strict=true, or is that no longer necessary?

I tried it and it worked without setting this

bidoubiwa:
**Update**

I tried without the `engine` and with only the `.nvmrc` using node 10 and having nvm locally, I didn't stop the `yarn install`

PR #970, opened 2021-06-07
Remove initial paragraph from Feature Reference intro

guimachiavelli:
First paragraph of the Feature Reference currently says: `Want to know something about MeiliSearch? Chances are high that it's written down here.`

Not sure what this adds. If I want to know about e.g. Core Concepts it's not actually in this section, so it's potentially misleading.


dichotommy (approved):
Nice change :)

dichotommy:
bors merge

PR #16, opened 2019-11-22
Added missing search params and advanced guide

bidoubiwa:

PR #973, opened 2021-06-07
Feature References audit — Authentication

guimachiavelli:
Some of the changes here are possibly bigger than they should at this stage. Feel free to push back if that's the case.

Adding @gmourier as a reviewer for the technical bits on keys and available routes.


reference/features/authentication.md
```diff
@@ -1,41 +1,45 @@
...
+When launching a MeiliSearch instance, you have the option of giving it a master key. By doing so, all routes will be protected and will require a key to be accessed.
 
-#### No master key
+You can specify a key by passing the `MEILI_MASTER_KEY` environment variable, or using the command line argument `--master-key`.
 
-If no master key is provided, all routes can be accessed without requiring any key.
+After adding the master key, you can retrieve both the private and the public keys on the [keys route](/reference/api/keys.md).
 
-## API Key
+## Using an authenticated MeiliSearch instance
```
gmourier, line 33:
```suggestion
## Communicate with a protected instance
```
I suggest something like this because the Meilisearch instance is not actually authenticated, it is the client that communicates with it that will authenticate itself through the header authentication key when it makes a request.

reference/features/authentication.md
```diff
@@ -1,41 +1,45 @@
...
   - `GET /indexes/:index_uid/documents/:doc_id`
-- Without any key, you can always access `GET /health`
-
-When a master key is provided to MeiliSearch, both the private and the public keys are automatically generated. **You cannot create any additional keys**.
 
-## Master Key
+When a master key is provided, both the private and the public keys are automatically generated. **You cannot create any additional keys**.
 
-When launching an instance, you have the option of giving a master key. By doing so, all routes will be protected and will require a key to be accessed.
+The only unprotected route available when using MeiliSearch in `production` is `GET /health`.
```
gmourier, line 23:
```suggestion
The only unprotected route available when using the built-in key-based authorization is `GET /health`.
```
Do we want to specify `production` since it can happen in `development` as soon as a master key has been configured ?

gmourier (changes requested):
Thanks, @guimachiavelli! Some thoughts you may consider before merging! Also, I think we are doing `Authorization` instead of `Authentication` because we are validating access to resources.

reference/features/authentication.md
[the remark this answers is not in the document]
guimachiavelli: Regarding consistency, shouldn't we also add links to the routes accessible with a public key (lines #16-19)? Or would that be a bit too much?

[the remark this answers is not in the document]
gmourier: It all depends on your guideline about the presence or not of links pointing to another documentation pages. Personally, I like to have links to referenced pages within a documentation to move faster around a subject. 🙂

reference/features/authentication.md
gmourier: I would say yes!

gmourier (approved):
👏 LGTM @guimachiavelli!

guimachiavelli:
@dichotommy do you also want to take a look at this?

reference/features/authentication.md
```diff
@@ -1,41 +1,45 @@
...
+The only unprotected route available when using the built-in key-based authorization is [`GET /health`](/reference/api/health).
 
-You can specify it by passing the `MEILI_MASTER_KEY` environment variable, or using the command line argument `--master-key`.
+## Adding a key
 
-You can retrieve both the private and the public keys using the master key on the [keys route](/reference/api/keys.md).
+When launching a MeiliSearch instance, you have the option of giving it a master key. By doing so, all routes will be protected and will require a key to be accessed.
 
-#### No master key
+You can specify a key by passing the `MEILI_MASTER_KEY` environment variable, or using the command line argument `--master-key`.
```
dichotommy, line 29:
````suggestion
You can specify the master key by setting the `MEILI_MASTER_KEY` environment variable, or passing the command line argument `--master-key` on launch.

#### Example

:::: tabs

::: tab Env

```bash
export MEILI_MASTER_KEY=whatever_you_want
meilisearch
```

:::

::: tab CLI

```bash
meilisearch --master-key=whatever_you_want
```

:::

::::
````
Not 100% sure I typed that right, but should create an example code block with two tabs, one for env and one for CLI. If you go with this suggestion, verify that it displays correctly before merging :)

Comment (dichotommy) [reply reference/features/authentication.md:29]: Could also ignore this for now and add it in a separate PR if preferred.

File: reference/features/authentication.md [hunk 4]
```diff
@@ -1,41 +1,45 @@
...
 
-If no master key is provided, all routes can be accessed without requiring any key.
+After adding the master key, you can retrieve both the private and the public keys on the [keys route](/reference/api/keys.md).
```
dichotommy, line 31:
```suggestion
After setting up the master key, you can retrieve both the private and the public keys with the [keys route](/reference/api/keys.md).
```

reference/features/authentication.md
```diff
@@ -1,41 +1,45 @@
...
+  - [`GET /indexes/:index_uid/documents/:document_id`](/reference/api/documents#get-one-document)
 
-## Master Key
+When a master key is provided, both the private and the public keys are automatically generated. **You cannot create any additional keys**.
 
-When launching an instance, you have the option of giving a master key. By doing so, all routes will be protected and will require a key to be accessed.
+The only unprotected route available when using the built-in key-based authorization is [`GET /health`](/reference/api/health).
 
-You can specify it by passing the `MEILI_MASTER_KEY` environment variable, or using the command line argument `--master-key`.
+## Adding a key
```
dichotommy, line 25:
```suggestion
## Adding the master key
```

reference/features/authentication.md
```diff
@@ -1,41 +1,45 @@
...
 
-## API Key
+## Communicating with a protected instance
 
-If a master key is set, on each API call, a key must be added to [the header](/reference/api/README.md#authentication).
+When using authentication, a key must be added to [the header](/reference/api/README.md#authentication) of each API call.
 
-If no or a wrong API key is provided in the header you will have no access to any route and you will receive the
-`HTTP/1.1 403 Forbidden` status code.
+We strongly recommend only using the master key to retrieve the public and private keys.
```
dichotommy, line 37:
```suggestion
We strongly discourage using the master key for API calls. It is intended only for retrieving the public and private keys.
```

reference/features/authentication.md
```diff
@@ -1,41 +1,45 @@
...
+If an invalid key is provided, you will receive the `HTTP/1.1 403 Forbidden` status code. You will receive the same error if you fail to provide a key when querying a protected route.
 
-Since both the private and the public keys are generated based on your master key, changing the master key will result in the modification of the two other keys.
+## Changing a key
 
-After having changed your master key, it is mandatory to restart the MeiliSearch server to ensure the renewal of the private and the public keys.
+**Changing the master key will automatically generate new private and public keys**. It is not possible to change one key without altering the others.
 
-**All keys will be changed**. Therefore, it is not possible to change only one of the keys.
+After changing the master key, it is mandatory to restart the MeiliSearch instance to ensure the renewal of the private and the public keys.
```
dichotommy, line 45:
```suggestion
After changing the master key, it is mandatory to restart the MeiliSearch instance to generate new private and public keys.
```

dichotommy (changes requested):
Fantastic work 🔥 🔥 🔥  There is more to do here, but it's already much clearer.

Some things to consider for future PRs:
- We should explicitly state somewhere that you activate authentication by setting the master key
- We might prefer language related to "security" to language related to "authentication". I don't think there's an easy or clear way to say "an instance which has authentication activated" (an authenticated instance? ehhh 😕), whereas "a secure instance" is very clear.
- It occurs to me that, based on the docs, it's not clear how to deactivate key-based authentication after you've set the master key once. I assume that you could do so with `export MEILI_MASTER_KEY=""` or `--master-key=""`.

Anyway, super job!

reference/features/authentication.md
guimachiavelli: sounds great, but one minor comment: what do you say we replace `whatever_you_want` with something like `[YOUR_MASTER_KEY]`? Sure, people could use whatever string they want, but generally we'd like them to use something that's hard to guess and brute-force, no?

dichotommy: Yes, great suggestion :)

dichotommy:
> How do you feel about "a protected instance"?

I like it, although I'm not sure if it's better than "a secure instance". I wonder if we could use both, or if that would be needlessly confusing.

(It's funny that good practices for general writing, such as avoiding repetition, are often bad practices for technical writing 🙃 )

> I'll check with the devs

Thanks! And thanks again for your work on this PR!

guimachiavelli:
> I like it, although I'm not sure if it's better than "a secure instance". I wonder if we could use both, or if that would be needlessly confusing.

My thoughts exactly. Is it bad practice to define the equivalence of two terms in the beginning of a document? `Adding the master key turns a MeiliSearch instance into a secured instance. Throughout this document we will use the terms secured instance and protected instance interchangeably.` Or something a bit better written than that; you get the idea.

dichotommy:
I think that's reasonable. You could also prime people to see the two terms as equivalent in a less direct way (more risky, but more space efficient): `**Adding the master key turns on security for your MeiliSearch instance, protecting your API routes from unauthorized access.**`

But in the end, the only justification for using these two different terms is aesthetics, right? Whereas clarity strikes me as being more important for almost every type of documentation.

If we're interested in both word variety and consistency, maybe we describe a MeiliSearch instance with a master key as `secure`, and its routes/endpoints as `protected`. It might not be a useful distinction, but I think it would make sense to a user.

guimachiavelli:
@dichotommy's comments addressed: accepted suggestions and added explicit instructions on how to activate and deactivate auth.

dichotommy (approved):
Looks good to me!

dichotommy (approved):

dichotommy:
bors try

dichotommy:
@guimachiavelli I think we will want to use GitHub's built-in `Squash and Merge` for this one, so let's either rebase this together on Monday (and hopefully get rid of those duplicated commits from Master), or copy your changes into a fresh branch 😄

dichotommy (approved):
Looks great, go ahead and squash that baby!

guimachiavelli:
Baby Squashed 👶

PR #980, opened 2021-06-11
[react example] fix JS example: import Meilisearch requires {}

deterralba:
The current example does not work with the latest version of the lib.


dichotommy (approved):
Thanks for spotting this @deterralba 🔥 🚀

dichotommy:
bors merge

PR #983, opened 2021-06-15
Update faq.md

dichotommy:
Make the wording a bit nicer.

### DIFF
--- a/resources/faq.md
+++ b/resources/faq.md
@@ -181,7 +181,7 @@ Until our first stable release (v1.0), MeiliSearch minor versions are not compat
 
 **The short answer: we don't know yet!**
 
-Indexing documents is a highly complex process, making it difficult to accurately estimate the size and memory use of a MeiliSearch database. In addition, our engine is improving almost constantly.
+Indexing documents is a highly complex process, making it difficult to accurately estimate the size and memory use of a MeiliSearch database. At this stage, our engine is changing constantly, which means benchmarks are constantly out of date.
 
 Once our engine is completely stable (v1), we will provide detailed benchmarks with different datasets. Until then, read on to learn how to optimize your MeiliSearch instance.
 
--- a/resources/faq.md
+++ b/resources/faq.md
@@ -179,11 +179,11 @@ Until our first stable release (v1.0), MeiliSearch minor versions are not compat
 
 ## What are the recommended requirements for hosting a MeiliSearch instance?
 
-**The short answer: we don't know.**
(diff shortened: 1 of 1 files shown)


resources/faq.md
```diff
@@ -179,11 +179,11 @@ Until our first stable release (v1.0), MeiliSearch minor versions are not compat
 
 ## What are the recommended requirements for hosting a MeiliSearch instance?
 
-**The short answer: we don't know.**
+**The short answer: we don't know yet!**
 
-The complexity involved in the indexing process makes it extremely difficult to accurately estimate the size and memory use of a MeiliSearch database, even knowing the size of the `.JSON` files beforehand.
+Indexing documents is a highly complex process, making it difficult to accurately estimate the size and memory use of a MeiliSearch database. In addition, our engine is improving almost constantly.
```
guimachiavelli, line 184: Isn't `almost constantly` a borderline contradiction in terms? If the problem is that some updates actually cause performance regressions (and thus we're not *improving* constantly), perhaps we can do something like:

```suggestion
Indexing documents is a highly complex process, making it difficult to accurately estimate the size and memory use of a MeiliSearch database. At this stage, our engine is changing constantly, which means benchmarks are constantly out of date.
```

dichotommy: Nice change, thanks! 😄

resources/faq.md
```diff
@@ -179,11 +179,11 @@ Until our first stable release (v1.0), MeiliSearch minor versions are not compat
 
 ## What are the recommended requirements for hosting a MeiliSearch instance?
 
-**The short answer: we don't know.**
+**The short answer: we don't know yet!**
 
-The complexity involved in the indexing process makes it extremely difficult to accurately estimate the size and memory use of a MeiliSearch database, even knowing the size of the `.JSON` files beforehand.
+Indexing documents is a highly complex process, making it difficult to accurately estimate the size and memory use of a MeiliSearch database. At this stage, our engine is changing constantly, which means benchmarks are constantly out of date.
```
guimachiavelli, line 184: Oops, my bad, repeated constantly.

```suggestion
Indexing documents is a highly complex process, making it difficult to accurately estimate the size and memory use of a MeiliSearch database. At this stage, our engine is changing constantly, which means benchmarks are often out of date.
```

guimachiavelli (approved):

resources/faq.md
dichotommy: I prefer it the way it is 😁

PR #984, opened 2021-06-15
Change search code in code sample template

bidoubiwa:
Update code sample template to remove old one


dichotommy (approved):
This will stop us getting errors from looking for `search_1`, which was changed to `search_post_1` and `search_get_1`.

We're leaving off `search_get_1` because it's only an option for CURL and JavaScript, plus we will remove this code sample from the docs soon.

PR #987, opened 2021-06-15
Feature reference audit: distinct attributes

guimachiavelli:
The biggest change here is the introduction of a initial paragraph with a quick explanation of what is a distinct attribute, as it felt necessary to establish that before diving into the example.

The rest are wording changes that (hopefully) makes the example easier to follow.


reference/features/distinct.md
```diff
@@ -1,16 +1,16 @@
 # Distinct attribute
 
-The **value** of a field whose attribute is set as a distinct attribute will always be **unique** in the returned documents.
+A distinct attribute is a field whose value is shared by various documents. It is most commonly used to prevent MeiliSearch from returning several instances of similar documents.
```
dichotommy, line 3:
```suggestion
The distinct attribute is a special, user-designated field. It is most commonly used to prevent MeiliSearch from returning a set of several similar documents, instead forcing it to return only one.
```

reference/features/distinct.md
```diff
@@ -1,16 +1,16 @@
...
 
-In such a case, there will **never be two, or more, occurrences of the same value** of that field in the different documents returned by MeiliSearch.
+The value of a field configured as a distinct attribute will always be unique among returned documents. This means there will **never be more than one occurrence of the same value** of a distinct attribute field in the different documents returned by MeiliSearch.
```
dichotommy, line 5:
```suggestion
The value of a field configured as a distinct attribute will always be unique among returned documents. This means that among all the different documents returned by MeiliSearch, **there will never be more than one occurrence of the same value attached to the distinct attribute**.
```

dichotommy: This is a tricky concept to describe (but very easy to demonstrate). Not sure if what I've written is clear; feel free to take or leave it.

reference/features/distinct.md
```diff
@@ -1,16 +1,16 @@
...
 
-When multiple documents have the same value for a distinct attribute, MeiliSearch returns the first one after applying [ranking rules](/learn/core_concepts/relevancy.md#ranking-rules). If documents are equivalent in terms of ranking, MeiliSearch returns the first in terms of `internal_id`.
+When multiple documents have the same value for a distinct attribute, MeiliSearch returns the first result after applying [ranking rules](/learn/core_concepts/relevancy.md#ranking-rules). If two or more documents are equivalent in terms of ranking, MeiliSearch returns the first result according to its `internal_id`.
```
dichotommy, line 7:
```suggestion
When multiple documents have the same value for the distinct attribute, MeiliSearch returns only the highest-ranked result after applying [ranking rules](/learn/core_concepts/relevancy.md#ranking-rules). If two or more documents are equivalent in terms of ranking, MeiliSearch returns the first result according to its `internal_id`.
```

reference/features/distinct.md
```diff
@@ -1,16 +1,16 @@
...
 
 ### Example
 
-Suppose you have an e-commerce dataset. For an index that contains information about jackets, you may have several identical items in different variations (color or size).
+Suppose you have an e-commerce dataset. For an index that contains information about jackets, you may have several identical items in different variations such as color or size.
```
dichotommy, line 11:
```suggestion
Suppose you have an e-commerce dataset. For an index that contains information about jackets, you may have several identical items with minor variations such as color or size.
```

reference/features/distinct.md
```diff
@@ -1,16 +1,16 @@
...
 
-As shown below, you have 3 documents that contain information about the same jacket. One of the jackets is brown, one is black, and the last one is blue.
+As shown below, this dataset contains three documents for a Lee jeans leather jacket. One of the jackets is brown, one is black, and the last one is blue.
```
dichotommy, line 13:
```suggestion
As shown below, this dataset contains three documents representing different versions of a Lee jeans leather jacket. One of the jackets is brown, one is black, and the last one is blue.
```

reference/features/distinct.md
```diff
@@ -38,13 +38,15 @@ As shown below, you have 3 documents that contain information about the same jac
 ]
 ```
 
-You may want to ignore the different colors of an item. To do so, you can set `product_id` as a `distinctAttribute`.
+If a user searched for a `lee leather jacket`, MeiliSearch would return all three instances by default. This could cause the results to be cluttered with almost identical variations of the same item.
```
dichotommy, line 41:
```suggestion
By default, a search for `lee leather jacket` would return all three documents. This might not be desired, since displaying nearly identical variations of the same item can make results appear cluttered.
```

reference/features/distinct.md
```diff
@@ -38,13 +38,15 @@ As shown below, you have 3 documents that contain information about the same jac
...
+
+In this case, you may want to ignore the different colors of a product. To do so, you can set `product_id` as a `distinctAttribute`.
```
dichotommy, line 43:
```suggestion
In this case, you may want to return only one document with the `product_id` corresponding to this Lee jeans leather jacket. To do so, you could set `product_id` as the `distinctAttribute`.
```

reference/features/distinct.md
```diff
@@ -38,13 +38,15 @@ As shown below, you have 3 documents that contain information about the same jac
...
 
 <CodeSamples id="distinct_attribute_guide_1" />
 
 By [setting `product_id` as a distinct attribute](/reference/api/distinct_attribute.md), search requests **will never return more than one jacket with the same `product_id`**.
```
dichotommy, line 47:
```suggestion
By setting `distinctAttribute` to `product_id`, search requests **will never return more than one document with the same `product_id`**.
```

reference/features/distinct.md
```diff
@@ -38,13 +38,15 @@ As shown below, you have 3 documents that contain information about the same jac
...
 
-For this example, querying for `lee leather jacket` would only return the first document found. The response could look like this:
+Once distinct attribute is configured, querying for `lee leather jacket` would only return the first document found. The response could look like this:
```
dichotommy, line 49:
```suggestion
After setting the distinct attribute as shown above, querying for `lee leather jacket` would only return the first document found. The response would look like this:
```

dichotommy (changes requested):
Thank you, this makes the article much clearer!

Just a note: we want to be careful about using the word "instance" outside of the context of "a MeiliSearch instance". It could be confusing since that's the primary way this term is used in the docs. Adding this to the style guide, so thanks for that!

**Still needs to be addressed:**

We should link conspicuously to the API reference for distinct attribute at the start and very end of the article.

These changes make the article much clearer; however, I'm still not sure if it's clear that there can be only one `distinctAttribute` per index (unlike similar settings such as `searchableAttributes`, where you can pass an array of attributes). Can we clarify this before merging?

reference/features/distinct.md
guimachiavelli: Yeah, it's difficult to put it in words. I'm not sure either of our attempts make it much clearer, to be honest. 

```suggestion
The value of a field configured as a distinct attribute will always be unique among returned documents. This means **there will never be more than one occurrence of the same value** in the distinct attribute field among the returned documents.
```

dichotommy (approved):
LGTM! 🙌🏻

PR #989, opened 2021-06-16
Feature reference audit: dumps

guimachiavelli:
Minor text edits, mostly adding links to options and API endpoints


reference/features/dumps.md
```diff
@@ -4,30 +4,32 @@ A dump is a compressed file containing an export of your MeiliSearch instance. I
...
 
-Because importing a dump is the same process as when documents are initially indexed by MeiliSearch, it can require some time and memory. If your dataset is very large, it is good practice to index documents in larger batches. This will speed up the indexing process at the cost of memory.
-
-[See here for more dumps options](/reference/features/configuration.md#dumps-destination)
+If your dataset is very large, it is good practice to index documents in larger batches. This will speed up the indexing process, but requires more memory. [You can configure dump batch size with command-line flags and environment variables](/reference/features/configuration#dump-batch-size).
 
 ## Use cases
 
-Dumps are used to restore your database after MeiliSearch updates or to communicate your database to other instances of MeiliSearch (e.g. running on different servers) without having to worry about their respective versions.
+Dumps are used to restore your database after updating MeiliSearch or to communicate between MeiliSearch instances without having to worry about their respective versions.
```
dichotommy, line 35:
```suggestion
Dumps are used to restore your database after [updating MeiliSearch](/create/how_to/updating.md) or to copy your database to other MeiliSearch instances without having to worry about their respective versions.
```

dichotommy (changes requested):
Related to the conversation we were having about API endpoints, I'm not a fan of the way they're displayed on this page, e.g.

> [`GET /dumps/:dump_uid/status`](/reference/api/dump#get-dump-status)

for the following reasons:

1. HTTP API routes can be intimidating/obtuse for non-devs
2. They're not useful for users who work with one of our SDKs, since they should be calling a dedicated function rather than making the HTTP API call manually.

What do you think of changing to the more descriptive format we discussed, e.g.

> the [`get dump status` endpoint](/reference/api/dump#get-dump-status)

guimachiavelli:
I see what you mean and need to think a bit more before I can agree or not…in the meantime, just to explore different perspectives:

- `GET` is definitely intimidating, but it's also precise and hard to miss, right? Like, I know right away I'm going to be taken to the "raw" API when I see this, which means that I can either avoid clicking it (because I know it's currently beyond me, if I'm a newbie) or have a very good idea where I'm going to end up (because I'm a "pro" and know what `GET` means). From a personal perspective as a dev, I prefer clarity;

- is `get dump status endpoint` necessarily more helpful to an SDK user? Would it be interesting to give both the SDK user and the non-SDK user a way to find what they need? E.g. `the get status endpoint or the dump function in your SDK` — I suppose it can be difficult since there's no way to keep track of every SDK and generic terms might not always fit (one SDK might use purely functional code, the other OO, etc)

dichotommy:
@guimachiavelli I would say that
> `get dump status` endpoint
is more useful to an SDK user than
> `GET /dumps/:dump_uid/status`
because it's descriptive about what the API endpoint will accomplish. Granted, this is often self-explanatory (`GET /dumps/.../status` is pretty clear: it gets the dump status), but sometimes it's really not (`DELETE /indexes/:index_uid/settings` does not delete the settings but reset them, for example).

guimachiavelli:
That's a good point on what an endpoint does not being immediately understood just from its path. I mean, it sort of makes sense that using `DELETE` on a setting will reset it to the default value, but it's definitely something you guess rather than something you immediately understand. 

Just to further develop my question regarding the SDK user vs endpoint bit: what do we want an SDK user to get from reading something like `you can use the get dump status endpoint`? They're unlikely to use the endpoint since they're already committed to the SDK, so is there anything we can give them in a situation like this, when we refer to the endpoint directly?

Colouring API links differently sounds like a nice idea, though perhaps it would make more sense to do it when we transition to docusaurus (and checking with a designer so we get something that establishes a coherent visual language). I don't see it done very often, though: the closest example I can find is something like [Stripe's colourcoding the HTTP methods](https://stripe.com/docs/api/products). That is to say, I wonder if there's any reason for it.

dichotommy:
> Just to further develop my question regarding the SDK user vs endpoint bit: what do we want an SDK user to get from reading something like you can use the get dump status endpoint? They're unlikely to use the endpoint since they're already committed to the SDK, so is there anything we can give them in a situation like this, when we refer to the endpoint directly?

Since we're talking about this in the context of linking to an API reference page from outside the API reference section, it would give all users an understanding of the page they're traveling to and the information they'll find there (so they can better make the decision of whether it's relevant to them). From there, an SDK user can click on the tab for their preferred language.

dichotommy (approved):
We can make changes later (or not) based on the result of our discussion around linking to the API references from outside of that section.

Your changes look great.

dichotommy (approved):
LGTM 🙌🏻 🎈

PR #994, opened 2021-06-16
Feature reference audit: field properties

guimachiavelli:
Just three minor tweaks:
- our css doesn't currently support bold links, I think, resulting in only the definite article showing as bold (e.g. `**The** searchable attributes list`);
- reduce repetition of search- (search, searched, searchable) in a sentence;
- clarify where all fields are stored _in the database_.


reference/features/field_properties.md
```diff
@@ -72,6 +72,6 @@ Suppose you manage a database that contains information about movies. By adding
 
 ## Data storing
 
-All fields are stored. **This behavior cannot be changed**.
+All fields are stored in the database. **This behavior cannot be changed**.
 
 Thus, if a field is missing from both the displayed-attributes list and the searchable-attributes list, it **will still be stored**. It will be possible to add it to either or both lists at any time.
```
dichotommy, line 77: > Thus, if a field is missing from both the displayed-attributes list and the searchable-attributes list [...]

This formatting ("the displayed-attributes list" / "the searchable-attributes list") is used throughout this article. It's a bit strange, because the names of these arrays in the settings object are `displayedAttributes` and `searchableAttributes` respectively. The hyphenated spelling used here ("displayed-attributes") only occurs, as far as I can tell, in the API route.

In general, we need to define style rules about when to refer to these settings by their in-engine names (`displayedAttributes`) and when to refer to them as concepts ("the displayed attributes list"). In either case, the hyphenated style here seems incorrect and unhelpful.

Since we are speaking about the list itself, I would argue we should use a more precise + technical style, like so:

> the `displayedAttributes` array

dichotommy (approved):
Thanks for this. We can merge with or without dealing with the problem described in my comment (but I am curious what you think!)

reference/features/field_properties.md
guimachiavelli: yeah, the hyphenated versions did make me raise an eyebrow, but since this was used throughout the article I opted for a conservative approach and leave things as they were.

There's precedence in other articles for using `displayedAttributes` — I'm currently auditing the search parameters and it has a few references to `displayedAttributes list`, for instance.

Perhaps we can start with `displayedAttributes list` in this PR for consistency's sake and change all instances of `list` to `array` in a future PR?

dichotommy: Yes, that makes a lot of sense. Note that `displayedAttributes` and similar (code-cased) words should always be enclosed in a code snippet.

reference/features/field_properties.md
```diff
@@ -72,6 +72,6 @@ Suppose you manage a database that contains information about movies. By adding
 
 ## Data storing
 
-All fields are stored. **This behavior cannot be changed**.
+All fields are stored in the database. **This behavior cannot be changed**.
 
-Thus, if a field is missing from both the displayed-attributes list and the searchable-attributes list, it **will still be stored**. It will be possible to add it to either or both lists at any time.
+Thus, if a field is missing from both the `displayedAttributes` list and the `searchableAttributes` list, it **will still be stored**. It will be possible to add it to either or both lists at any time.
```
dichotommy, line 77:
```suggestion
Thus, even if a field is missing from both the `displayedAttributes` list and the `searchableAttributes` list, **it is still stored in the database** and can be added to either or both lists at any time.
```

dichotommy (approved):
Approving this, while also proposing some extremely minor and mostly aesthetic changes.

dichotommy (approved):
⛵ 
(just to add a little bit of variety to our transportation-related emoji game)

PR #996, opened 2021-06-17
Feature reference audit: known limitations

guimachiavelli:
A few wording suggestions and cutting bits of text that don't add much in the way of actual information.

### DIFF
--- a/reference/features/known_limitations.md
+++ b/reference/features/known_limitations.md
@@ -32,6 +32,6 @@ Currently, MeiliSearch has a number of known limitations. Some of these limitati
 
 ### Payload size
 
-**Limitation:** The default limit for the payload size is around __100MB__.
+**Limitation:** The default limit for the payload size is __~100MB__.
 
-**Explanation:** [This limit can be modified](/reference/features/configuration.md#payload-limit-size).
+**Explanation:** MeiliSearch memory consumption can be an issue in some systems and keeping payload sizes relatively small ensures an optimal experience for most users. [This limit can be modified](/reference/features/configuration.md#payload-limit-size) if you want to take full advantage of particularly powerful hardware.
--- a/reference/features/known_limitations.md
+++ b/reference/features/known_limitations.md
@@ -1,27 +1,24 @@
 # Known limitations
 
-MeiliSearch has known limitations, some of which may be addressed in future versions.
-
-Some of these limitations have been made by MeiliSearch developers for relevancy and/or performance, while others are enforced by [LMDB](/reference/under_the_hood/storage.md), the key-value store used by MeiliSearch.
+Currently, MeiliSearch has a number of known limitations. Some of these limitations are the result of design trade-offs when balancing relevancy and performance, while others are dictated by [LMDB](/reference/under_the_hood/storage.md), the key-value store we use under the hood.
 
 ## Design limitations
 
 ### Number of query words
 
-The maximum number of words taken into account for each [search query](/reference/features/search_parameters.md#query-q) is 10. If a search query includes more than 10 words, all words after the 10th will be ignored.
+The maximum number of terms taken into account for each [search query](/reference/features/search_parameters.md#query-q) is 10. **If a search query includes more than 10 words, all words after the 10th will be ignored.**
 
-This is because large queries can lead to long response times, and MeiliSearch's goal is always to provide a [fast search-as-you-type experience](/learn/what_is_meilisearch/philosophy.md#front-facing-search).
+Large queries can lead to long response times and MeiliSearch's goal is to provide a [fast search-as-you-type experience](/learn/what_is_meilisearch/philosophy.md#front-facing-search).
 
 ### Database size
 
-MeiliSearch uses two databases: the first one for storage and the second one for updates.
-On launch LMDB needs to know the size that it can allocate on disk. This space will be reserved on disk for LMDB, thus MeiliSearch. This space will also be allocated as virtual memory.
+MeiliSearch uses two databases: one for storage and one for updates. On launch, LMDB needs to know the maximum size that it will need to reserve on disk for both of them.
 
-The maximum database size is by default __100GiB__ for each database. This size can be modified using the options `--max-mdb-size` & `--max-udb-size` as described in the [configuration guide](/reference/features/configuration.md#max-mdb-size).
+The default maximum database size is __100GiB__. This size can be modified using the options `--max-mdb-size` & `--max-udb-size` as described in the [configuration guide](/reference/features/configuration.md#max-mdb-size).
 
 ### Number of indexes
 
-You can create __up to 200 indexes__ in MeiliSearch. This limit has been hard set for performance reasons.
+You can create __up to 200 indexes__ in MeiliSearch. This limit has been hard-coded for performance reasons.
 
 ### Maximum words per attribute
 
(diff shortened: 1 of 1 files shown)


reference/features/known_limitations.md
```diff
@@ -1,27 +1,24 @@
 # Known limitations
 
-MeiliSearch has known limitations, some of which may be addressed in future versions.
-
-Some of these limitations have been made by MeiliSearch developers for relevancy and/or performance, while others are enforced by [LMDB](/reference/under_the_hood/storage.md), the key-value store used by MeiliSearch.
+Currently, MeiliSearch has a number of known limitations. Some of these limitations are the result of design trade-offs when balancing relevancy and performance, while others are dictated by [LMDB](/reference/under_the_hood/storage.md), the key-value store we use under the hood.
```
dichotommy, line 3:
```suggestion
Currently, MeiliSearch has a number of known limitations. Some of these limitations are the result of intentional design choices, while others come from [LMDB](/reference/under_the_hood/storage.md), the key-value store MeiliSearch uses [under the hood](/reference/under_the_hood).
```

reference/features/known_limitations.md
```diff
@@ -1,27 +1,24 @@
...
 
 ## Design limitations
 
 ### Number of query words
 
-The maximum number of words taken into account for each [search query](/reference/features/search_parameters.md#query-q) is 10. If a search query includes more than 10 words, all words after the 10th will be ignored.
+The maximum number of terms taken into account for each [search query](/reference/features/search_parameters.md#query-q) is 10. **If a search query includes more than 10 words, all words after the 10th will be ignored.**
 
-This is because large queries can lead to long response times, and MeiliSearch's goal is always to provide a [fast type-as-you-search experience](/learn/what_is_meilisearch/philosophy.md#front-facing-search).
+Large queries can lead to long response times and MeiliSearch's goal is to provide a [fast search-as-you-type experience](/learn/what_is_meilisearch/philosophy.md#front-facing-search).
```
dichotommy, line 11: After your changes, I find this sentence a bit confusing:

> Large queries can lead to long response times

As we just said, there's a limit of 10 query terms max, making it impossible for users to make large queries. But this sentence makes it sound as though it may be possible (but merely discouraged).

This could be solved by adopting a "Limitation: ... Explanation: ..." format, as described in a separate comment.

reference/features/known_limitations.md
```diff
@@ -1,27 +1,24 @@
...
 
 ### Database size
 
-MeiliSearch uses two databases: the first one for storage and the second one for updates.
-On launch LMDB needs to know the size that it can allocate on disk. This space will be reserved on disk for LMDB, thus MeiliSearch. This space will also be allocated as virtual memory.
+MeiliSearch uses two databases: one for storage and one for updates. On launch, LMDB needs to know the maximum size that it will need to reserve on disk for both of them.
 
-The maximum database size is by default __100GiB__ for each database. This size can be modified using the options `--max-mdb-size` & `--max-udb-size` as described in the [configuration guide](/reference/features/configuration.md#max-mdb-size).
+The default maximum database size is __100GiB__. This size can be modified using the options `--max-mdb-size` & `--max-udb-size` as described in the [configuration guide](/reference/features/configuration.md#max-mdb-size).
```
dichotommy, line 17:
```suggestion
The default maximum database size is __100GiB__. This size can be modified using the options `--max-mdb-size` & `--max-udb-size` as described in the [configuration reference](/reference/features/configuration.md#max-mdb-size).
```

dichotommy (commented):
What do you think of adopting a format like this for each section in this article?

```
Limitation: [...]
Explanation: [...]
```

"Explanation" could be replaced with "reason", "reasoning", or "justification".

reference/features/known_limitations.md
guimachiavelli: Hmmm…my interpretation of `large queries` is not that the query _contains_ many terms, but that it _returns_ many documents.

This confusion means that I need to 1) confirm what is the correct interpretation of the original sentence and 2) rephrase it anyway to avoid misunderstandings.

guimachiavelli:
I really like that format idea. Will implement it and re-request your review once it's done.

guimachiavelli:
The limitation/explanation format works so well it made one weird thing suddenly very glaring: every limitation listed in this page presents a very good reason for its existence except `Payload size`. 

For that specific limitation, we just state the default value and say people can change it via a command-line option.

Why is `payload size` here? Should it even be in limitations if a user can simply bump that number to whatever value they need?

@bidoubiwa, do you have any insight on this?

reference/features/known_limitations.md
guimachiavelli: Ignore my previous comment as it makes no sense.

The limitation/explanation format helps, but perhaps we should make this more explicit.

```suggestion
The more terms in a query, the longer it will take MeiliSearch to return matching documents. This goes against our goal of providing a [fast search-as-you-type experience](/learn/what_is_meilisearch/philosophy.md#front-facing-search).
```

dichotommy:
> Why is payload size here? Should it even be in limitations if a user can simply bump that number to whatever value they need?

Valid question. My impression is that large payloads can make the instance more likely to overuse RAM and get killed by the OS. We want to encourage users to batch document addition so we have a default payload limit that's relatively low, but we also want users with powerful hardware to have the flexibility to add documents in much larger batches, hence the ability to change the value.

In other words, the payload size limit itself isn't a limitation, it's more of a "recommendation" to do things in a certain way so as to avoid the real limitation, which is memory-use. It might be worth changing the docs to reflect that, but I'm not sure how public we want to be about the memory consumption problems. 

Perhaps one of the @meilisearch/core-team could evaluate my answer?

reference/features/known_limitations.md
dichotommy: I'm not convinced we should go with this phrasing, since the shortest queries can sometimes take the longest (single-letter queries on large datasets, for example). I think it also probably depends whether you have the `words` ranking rule enabled. This seems like it could be misleading, but then again, I'm no expert! We could discuss it tomorrow in the docs sync meeting.

guimachiavelli: Yeah, you're right. It's a bit of a bold statement which very likely simplifies what actually happens under the hood.

```suggestion
Queries with many search terms can lead to long response times. This goes against our goal of providing a [fast search-as-you-type experience](/learn/what_is_meilisearch/philosophy.md#front-facing-search).
```

reference/features/known_limitations.md
```diff
@@ -1,36 +1,37 @@
...
 
-This limit is enforced for relevancy reasons. The more words there are in a given attribute, the less relevant the search queries will be.
+**Explanation:** This limit is enforced for relevancy reasons. The more words there are in a given attribute, the less relevant the search queries will be.
 
 ## Other limitations
 
 ### Payload size
 
-The default limit for the payload size is around __100MB__. [This limit can be modified](/reference/features/configuration.md#payload-limit-size).
+**Limitation:** The default limit for the payload size is around __100MB__.
```
guimachiavelli, line 35: @dichotommy how do you feel about `around 100MB`? I'm a bit bothered because it feels so imprecise. Although functionally similar, I'm tempted to replace it with `~100MB`.

guimachiavelli:
Thanks, @dichotommy, that was very useful and allowed me to write a brief explanation on what is the rationale of this limitation without going too deep into it.

dichotommy (dismissed):
Looks great, super work!

guimachiavelli:
Had to manually commit a fix for the `large query` bit which rendered the previous review stale.

dichotommy (approved):
🚀

PR #998, opened 2021-06-17
Update settings.md

dichotommy:
rewording. motivation: index settings is an object so I dislike describing it as a list.

### DIFF
--- a/reference/api/settings.md
+++ b/reference/api/settings.md
@@ -1,6 +1,6 @@
 # All settings
 
-Index settings are represented as a [JSON object literal](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/JSON), containing a field for each possible customization option.
+[Index](/learn/core_concepts/indexes.md) settings are represented as a [JSON object literal](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/JSON), containing a field for each possible customization option.
 
 It is possible to modify all the settings at once using the [`update settings` endpoint](#update-settings), or individually using the dedicated routes.
 
--- a/reference/api/settings.md
+++ b/reference/api/settings.md
@@ -1,6 +1,6 @@
 # All settings
 
-Index settings are represented as a [JSON object](https://www.w3schools.com/js/js_json_objects.asp) containing a field for each possible customization option.
+Index settings are represented as a [JSON object literal](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/JSON), containing a field for each possible customization option.
 
 It is possible to modify all the settings at once using the [`update settings` endpoint](#update-settings), or individually using the dedicated routes.
 
--- a/reference/api/settings.md
+++ b/reference/api/settings.md
@@ -1,8 +1,8 @@
 # All settings
 
-`Settings` is a list of all the **customization** possible for an index.
+Index settings are represented as a [JSON object](https://www.w3schools.com/js/js_json_objects.asp) containing a field for each possible customization option.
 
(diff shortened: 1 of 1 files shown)


reference/api/settings.md
```diff
@@ -1,8 +1,8 @@
 # All settings
 
-`Settings` is a list of all the **customization** possible for an index.
+Index settings are represented as a [JSON object](https://www.w3schools.com/js/js_json_objects.asp) containing a field for each possible customization option.
```
guimachiavelli, line 3: I am a bit biased against w3schools — perhaps because when I first started learning HTML/CSS/JS their code samples were riddled with bad practices. I usually prefer to link to the MDN whenever possible, though their pages on JSON go to deep into the subject: [Working with JSON](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/JSON), JSON [reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON), [glossary term](https://developer.mozilla.org/en-US/docs/Glossary/JSON).

Also, I believe calling something a JSON object is technically wrong, since JSON is essentially a string that can be converted into a JS object.

```Suggestion
Index settings are represented as a [JSON object literal](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/JSON), containing a field for each possible customization option.
```

guimachiavelli: The W3Schools page actually warns against saying `JSON object` 😬. We should definitely avoid that expression, especially if we are linking to that page.

Screen Shot 2021-06-21 at 11 40 27

dichotommy: Disagree on this. The [JSON specification](https://www.json.org/json-en.html) (linked to by [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON)) uses the term object explicitly:

> An object is an unordered set of name/value pairs. An object begins with { left brace and ends with } right brace. Each name is followed by : colon and the name/value pairs are separated by , comma.

dichotommy: I agree that linking to w3schools isn't ideal (but then, what is?). Maybe the "No, really, what is JSON?" section of that "Working with JSON" page you linked could be a good alternative? However, I dislike that it includes a lot of unnecessary information (from our perspective) about manipulating JSON with JavaScript.

reference/api/settings.md
```diff
@@ -1,8 +1,8 @@
 # All settings
 
-`Settings` is a list of all the **customization** possible for an index.
+Index settings are represented as a [JSON object literal](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/JSON), containing a field for each possible customization option.
```
guimachiavelli, line 3: Given how index can be a lot of things in programming (e.g. an array index), it could be useful to make this explicit. What do you think about linking to `/learn/core_concepts/indexes.md`?

```suggestion
[Index](/learn/core_concepts/indexes.md) settings are represented as a [JSON object literal](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/JSON), containing a field for each possible customization option.
```

guimachiavelli (dismissed):
This is much better. Only one minor comment regarding using the term index.

reference/api/settings.md
dichotommy: Sure, let's go with this. The other option would be adding a glossary tooltip, but that's not great for accessibility reasons.

guimachiavelli (approved):
LGTM 🏯

dichotommy:
bors merge

PR #1013, opened 2021-06-24
Update storage.md

curquiza:
According to v0.21.0 and the current movies.json (that was changed a bit compared to the last time the page was done).
Plus, the page said only 6.3MB was used in RAM -> that was not true.

### DIFF
--- a/reference/under_the_hood/storage.md
+++ b/reference/under_the_hood/storage.md
@@ -56,11 +56,11 @@ We did some measurements on the default [movies.json](https://github.com/meilise
 
 This dataset is a JSON file of 8.6 MB and has 19,553 documents. When we index this file in MeiliSearch, the amount of disk space taken by LMDB is 122MB.
 
-| Raw JSON | MeiliSearch database size on disk | Real memory size | Private memory size     | Virtual memory size |
-| -------- | --------------------------------- | ---------------- | ----------------------- | ------------------- |
-| 8.6 MB   | 122 MB                            | ≃ 6.3 MB         | 120 MB (≤ size on disk) | 204 Gb (memory map) |
+| Raw JSON | MeiliSearch database size on disk | Resident memory size | Virtual memory size |
+| -------- | --------------------------------- | -------------------- | ------------------- |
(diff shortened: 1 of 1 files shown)


guimachiavelli (dismissed):
looks good to me. Thanks for the PR! 😄 

EDIT: on second thought, I prefer to get @dichotommy's approval before merging on the v0.21 branch

reference/under_the_hood/storage.md
```diff
@@ -56,11 +56,11 @@ We did some measurements on the default [movies.json](https://github.com/meilise
 
 This dataset is a JSON file of 8.6 MB and has 19,553 documents. When we index this file in MeiliSearch, the amount of disk space taken by LMDB is 122MB.
 
-| Raw JSON | MeiliSearch database size on disk | Real memory size | Private memory size     | Virtual memory size |
-| -------- | --------------------------------- | ---------------- | ----------------------- | ------------------- |
-| 8.6 MB   | 122 MB                            | ≃ 6.3 MB         | 120 MB (≤ size on disk) | 204 Gb (memory map) |
+| Raw JSON | MeiliSearch database size on disk | Resident memory size | Virtual memory size |
```
dichotommy, line 59: Would it not be clearer to say `RAM` instead of `Resident memory size`? I know they are not *exactly* the same, but I believe they are functionally the same for almost all purposes.

```suggestion
| Raw JSON | MeiliSearch database size on disk | RAM use | Virtual memory use |
```

dichotommy (approved):
Thanks for updating this :)

reference/under_the_hood/storage.md
dichotommy: I suggest changing `size` to `use` to make a distinction between memory and space. Even though that distinction really doesn't exist in the case of virtual memory, I think it's still helpful for the user :)

curquiza: @Kerollmops, WDYT?

Kerollmops: It seems ok to me, @dichotommy you can find [more information on this StackOverflow message](https://askubuntu.com/a/112407).

PR #1014, opened 2021-06-24
Update snapshots.md

### LINKED ISSUE #999: Question about snapshot feature
From [the doc about snapshot](https://docs.meilisearch.com/reference/features/snapshots.html#creating-snapshots):

> The first snapshot is created on launching MeiliSearch. After that, snapshots are created routinely on a set interval until you deactivate snapshots or end the MeiliSearch instance. By default, one snapshot is taken every 24 hours.

But it does not describe how exactly one could **deactivate snapshot**.

Initially I thought snapshotting was like [dumping](https://docs.meilisearch.com/reference/api/dump.html#create-a-dump), which is accessible via an HTTP API. But it's not. Turns out that snapshotting is actually enabled via a command-line option, and the MeiliSearch instance itself controls it.

So how can we deactivate snapshots without ending the MeiliSearch instance? Is it a planned feature yet to be implemented, or is it just a typo in the doc?

dichotommy:
Fixes #999

### DIFF
--- a/reference/features/snapshots.md
+++ b/reference/features/snapshots.md
@@ -22,7 +22,7 @@ meilisearch --schedule-snapshot=true --snapshot-dir mySnapShots/
 
 Now snapshots are created in `mySnapShots/` directory.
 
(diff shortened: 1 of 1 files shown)


CaroFG (approved):

dichotommy:
bors merge

Issue #784, opened 2021-02-16
Create a telemetry information page


fharper:
We should write about how we use Amplify on its own page on in the FAQ. Which data we retrieve, and how we anonymize it. Once it's done, let's add the link to the page in the starting

fharper:
Information at https://github.com/meilisearch/MeiliSearch/issues/720

react-learner:
Related to #908

PR #1020, opened 2021-06-29
Add Telemetry page

### LINKED ISSUE #908: Create telemetry page
Info on telemetry/analytics is currently located here: https://docs.meilisearch.com/reference/features/configuration.html#analytics

It should be moved to its own page with more detail on what the information is used for, the fact that it is GDPR compliant, the opt-in procedure, etc.

Will probably be added to the new docs website (Docusaurus), not the current one (VuePress).

dichotommy:
First draft of the new telemetry page, to be included in v0.21.

Related to: https://github.com/meilisearch/specifications/blob/main/text/0034-telemetry-policies.md , https://github.com/meilisearch/MeiliSearch/issues/1334 , https://github.com/meilisearch/transplant/issues/180 , #908 #784


.vuepress/config.js
```diff
@@ -117,6 +117,7 @@ module.exports = {
             '/reference/features/snapshots',
             '/reference/features/stop_words',
             '/reference/features/synonyms',
+            '/reference/features/telemetry',
```
dichotommy, line 120: The location of this page isn't super important since we will be restructuring the site during the move to Docusaurus (ETA Q4 2021) but I'm still curious if someone thinks there's a better spot for it :)

reference/features/telemetry.md
```diff
@@ -0,0 +1,50 @@
...
+
+MeiliSearch collects anonymized data from users in order to improve our product. This can be [deactivated at any time](#how-to-disable-data-collection).
+
+Note that while MeiliSearch is running in [`development` mode](/reference/features/configuration.md#environment), no data is collected.
+
+## What kind of data do you collect?
+
+We use two data-collection tools: [Amplitude](https://amplitude.com/amplitude-analytics) and [Sentry](https://sentry.io/welcome/). Both can be disabled by [following the below procedure](#how-to-disable-data-collection).
+
+**Amplitude** collects information about how and where MeiliSearch is being used, including the amount of time a MeiliSearch instance has been running, the version it's using, the device that's running it, and the number of documents per index. It also tracks the number of unique users, which is our equivalent of counting downloads or installs.
```
dichotommy, line 11: Would a bulleted list be better? We could just copy and paste the list from @gmourier 's specification:
```
- Cohort: Servers that are running for more than seven days with the old version of MS (< v0.10.1)
- Cohort: Servers that are running for more than seven days with the new version of MS (> v0.10.1)
- Cohort: Servers that are running for more than 30 days with the old version of MS (< v0.10.1)
- Cohort: Servers that are running for more than 30 days with the new version of MS (> v0.10.1)
- Graph: The total of every telemetry messages received
- Graph: The total of every unique id seen
- Graph: The number of servers that are running at least since 7 days
- Graph: The number of servers that are running at least since 30 days
- User Composition: Device composition
- User Composition: Version composition
- Indexes size distribution for more than seven days (<200k not taken into account)
- Indexes size distribution for the others
```

reference/features/telemetry.md
```diff
@@ -0,0 +1,50 @@
...
+
+Note that this data is aggregated, meaning that we never view individual statistics.
```
dichotommy, line 13: I'm not 100% sure this is true, can anyone confirm?

dichotommy: Also `statistics` is probably the wrong word 😢

reference/features/telemetry.md
gmourier: If we choose to go towards a bullet list (which is a good idea I think) we should be able to simplify it. The specification describes internal details at the Amplitude level on how we organize the metrics. Not sure if the details of Cohort, Graph etc.. are useful for the reader. In fact, I think the four lines about cohorts can be removed because the lines 
```
- Graph: The number of servers that are running at least since 7 days 
- Graph: The number of servers that are running at least since 30 days
```
are identical in terms of information for the end user.

If I'm not mistaken (@qdequele, please correct me if I'm wrong), a cohort is in fact a segment made up of "users" (MeiliSearch instance in our case) who share the same traits (e.g., active for more than 30 days or active for more than 7 days).  We can feed different tables or graphs to compare several cohorts.

reference/features/telemetry.md
gmourier: > Also `statistics` is probably the wrong word 😢

Does the term `metrics` can be a better fit?

gmourier: > I'm not 100% sure this is true, can anyone confirm?

In fact, we have the possibility to have access to a "user" (instance) page, this is what Amplitude calls ["User Look-up"](https://help.amplitude.com/hc/en-us/articles/229313067-User-Look-Up#User-Profile).

In practice, the data we visualize will always be aggregated data. It is possible to see custom fields values in the page of a "user", in our case, what I call custom fields are the number of days of use, or the version of MeiliSearch. These custom fields are then used to segment the "user" in a specific cohort.

This page does not allow us to identify a user with the data that we store.

TLDR; We can view individual data per "user" (instance) despite the fact that we don't store anything that allows us to identify a person or a group of people. That specific view doesn't bring anything of value in its current state for the core product.

gmourier:
Thanks, @dichotommy for this brand new documentation page! 🙏

A few little feedbacks!

If I remember we found the term Analytics more accurate than Telemetry because Telemetry is more of a remote data collection technique while Analytics seems to be more precise on how it is used.

I think it's important to have a section dedicated to the Why in order to make our readers understand our objective behind it. To resume it our goal behind that Analytics page is to have quantitative data in order to improve our product to have a maximum positive impact to solve their needs. 

What do you think about it?

reference/features/telemetry.md
dichotommy: Understood, thank you very much @gmourier for the clarification! 😄

dichotommy:
> If I remember we found the term Analytics more accurate than Telemetry because Telemetry is more of a remote data collection technique while Analytics seems to be more precise on how it is used.

My opinion has flipped on this. The primary reason I now prefer "Telemetry" over "Analytics" is that "Analytics" could suggest a feature we're offering to the user (analytics of their search history) rather than describing the data we collect from them. I think "Telemetry" is more clear on this point, and have noticed it being used consistently on other documentation sites ([Stripe](https://stripe.com/docs/cli/telemetry), [Gatsby](https://www.gatsbyjs.com/docs/telemetry/), [Algolia](https://telemetry.algolia.com)).

> I think it's important to have a section dedicated to the Why in order to make our readers understand our objective behind it.

You're absolutely right, I will add this! Thanks again for the feedback ❤️

gmourier:
> > If I remember we found the term Analytics more accurate than Telemetry because Telemetry is more of a remote data collection technique while Analytics seems to be more precise on how it is used.
> 
> My opinion has flipped on this. The primary reason I now prefer "Telemetry" over "Analytics" is that "Analytics" could suggest a feature we're offering to the user (analytics of their search history) rather than describing the data we collect from them. I think "Telemetry" is more clear on this point, and have noticed it being used consistently on other documentation sites ([Stripe](https://stripe.com/docs/cli/telemetry), [Gatsby](https://www.gatsbyjs.com/docs/telemetry/), [Algolia](https://telemetry.algolia.com)).
> 

I completely agree with this opinion!

.vuepress/config.js
curquiza: I think the place you choose is nice.
Or maybe in the dropdown of `Resources`?

curquiza (approved):
I like this page!!!
Is it possible to add a question in the FAQ? It might be something people would naturally ask

gmourier (approved):
LGTM 🥳

dichotommy:
Great idea @curquiza , thanks!

reference/features/telemetry.md
```diff
@@ -0,0 +1,76 @@
...
+- Technical specs of the device running MeiliSearch
+- Number of documents per index
+
+**Sentry** collects information about errors that occur while using MeiliSearch—in other words, bug reports and diagnostics.
+
+**We will absolutely never identify users, track users, or collect private information**, including IP address, email, or website URLs. All of our data collection is **GDPR-compliant** and **secure**.
+
+::: tip
+
+In the future, we may collect additional metrics relating to new features. To have the most up-to-date info, be sure to check the release notes after each new version. <Badge text="soon" type="warn"/>
```
guimachiavelli, line 32: This feels like we're saying "we've created this page and will henceforth never touch it again."

```suggestion
In the future, we may collect additional metrics relating to new features. We will update this page if that happens, but be sure to check the release notes after each new version in case we lag a bit behind. <Badge text="soon" type="warn"/>
```

.vuepress/config.js
guimachiavelli: nitpicking: is telemetry an actual feature? It feels a bit out-of-place to me because it is, at best, a necessary evil for users. As in, we have legitimate reasons for collecting data, but it's not something that directly benefits them in the same way e.g. snapshots do.

I prefer @curquiza's suggestion. Alternatively, perhaps `learn/what_is_meilisearch/`, just above `contact us`?

reference/features/telemetry.md
```diff
@@ -0,0 +1,76 @@
...
+
+Note that while MeiliSearch is running in [`development` mode](/reference/features/configuration.md#environment), no data is collected.
+
+## Why collect telemetry data?
+
+We collect anonymized data for only two reasons: so that we can improve our product, and so that we can continue working on this project full-time.
+
+In order to better address our users' needs and chart a clear path forward for MeiliSearch, we need reliable quantitative information. The information we collect helps us fix bugs, evaluate the success of certain features, and identify our strengths and weaknesses as a product.
+
+Usage metrics also help us justify our existence to investors so that we can keep the MeiliSearch project alive. They prove what we all believe: that MeiliSearch is a great product that can be of use to a large number of people around the world.
```
guimachiavelli, line 13: The paragraph on how we use the data we collect is convincing, but the one on the investors feels a bit muddier, as if it's going roundabout instead of saying `we need to prove people are actually using MS`.

reference/features/telemetry.md
```diff
@@ -0,0 +1,76 @@
...
+
+## What kind of data do we collect?
```
guimachiavelli, line 15: Would it be better to say **what** we collect first and **why** we collect it later? I'm thinking a user might care more about what is being collected than why. 

Personally I will never believe a company when they say they only collect data for benign purposes and assume they will eventually sell it to someone; hence I would be more interested in seeing what I will be giving them.

reference/features/telemetry.md
```diff
@@ -0,0 +1,76 @@
...
+```
+
+:::
+
+::: tab Cloud service
+
+```bash
+# The following procedure should work for DO, GCP, and AWS
+# First, open /etc/systemd/system/meilisearch.service with a text editor
+# vim /etc/systemd/system/meilisearch.service
```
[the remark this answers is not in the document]
dichotommy: Ah right, I forgot `nano` displays basic commands at the bottom of the window. You're right, that's better.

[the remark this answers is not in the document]
dichotommy: Do you think I should uncomment the `nano /etc/systemd/system/meilisearch.service` line? It makes the rest less copy+pasteable, but that way no one will miss it.

.vuepress/config.js
dichotommy: Okay, thanks for the feedback. I will take a look at where other doc sites tend to stick their telemetry files, because all of these options seem to have tradeoffs.

dichotommy: Okay, thanks for the feedback. I will take a look at where other doc sites tend to stick their telemetry files. All of these options seem to have tradeoffs, so I'd rather follow convention (if there is one).

dichotommy: Ended up deciding to go with @guimachiavelli 's suggestion to place it in `/learn/what_is_meilisearch/`. It's not an ideal long-term resting place, since it's not a key aspect of the product that we want to show off to first-time-users. However, considering that this issue has caused some distrust in the community, I feel that putting it front and center like this is strangely fitting. Transparency forever, I guess ✌🏻

guimachiavelli (approved):
LGTM! 🎈

PR #924, opened 2021-05-05
Prepare documentation for v0.21

dichotommy:
# v0.21

This PR will include all changes to the documentation coming in v0.21.

### https://github.com/meilisearch/transplant/issues/85 Changes to Ranking Rules
- [x] Fix in Relevancy
- [x] Fix in Ranking Rules API ref
- [x] PR created: https://github.com/meilisearch/documentation/pull/974
- [x] PR merged

### https://github.com/meilisearch/specifications/pull/34 https://github.com/meilisearch/transplant/issues/180 New telemetry page
- [x] Create new docs page detailing MeiliSearch default telemetry, explanation for why we take this data, and instructions for opting in/out
- [x] PR created https://github.com/meilisearch/documentation/pull/1020
- [x] PR merged

### https://github.com/meilisearch/specifications/blob/main/text/0043-phrase-query.md New feature: Phrase Search
- [x] Add mentions of phrase search in search parameters feature reference, search API reference
- [x] PR created https://github.com/meilisearch/documentation/pull/1021
- [x] PR merged

### https://github.com/meilisearch/transplant/issues/81 https://github.com/meilisearch/transplant/issues/140 Changes to Facets + Filters
- `attributesForFaceting` renamed to `filterableAttributes`
- `facetFilters` parameter no longer exists
- `filters` parameter changed to `filter`, and now works for facets as well but requires fields to be defined in `filterableAttributes`
- `facetsDistribution` no longer returns zero values (see: https://github.com/meilisearch/transplant/issues/234)
- `filter` can use both logical and array syntax
- [x] PR created: https://github.com/meilisearch/documentation/pull/1041
- [x] Create new Filtering and faceted search feature ref
- [x] Remove Filtering feature ref
- [x] Remove Faceted Search feature ref
- [x] Fix in Search Parameters feature ref
- [x] Fix in Settings feature ref
- [x] Fix in attributesForFaceting API ref
- [x] Add redirects from `attributesForFaceting` to `filterableAttributes`
- [x] Add warning: `exhaustiveFacetsCount` is not currently implemented and always returns `false`
- [x] PR merged

### https://github.com/meilisearch/transplant/issues/107 Addition of new update status value `processing`
New full list of update statuses:
> enqueued: the update is enqueued and will start to be processed soon
**> processing: the update is processing.**
> processed: the update is processed and it succeeded.
> failed: A failure occurred during when processing.

- [x] create PR #1046 
- [x] review `learn/advanced/asynchronous_updates.md` (see issue #902)
[...]

PR #1021, opened 2021-07-01
Add phrase search

guimachiavelli:
Updates learn/what_is_meilisearch/features, reference/api/search, and reference/features/search_parameters so they all briefly mention phrase search and how to use it.

### DIFF
--- a/learn/what_is_meilisearch/features.md
+++ b/learn/what_is_meilisearch/features.md
@@ -44,6 +44,12 @@ If you make a search without inputting any query words, MeiliSearch will return 
 
 Placeholder search is not affected by MeiliSearch's [default ranking rules](/reference/features/settings.md#ranking-rules)—only custom rules added by a user. If no custom rules have been set, the results are displayed in the order of their internal database position.
 
+## Phrase search
+
+If you enclose search terms in double quotes (`"`), MeiliSearch will only return documents that  contain all terms in the order they were given.
+
+This is particularly useful when searching for documents that can only have one relevant result, like a book's ISBN code. It is also possible to combine phrase searches with MeiliSearch's basic syntax so only parts of a query are matched strictly.
+
 ## Comprehensive language support
 
 [MeiliSearch is multilingual](/reference/features/language.md)! We aim to support every language represented in our global community.
--- a/learn/what_is_meilisearch/features.md
+++ b/learn/what_is_meilisearch/features.md
@@ -48,7 +48,7 @@ Placeholder search is not affected by MeiliSearch's [default ranking rules](/ref
 
 If you enclose search terms in double quotes (`"`), MeiliSearch will only return documents that contain all terms in the order they were given.
 
-This is particularly useful when searching for documents that can only have one relevant result, like a book's ISBN code. It is also possible to combine phrase searches with MeiliSearch's basic syntax so only parts of a query are matched strictly.
+Phrase search is particularly useful when looking for a single result, such as searching for a book by ISBN. It is also possible to combine phrase searches with MeiliSearch's basic syntax so only parts of a query are matched strictly.
 
 ## Comprehensive language support
 
--- a/learn/what_is_meilisearch/features.md
+++ b/learn/what_is_meilisearch/features.md
@@ -46,7 +46,7 @@ Placeholder search is not affected by MeiliSearch's [default ranking rules](/ref
 
 ## Phrase search
 
-If you enclose search terms in double quotes (`"`), MeiliSearch will only return documents that  contain all terms in the order they were given.
+If you enclose search terms in double quotes (`"`), MeiliSearch will only return documents that contain all terms in the order they were given.
 
 This is particularly useful when searching for documents that can only have one relevant result, like a book's ISBN code. It is also possible to combine phrase searches with MeiliSearch's basic syntax so only parts of a query are matched strictly.
 
--- a/learn/what_is_meilisearch/features.md
+++ b/learn/what_is_meilisearch/features.md
@@ -46,7 +46,7 @@ Placeholder search is not affected by MeiliSearch's [default ranking rules](/ref
 
 ## Phrase search
 
-If you enclose search terms in double quotes (`"`), MeiliSearch will only return documents that contain all terms in the order they were given.
+If you enclose search terms in double quotes (`"`), MeiliSearch will only return documents that contain those terms in the order they were given. This gives users the option to make more precise search queries.
 
 Phrase search is particularly useful when looking for a single result, such as searching for a book by ISBN. It is also possible to combine phrase searches with MeiliSearch's basic syntax so only parts of a query are matched strictly.
 
--- a/reference/api/search.md
+++ b/reference/api/search.md
@@ -133,6 +133,10 @@ This route should only be used when no API key is required. If an API key is req
 
 When the `q` parameter is not specified, a [placeholder](/reference/features/search_parameters.md#query-q) search is run instead.
 
+#### Phrase search
+
+Query terms enclosed in double quotes are parsed as [phrase searches](/reference/features/search_parameters.md#query-q).
+
 ### Response
 
 | field                | Description                    |    type    |
--- a/reference/api/search.md
+++ b/reference/api/search.md
@@ -43,11 +43,11 @@ This is the preferred route to perform search when an API key is required, as it
 
 #### Placeholder search
 
-When the `q` parameter is not specified, a [placeholder](/reference/features/search_parameters.md#query-q) search is run instead.
+When no search query is specified, a [placeholder search](/reference/features/search_parameters.md#query-q) is run instead.
 
 #### Phrase search
 
-Query terms enclosed in double quotes are parsed as [phrase searches](/reference/features/search_parameters.md#query-q).
+Query terms enclosed in double quotes are treated as [phrase searches](/reference/features/search_parameters.md#query-q).
 
 ### Response
 
@@ -139,11 +139,11 @@ This route should only be used when no API key is required. If an API key is req
 
 #### Placeholder search
 
-When the `q` parameter is not specified, a [placeholder](/reference/features/search_parameters.md#query-q) search is run instead.
+When no search query is specified, a [placeholder search](/reference/features/search_parameters.md#query-q) is run instead.
 
 #### Phrase search
 
-Query terms enclosed in double quotes are parsed as [phrase searches](/reference/features/search_parameters.md#query-q).
+Query terms enclosed in double quotes are treated as [phrase searches](/reference/features/search_parameters.md#query-q).
 
 ### Response
 
--- a/reference/api/search.md
+++ b/reference/api/search.md
@@ -41,6 +41,14 @@ This is the preferred route to perform search when an API key is required, as it
 
 [Learn more about how to use the search parameters](/reference/features/search_parameters.md).
 
+#### Placeholder search
+
+When the `q` parameter is not specified, a [placeholder](/reference/features/search_parameters.md#query-q) search is run instead.
+
+#### Phrase search
+
+Query terms enclosed in double quotes are parsed as [phrase searches](/reference/features/search_parameters.md#query-q).
+
 ### Response
 
 | field                | Description                    |    type    |
--- a/reference/features/search_parameters.md
+++ b/reference/features/search_parameters.md
@@ -44,7 +44,7 @@ Placeholder search returns the documents that best match other search parameters
 
 ### Phrase search
 
-You can enclose search terms in double quotes (`"`), in which case MeiliSearch will only return documents containing all terms in the order they were given. This is called a [phrase search](/learn/what_is_meilisearch/features.md#phrase-search).
+If you enclose search terms in double quotes (`"`), MeiliSearch will only return documents containing those terms in the order they were given. This is called a **phrase search**.
 
 Phrase searches are case-insensitive and ignore [soft separators such as `-`, `,`, and `:`](/reference/under_the_hood/datatypes.md). Hard separators within a phrase search effectively split the query into different phrase searches: `"Octavia.Butler"` will return the same results as `"Octavia" "Butler"`.
 
--- a/reference/features/search_parameters.md
+++ b/reference/features/search_parameters.md
@@ -48,7 +48,7 @@ If you enclose search terms in double quotes (`"`), MeiliSearch will only return
 
 Phrase searches are case-insensitive and ignore [soft separators such as `-`, `,`, and `:`](/reference/under_the_hood/datatypes.md). Using a hard separator within a phrase search effectively splits it into multiple separate phrase searches: `"Octavia.Butler"` will return the same results as `"Octavia" "Butler"`.
 
-You can mix phrase searches with regular queries: MeiliSearch will first fetch all documents with exact matches and [then proceed with its default behavior](/learn/core_concepts/relevancy.md).
+You can combine phrase search and normal queries in a single search request. In this case, MeiliSearch will first fetch all documents with exact matches to the given phrase(s), and [then proceed with its default behavior](/learn/core_concepts/relevancy.md).
 
 #### Example
 
--- a/reference/features/search_parameters.md
+++ b/reference/features/search_parameters.md
@@ -46,7 +46,7 @@ Placeholder search returns the documents that best match other search parameters
 
 If you enclose search terms in double quotes (`"`), MeiliSearch will only return documents containing those terms in the order they were given. This is called a **phrase search**.
 
-Phrase searches are case-insensitive and ignore [soft separators such as `-`, `,`, and `:`](/reference/under_the_hood/datatypes.md). Hard separators within a phrase search effectively split the query into different phrase searches: `"Octavia.Butler"` will return the same results as `"Octavia" "Butler"`.
+Phrase searches are case-insensitive and ignore [soft separators such as `-`, `,`, and `:`](/reference/under_the_hood/datatypes.md). Using a hard separator within a phrase search effectively splits it into multiple separate phrase searches: `"Octavia.Butler"` will return the same results as `"Octavia" "Butler"`.
 
 You can mix phrase searches with regular queries: MeiliSearch will first fetch all documents with exact matches and [then proceed with its default behavior](/learn/core_concepts/relevancy.md).
 
--- a/reference/features/search_parameters.md
+++ b/reference/features/search_parameters.md
@@ -42,6 +42,12 @@ When `q` isn't specified, a **placeholder search** is performed. Placeholder sea
 
 Placeholder search returns the documents that best match other search parameters, sorted according to that index's [ranking rules](/reference/features/settings.md#ranking-rules). This feature also supports [faceting](/reference/features/faceted_search.md) and [filtering](/reference/features/filtering.md)
 
+### Phrase search
+
+You can enclose search terms in double quotes (`"`), in which case MeiliSearch will only return documents containing all terms in the order they were given. This is called a [phrase search](/learn/what_is_meilisearch/features.md#phrase-search).
+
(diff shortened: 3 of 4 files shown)


reference/features/search_parameters.md
```diff
@@ -42,6 +42,12 @@ When `q` isn't specified, a **placeholder search** is performed. Placeholder sea
 
 Placeholder search returns the documents that best match other search parameters, sorted according to that index's [ranking rules](/reference/features/settings.md#ranking-rules). This feature also supports [faceting](/reference/features/faceted_search.md) and [filtering](/reference/features/filtering.md)
 
+### Phrase search
+
+You can enclose search terms in double quotes (`"`), in which case MeiliSearch will only return documents containing all terms in the order they were given. This is called a [phrase search](/learn/what_is_meilisearch/features.md#phrase-search).
+
+Phrase searches are case-insensitive and ignore soft separators such as `-`, `,`, and `:`. You can mix them with regular queries: MeiliSearch will first fetch all documents with exact matches and [then proceed with its default behavior](/learn/core_concepts/relevancy.md).
```
curquiza, line 49:
```suggestion
Phrase searches ignore soft separators such as `-`, `,`, and `:`. You can mix them with regular queries: MeiliSearch will first fetch all documents with exact matches and [then proceed with its default behavior](/learn/core_concepts/relevancy.md).
```

guimachiavelli: Thanks for the input, @curquiza! Does this mean we are deviating from the spec? 

For reference: [phrase query spec, usage heading](https://github.com/meilisearch/specifications/blob/main/text/0043-phrase-query.md#phrase-query-usage)

learn/what_is_meilisearch/features.md
```diff
@@ -44,6 +44,12 @@ If you make a search without inputting any query words, MeiliSearch will return
 
 Placeholder search is not affected by MeiliSearch's [default ranking rules](/reference/features/settings.md#ranking-rules)—only custom rules added by a user. If no custom rules have been set, the results are displayed in the order of their internal database position.
 
+## Phrase search
```
guimachiavelli, line 47: @gmourier: I have preferred `phrase search` over `phrase query`, is that ok?

reference/features/search_parameters.md
curquiza: Indeed @ManyTheFish can you confirm the mistake?
FYI @gmourier

ManyTheFish: Phrase searches are case-insensitive.

reference/features/search_parameters.md
```diff
@@ -42,6 +42,12 @@ When `q` isn't specified, a **placeholder search** is performed. Placeholder sea
...
+
```
ManyTheFish, line 50: We spoke about soft separators, but hard separators are handled differently. When we have a Hard separator in a phrase search, we split the phrase into 2 different phrases.
`hello "world! Nice to see you"` would be considered as `hello "world" "Nice to see you"`.
Because in meilisearch, a Hard separator is considered as a split of 2 unrelated contexts.

reference/features/search_parameters.md
curquiza: Sorry! I confused case-sensitive and case-insensitive!!!! My bad everyone! 😂 I definitely need holidays!

learn/what_is_meilisearch/features.md
gmourier: I have used both until then! I have no particular input, the two seem good to me! I will update the feature name on the public roadmap. Thanks!

learn/what_is_meilisearch/features.md
```diff
@@ -44,6 +44,12 @@ If you make a search without inputting any query words, MeiliSearch will return
...
+
+If you enclose search terms in double quotes (`"`), MeiliSearch will only return documents that  contain all terms in the order they were given.
```
gmourier, line 49:
```suggestion
If you enclose search terms in double quotes (`"`), MeiliSearch will only return documents that contain all terms in the order they were given.
```

reference/features/search_parameters.md
gmourier: Should we link to the full list of soft separators?

gmourier (changes requested):

reference/features/search_parameters.md
guimachiavelli: I thought about that, but couldn't find an in-house resource. Do you have a good external resource we could link to?

reference/features/search_parameters.md
guimachiavelli: Thanks for that, @ManyTheFish — I'll add an extra sentence about hard separators.

reference/features/search_parameters.md
gmourier: We have a documentation page in the "Under the hood" section that describes our soft and hard separators https://docs.meilisearch.com/reference/under_the_hood/datatypes.html but it does not describe the full list.

Here the source code https://github.com/meilisearch/tokenizer/blob/0db4b8fca263bfc89cb4c4453877a95e63806a73/src/detection.rs#L49

@ManyTheFish Please tell me if I'm wrong about the file 😇. Also, I didn't see the `—` char we identified some time ago. We had added an issue but it seems to me that there was a mix-up do you recall?

guimachiavelli: @gmourier thanks for the link! 

Is there a reason why the datatypes page doesn't list all separators? Do we use different lists of soft/hard separators depending on what we're doing?

If not, I'll open an issue and update it.

bidoubiwa:
Because SDK's may have a different way to write `" "` in a string it could be interesting to have a code-sample showcasing a phrase search. See [this comment](https://github.com/meilisearch/documentation/pull/924#issuecomment-874075450). For example in [GO](https://www.programmersought.com/article/62454930882/) where string have to be wrapper in double quotes. 

I'm not sure it requires to add the response body though.

guimachiavelli:
@bidoubiwa I was thinking about that, actually. Wouldn't adding a code example make this section a bit longer than it should? Is that even an actual problem? Thoughts, @dichotommy ?

reference/features/search_parameters.md
```diff
@@ -42,6 +42,18 @@ When `q` isn't specified, a **placeholder search** is performed. Placeholder sea
 
 Placeholder search returns the documents that best match other search parameters, sorted according to that index's [ranking rules](/reference/features/settings.md#ranking-rules). This feature also supports [faceting](/reference/features/faceted_search.md) and [filtering](/reference/features/filtering.md)
 
+### Phrase search
+
+You can enclose search terms in double quotes (`"`), in which case MeiliSearch will only return documents containing all terms in the order they were given. This is called a [phrase search](/learn/what_is_meilisearch/features.md#phrase-search).
```
dichotommy, line 47:
```suggestion
If you enclose search terms in double quotes (`"`), MeiliSearch will only return documents containing those terms in the order they were given. This is called a **phrase search**.
```
Feels weird to link to a less detailed section on the same subject :)

reference/features/search_parameters.md
```diff
@@ -42,6 +42,18 @@ When `q` isn't specified, a **placeholder search** is performed. Placeholder sea
...
+
+Phrase searches are case-insensitive and ignore [soft separators such as `-`, `,`, and `:`](/reference/under_the_hood/datatypes.md). Hard separators within a phrase search effectively split the query into different phrase searches: `"Octavia.Butler"` will return the same results as `"Octavia" "Butler"`.
```
dichotommy, line 49:
```suggestion
Phrase searches are case-insensitive and ignore [soft separators such as `-`, `,`, and `:`](/reference/under_the_hood/datatypes.md). Using a hard separator within a phrase search effectively splits it into multiple separate phrase searches: `"Octavia.Butler"` will return the same results as `"Octavia" "Butler"`.
```

reference/features/search_parameters.md
```diff
@@ -42,6 +42,18 @@ When `q` isn't specified, a **placeholder search** is performed. Placeholder sea
...
+
+You can mix phrase searches with regular queries: MeiliSearch will first fetch all documents with exact matches and [then proceed with its default behavior](/learn/core_concepts/relevancy.md).
```
dichotommy, line 51:
```suggestion
You can combine phrase search and normal queries in a single search request. In this case, MeiliSearch will first fetch all documents with exact matches to the given phrase(s), and [then proceed with its default behavior](/learn/core_concepts/relevancy.md).
```

dichotommy (changes requested):
Thanks for writing this 🔥

learn/what_is_meilisearch/features.md
```diff
@@ -44,6 +44,12 @@ If you make a search without inputting any query words, MeiliSearch will return
 
 Placeholder search is not affected by MeiliSearch's [default ranking rules](/reference/features/settings.md#ranking-rules)—only custom rules added by a user. If no custom rules have been set, the results are displayed in the order of their internal database position.
 
+## Phrase search
+
+If you enclose search terms in double quotes (`"`), MeiliSearch will only return documents that contain all terms in the order they were given.
```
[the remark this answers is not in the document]
guimachiavelli: MInor thing, but `freedom` here gives me marketing-speak vibes. What about `option` instead?

```suggestion
If you enclose search terms in double quotes (`"`), MeiliSearch will only return documents that contain those terms in the order they were given. This gives users the option to make more precise search queries.
```

reference/api/search.md
```diff
@@ -133,6 +133,10 @@ This route should only be used when no API key is required. If an API key is req
 
 When the `q` parameter is not specified, a [placeholder](/reference/features/search_parameters.md#query-q) search is run instead.
 
+#### Phrase search
+
+Query terms enclosed in double quotes are parsed as [phrase searches](/reference/features/search_parameters.md#query-q).
+
```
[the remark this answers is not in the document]
guimachiavelli: What do you think about having this content repeated on both POST and GET? I'm not usually a fan of duplicated content, but having it on only one might give the impression that the two routes do not offer the same functionality.

reference/under_the_hood/concat.md
```diff
@@ -26,7 +26,7 @@ This consists of finding the most interesting place to separate the words and to
 
 This is achieved by finding the best frequency of the separate words in the dictionary of all words in the dataset. It will look out that both words have a minimum of interesting results, and not just one of them.
 
-Split words are not considered as multiple words in a search query because they must stay next to each other, this behavior is called **Phrase Queries**.
+Split words are not considered as multiple words in a search query because they must stay next to each other, this behavior is called [phrase search](/reference/features/search_parameters.md#phrase-search).
```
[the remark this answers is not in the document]
guimachiavelli: You're absolutely right, monkey brain went "this is not right…in this house we say phrase searches!"

reference/api/search.md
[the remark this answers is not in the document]
dichotommy: Yes sounds good, we're just going to delete GET eventually anyway so no harm in repeating it.

reference/api/search.md
```diff
@@ -41,6 +41,14 @@ This is the preferred route to perform search when an API key is required, as it
 
 [Learn more about how to use the search parameters](/reference/features/search_parameters.md).
 
+#### Placeholder search
+
+When the `q` parameter is not specified, a [placeholder](/reference/features/search_parameters.md#query-q) search is run instead.
```
dichotommy, line 46:
```suggestion
When no search query is specified, a [placeholder search](/reference/features/search_parameters.md#query-q) is run instead.
```

reference/api/search.md
```diff
@@ -133,6 +141,10 @@ This route should only be used when no API key is required. If an API key is req
 
 When the `q` parameter is not specified, a [placeholder](/reference/features/search_parameters.md#query-q) search is run instead.
```
dichotommy, line 142:
```suggestion
When no search query is specified, a [placeholder search](/reference/features/search_parameters.md#query-q) is run instead.
```

reference/api/search.md
dichotommy: Not relevant to this PR, but I want to make this change since `q` parameter isn't really exposed to users of the SDKs (they have dedicated search functions).

reference/api/search.md
```diff
@@ -41,6 +41,14 @@ This is the preferred route to perform search when an API key is required, as it
 
 [Learn more about how to use the search parameters](/reference/features/search_parameters.md).
 
+#### Placeholder search
+
+When the `q` parameter is not specified, a [placeholder](/reference/features/search_parameters.md#query-q) search is run instead.
+
+#### Phrase search
+
+Query terms enclosed in double quotes are parsed as [phrase searches](/reference/features/search_parameters.md#query-q).
```
dichotommy, line 50:
```suggestion
Query terms enclosed in double quotes are treated as [phrase searches](/reference/features/search_parameters.md#query-q).
```

reference/api/search.md
```diff
@@ -133,6 +141,10 @@ This route should only be used when no API key is required. If an API key is req
 
 When the `q` parameter is not specified, a [placeholder](/reference/features/search_parameters.md#query-q) search is run instead.
 
+#### Phrase search
+
+Query terms enclosed in double quotes are parsed as [phrase searches](/reference/features/search_parameters.md#query-q).
```
dichotommy, line 146:
```suggestion
Query terms enclosed in double quotes are treated as [phrase searches](/reference/features/search_parameters.md#query-q).
```

reference/features/search_parameters.md
```diff
@@ -42,6 +42,18 @@ When `q` isn't specified, a **placeholder search** is performed. Placeholder sea
 
 Placeholder search returns the documents that best match other search parameters, sorted according to that index's [ranking rules](/reference/features/settings.md#ranking-rules). This feature also supports [faceting](/reference/features/faceted_search.md) and [filtering](/reference/features/filtering.md)
 
+### Phrase search
+
+If you enclose search terms in double quotes (`"`), MeiliSearch will only return documents containing those terms in the order they were given. This is called a **phrase search**.
+
+Phrase searches are case-insensitive and ignore [soft separators such as `-`, `,`, and `:`](/reference/under_the_hood/datatypes.md). Using a hard separator within a phrase search effectively splits it into multiple separate phrase searches: `"Octavia.Butler"` will return the same results as `"Octavia" "Butler"`.
+
+You can combine phrase search and normal queries in a single search request. In this case, MeiliSearch will first fetch all documents with exact matches to the given phrase(s), and [then proceed with its default behavior](/learn/core_concepts/relevancy.md).
```
dichotommy, line 51: I'm having second thoughts about this last sentence, not sure it adds anything. What do you think of getting rid of it?

```suggestion
You can combine phrase search and normal queries in a single search request.
```

dichotommy (approved):
LGTM! All of my comments are optional.

I'm still figuring out the merging order of all of our open PRs, but I think this one certainly must be merged after Search Parameters.

Mostly unrelated thought: we really need to clean up the descriptions of placeholder search and phrase search in `/learn/what_is_meilisearch/features.md` to make them shorter and less technical, more in line with the other descriptions on that page.

reference/features/search_parameters.md
guimachiavelli: I'm leaning towards leaving it because this explains exactly why the example (`"african american" poet`) works.

dichotommy: Hmm, I don't completely understand but it's probably just the sleep deprivation talking.

guimachiavelli: `"African American" poet` works well because it matches `African American` strictly , but uses the regular MeiliSearch process for `poet`, meaning it will match also match words like `poem` (possibly also `poetic` and `poetry`? Not 100% sure on those two, to be honest).

ManyTheFish: Hey @guimachiavelli! In your example `poet` has less than 5 characters, meaning that we don't allow a typo on this word 😅 but if the word would be longer we would allow 1 typo on this word.

Meanwhile, because `poet` is not in the `".."`, meilisearch will return documents where `poet` isn't near `African American`.

dichotommy:
It just occurred to me that I have no clue how phrase search interacts with the `words` ranking rule. Is a phrase `".."` treated as a single `word`? For those who aren't familiar with the rule, that would mean that order matters, so `"african american" poet` might return results in a different order than `poet "african american"`.

An extension of this question: can a mixed query return results that don't contain the given phrase at all? e.g., a search for `poet "african american"` could return documents containing only `poet` if there were no documents containing `african american`, `african-american`, etc. in the dataset, and the `words` ranking rule was used.

guimachiavelli:
@dichotommy that's a very good question. If the sentence currently on line 51 is correct (and I think it is but am happy to be corrected otherwise), phrase searches are always given priority over MeiliSearch's normal behaviour. 

Testing this with v0.21 RC1 seems to support that. If I search for `batman "shifu"` (which are more or less mutually exclusive given that there's very low chance of intersection there), I only get two hits, both of them containing only `shifu` but no `batman`.

reference/features/search_parameters.md
guimachiavelli: @ManyTheFish I stand corrected! Thanks for that!

ManyTheFish:
> It just occurred to me that I have no clue how phrase search interacts with the `words` ranking rule. Is a phrase `".."` treated as a single `word`? For those who aren't familiar with the rule, that would mean that order matters, so `"african american" poet` might return results in a different order than `poet "african american"`.
> 
> An extension of this question: can a mixed query return results that don't contain the given phrase at all? e.g., a search for `poet "african american"` could return documents containing only `poet` if there were no documents containing `african american`, `african-american`, etc. in the dataset, and the `words` ranking rule was used.

@dichotommy @guimachiavelli, in milli we have an intermediate step between the `query parsing` and the `bucket sort`:

## The `QueryTree` generation ✨ 

In this part, we "list" all the possibilities in a Tree: each word with or without typos, ngrams, split words...
This `QueryTree` will be interpreted by each criterion to fetch documents.
So the `QueryTree` rules how the criteria work giving a "search playground".

### The rules
The `QueryTree` is able to [rules the words](https://youtu.be/jlm4QqNAPiE?t=50):
- Typo: we allow typos on the word
- Exact: we forbid  typos on the word
- Prefix: we allow the word to be considered as a prefix

The `QueryTree` is able to define rules on branches (+or- a group of words):
- OR: we allow any document that contains one of the words
- AND: we allow any document that contains all the words
- PHRASE: we allow any document containing all the words side by side without typos or prefixes

This last rule will force criteria to interpret phrase search differently.

## example (sorry)
### with `"african american" poet`

```yaml
OR(WORD)
  PHRASE ["african", "american"]
  AND
    PHRASE ["african", "american"]
    PrefixExact { word: "poet" }
```

### with `african american poet`
```yaml
OR(WORD)
  Tolerant { word: "african", max typo: 1 }
  OR
    AND
      Tolerant { word: "african", max typo: 1 }
      Tolerant { word: "american", max typo: 1 }
    Tolerant { word: "africanamerican", max typo: 2 }
  OR
    AND
      Tolerant { word: "african", max typo: 1 }
      OR
        AND
          Tolerant { word: "american", max typo: 1 }
          PrefixExact { word: "poet" }
        PrefixTolerant { word: "americanpoet", max typo: 2 }
    AND
      Tolerant { word: "africanamerican", max typo: 2 }
      PrefixExact { word: "poet" }
    PrefixTolerant { word: "africanamericanpoet", max typo: 2 }
```

As we can see, the `QueryTree` let fewer choices to the criteria 🤓

I hope it was clear 😅, don't hesitate to ping me if you have more question ☺️

gmourier:
@ManyTheFish Thank you for this explanation it is much clearer for me about the inner workings. Seeing that, I understand that the number of typos increases according to the size of the word?

ManyTheFish:
@gmourier, yes if the word length:
- is lower than 5 bytes, no typos are allowed
- is higher or equal to 5 bytes and lower than 9 bytes, 1 typo is allowed
- is higher or equal to 9, 2 typos are allowed

PR #1031, opened 2021-07-08
Replace dashboard medias

mdubus:
## What's inside this PR ? 

Replacement of a few medias still showing the old mini-dashboard.

## How to test ?

Go to the following routes : 
- http://localhost:8080/learn/getting_started/quick_start.html#web-interface
- http://localhost:8080/learn/what_is_meilisearch/features.html#synonyms
- http://localhost:8080/reference/features/web_interface.html


mdubus:
@bidoubiwa do you know why the `.vuepress/components/MovieGif.vue` component has a fixed height ? Because it's kind of messing up the page if my gif doesn't have the same size as the previous one 🤔 
What do you think about replacing the fixed heights with an `auto` height ?

guimachiavelli:
I believe that you need to specify the dimensions because videos can be very fiddly when it comes to sizing. Perhaps using a CSS proportional box would help? We'd still have to keep a consistent ratio between every image, so not sure it actually helps that much.

bidoubiwa:
Hey @mdubus, I have trouble reproducing the effect that you are describing. Could you show me in which circumstances it is weird? 

> Because it's kind of messing up the page if my gif doesn't have the same size as the previous one

I used fixed size based on the proportion of the gifs I made, so if you would like to use the same solution (which may really not be the best) you'll have to do the same calculation.

In your suggestion of putting a fixed weight, do you mean to do it with the `@media` query? Or a global fixed size? 

> I believe that you need to specify the dimensions because videos can be very fiddly when it comes to sizing.

@guimachiavelli, I think if we use a fixed width and an auto height, like @mdubus suggests, we keep the proportions without using javascript. 

What would be the counterparts of @mdubus solution? @mdubus could you maybe apply it to this PR so we can try it out?

mdubus:
Hi ! 
Some clarification here : 

The problem can be viewed on [this page](http://localhost:8080/learn/getting_started/quick_start.html#search) (there are a lot of white space between the text and the GIF): 

Capture d’écran 2021-07-12 à 09 41 58

And here in blue, the size taken by the div, and in red the real size of the GIF: 
Capture d’écran 2021-07-12 à 09 46 06



On the [features pages](http://localhost:8080/learn/what_is_meilisearch/features.html#synonyms), the GIF has the correct size. By inspecting the element, we can see that the GIF has just a `max-width: 100%`: 
Capture d’écran 2021-07-12 à 09 53 05

And in the markdown of the page, I see that we use only the GIF as an image `search demo gif` and not the `MovieGif` component. 

So, 3 propositions here: 
- We can keep the fixed width of the component, and replace the fixed height with an `height: auto`
- We can remove the height of the component, and replace the fixed width with a `max-width: 100%`
- We can remove the `MovieGif` (he is the only one to use the video files, but they seem to be useless: if I remove them in the code, everything still works fine !) and use only the markdown syntax with the path to the gif.

What do you think @bidoubiwa @guimachiavelli  ?

bidoubiwa:
The component is not only used for CSS. It is used to add the following HTML to every page. This template is specific to how `video` tag works in HTML. 
The browser will search for compatibility with `webm` and if it does not find any he will go to the following which is `mp4`. The jpg image is used as thumbnail of the video.  

```
<video class="video" autoplay="true" poster="truemen.jpg" loop muted>
      <source src="/movie-gifs/trumen-fast.webm" type="video/webm">
      <source src="/movie-gifs/trumen-fast.mp4" type="video/mp4">
    </video
```

Videos are way less heavy than GIF and should be preferred to avoid unnecessary download for the user. 

The gif in synonyms should be changed to a video as well.


So the only two options left are: 
- We can keep the fixed width of the component, and replace the fixed height with an height: auto
- We can remove the height of the component, and replace the fixed width with a max-width: 100%

What do you recommend @mdubus? I think you prefer the first option, if you do could you update this PR with a working version of your preferred solution?

bidoubiwa:
Could you remove one BR because now there is a difference in vertical margin between the upper paragraph and the below title.

bidoubiwa:
Also I think the mobile version was better in the previous version of the doc. 


## Previous version of Doc

Screenshot 2021-07-12 at 12 26 30

Screenshot 2021-07-12 at 12 26 24

## Your PR

Screenshot 2021-07-12 at 12 27 25

Screenshot 2021-07-12 at 12 27 19

bidoubiwa:
After further investigation the problem appears to come from the mobile preview precisely from the chrome dev tools. The problems does not appear to occur on real mobile devices and different browsers.

bidoubiwa (changes requested):
None related code is present in `quick_start.md`

bidoubiwa (approved):
LGTM 🔥

PR #1049, opened 2021-07-29
Update OAS file to match v0.21

gmourier:
- Remove OAS definition from the documentation repository
- Remove UI building from the documentation repository
- Updates links to point to bump.sh lastest version and OAS version to 3.1.0

@guimachiavelli Unfortunately swaggerUI is not able to handle the `3.1.0` OAS version format at the moment. I tried to update the swagger dists to the latest stable version to check if `3.1.0` could be generated with swaggerUI which provides a test interface. https://github.com/swagger-api/swagger-ui/issues/5891

Bump.sh doesn't have a live test interface yet. By merging this PR we have an updated OAS interface with 0.21 but we lose the possibility to test the API from the web.

I'm looking for a simple workaround to keep this swaggerUI interface for live testing.

Edit: I found a workaround by forcing the version of the open-api file to be `3.0.3` instead of `3.1.0` but the version generated by swaggerUI no longer offers all the examples since the 3.1.0 format changed it and it forces us to maintain 2 versions just to be able to keep swaggerUI.

According to the stats I have this page is very little visited, if ever.


guimachiavelli (approved):
LGTM 🎂 

I will merge this into the prepare v0.21 branch so we deploy it only once everything's ready.

PR #1050, opened 2021-07-29
Fix permalink on / for faq & open-api-readme

tpayet:
When on /, if I click on resources > faq with right-click and open new tab I have a 404.

This seems to fix this behavior. Let me know if I should open this PR against master or the v021 branch


guimachiavelli (approved):
Looks great, thanks for fixing this! 🦜 

Will merge it into v0.21 instead of master, if that's alright with you.

PR #1055, opened 2021-08-04
v0.21: searchable attributes

guimachiavelli:
Manual updates to `searchableAttributes` change the order of the returned documents' fields.

- [x] reference/api/searchable_attributes: add warning  (L#12?)
- [x] reference/features/field_properties: add warning  (L#48?)

### DIFF
--- a/reference/api/searchable_attributes.md
+++ b/reference/api/searchable_attributes.md
@@ -43,7 +43,7 @@ List the settings.
 Update the [searchable attributes](/reference/features/field_properties.md#searchable-fields) of an index.
 
 ::: warning
-Manually updating `searchableAttributes` will change the order of the returned documents' fields.
+Manually updating `searchableAttributes` will change the displayed order of document fields in the JSON response.
 :::
 
 #### Path variables
--- a/reference/api/searchable_attributes.md
+++ b/reference/api/searchable_attributes.md
@@ -43,7 +43,7 @@ List the settings.
 Update the [searchable attributes](/reference/features/field_properties.md#searchable-fields) of an index.
 
 ::: warning
-Manually updating `searchableAttributes` will change the displayed order of document fields in the JSON response.
+Due to an implementation bug, manually updating `searchableAttributes` will change the displayed order of document fields in the JSON response. This behavior is inconsistent and will be fixed in a future release.
 :::
 
 #### Path variables
--- a/reference/api/searchable_attributes.md
+++ b/reference/api/searchable_attributes.md
@@ -42,6 +42,10 @@ List the settings.
 
 Update the [searchable attributes](/reference/features/field_properties.md#searchable-fields) of an index.
 
+::: warning
+Manually updating `searchableAttributes` will change the order of the returned documents' fields. This will be fixed in a future MeiliSearch release.
+:::
+
 #### Path variables
 
 | Variable      | Description   |
--- a/reference/api/searchable_attributes.md
+++ b/reference/api/searchable_attributes.md
@@ -43,7 +43,7 @@ List the settings.
 Update the [searchable attributes](/reference/features/field_properties.md#searchable-fields) of an index.
 
 ::: warning
-Manually updating `searchableAttributes` will change the order of the returned documents' fields. This will be fixed in a future MeiliSearch release.
+Manually updating `searchableAttributes` will change the order of the returned documents' fields.
 :::
 
 #### Path variables
--- a/reference/features/field_properties.md
+++ b/reference/features/field_properties.md
@@ -45,7 +45,7 @@ You may want to make some attributes non-searchable, or change the [attribute ra
 After manually updating the `searchableAttributes` list, **subsequent new attributes will no longer be automatically added** unless the settings are [reset](/reference/api/searchable_attributes.md#reset-searchable-attributes).
 
 ::: warning
-Manually updating `searchableAttributes` will change the order of the returned documents' fields.
+Manually updating `searchableAttributes` will change the displayed order of document fields in the JSON response.
 :::
 
 #### Example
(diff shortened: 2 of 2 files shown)


reference/api/searchable_attributes.md
```diff
@@ -42,6 +42,10 @@ List the settings.
 
 Update the [searchable attributes](/reference/features/field_properties.md#searchable-fields) of an index.
 
+::: warning
+Manually updating `searchableAttributes` will change the order of the returned documents' fields.
```
curquiza, line 46:
```suggestion
Manually updating `searchableAttributes` will change the displayed order of the returned documents' fields in the JSON response.
```

curquiza: Maybe my sentence is too long, I just don't want the users to think it will change another kind of order, like the attribute order for the relevancy

reference/features/field_properties.md
```diff
@@ -42,8 +42,10 @@ If you'd like to restore your searchable attributes list to this default behavio
 
 You may want to make some attributes non-searchable, or change the [attribute ranking order](/learn/core_concepts/relevancy.md#attribute-ranking-order) after documents have been indexed. To do so, simply place the attributes in the desired order and send the updated list using the [update searchable attributes endpoint](/reference/api/searchable_attributes.md#update-searchable-attributes).
 
+After manually updating the `searchableAttributes` list, **subsequent new attributes will no longer be automatically added** unless the settings are [reset](/reference/api/searchable_attributes.md#reset-searchable-attributes).
+
 ::: warning
-Be aware that after manually updating the `searchableAttributes` list, subsequent new attributes will no longer be automatically added unless the settings are [reset](/reference/api/searchable_attributes.md#reset-searchable-attributes).
+Manually updating `searchableAttributes` will change the order of the returned documents' fields.
```
curquiza, line 48: Same here

reference/api/searchable_attributes.md
guimachiavelli: This is great, just a minor alteration
```suggestion
Manually updating `searchableAttributes` will change the displayed order of document fields in the JSON response.
```

reference/features/field_properties.md
guimachiavelli:
```suggestion
Manually updating `searchableAttributes` will change the displayed order of document fields in the JSON response.
```

curquiza (approved):

reference/api/searchable_attributes.md
```diff
@@ -42,6 +42,10 @@ List the settings.
 
 Update the [searchable attributes](/reference/features/field_properties.md#searchable-fields) of an index.
 
+::: warning
+Manually updating `searchableAttributes` will change the displayed order of document fields in the JSON response.
```
CaroFG, line 46: For me, it's not clear in which manner the order will be affected. Attributes are going to be displayed in the order specified in the `searchableAttributes` list?

reference/features/field_properties.md
```diff
@@ -42,8 +42,10 @@ If you'd like to restore your searchable attributes list to this default behavio
 
 You may want to make some attributes non-searchable, or change the [attribute ranking order](/learn/core_concepts/relevancy.md#attribute-ranking-order) after documents have been indexed. To do so, simply place the attributes in the desired order and send the updated list using the [update searchable attributes endpoint](/reference/api/searchable_attributes.md#update-searchable-attributes).
 
+After manually updating the `searchableAttributes` list, **subsequent new attributes will no longer be automatically added** unless the settings are [reset](/reference/api/searchable_attributes.md#reset-searchable-attributes).
+
 ::: warning
-Be aware that after manually updating the `searchableAttributes` list, subsequent new attributes will no longer be automatically added unless the settings are [reset](/reference/api/searchable_attributes.md#reset-searchable-attributes).
+Manually updating `searchableAttributes` will change the displayed order of document fields in the JSON response.
```
CaroFG, line 48: In this case I also wonder what prevails: the order in the `searchableAttributes` list or the order in the `displayedAttributes` one?

reference/api/searchable_attributes.md
guimachiavelli: I think you are correct, but I'm not sure whether that's information we want to emphasise because this is a bug (or at least an unintended side-effect of other changes) which we will fix. Eventually. Meaning, even if it always follows the `searchableAttributes` order, I'm not sure we want people to rely on it or expect this will always be the case.

What do you think?

CaroFG: Well, even if it's a bug and it will change later, as a user I want to know what to expect if I update them manually... When we release another version we will update the information accordingly

reference/features/field_properties.md
guimachiavelli: This is a bug related only to `searchableAttributes`. `displayedAttributes` has no effect on field order.

reference/api/searchable_attributes.md
guimachiavelli:
```suggestion
Due to an implementation bug, manually updating `searchableAttributes` will change the displayed order of document fields in the JSON response. This behavior wields inconsistent results and will be fixed in a future release.
```

CaroFG: Oh, I didn't mean saying explicitly it is a bug, but explaining how is the order affected.  I was thinking more of something like: `Manually updating `searchableAttributes` will make the document fields in the JSON response appear in the same order of the searchableAttributes' list.` It's a horribly written sentence, but I am not inspired ATM 😅

guimachiavelli: After speaking to @curquiza about this, although (from a quick test, at least) it seems like that is the case, we don't fully understand the mechanism behind this behaviour at the moment (or how to fix it). 

Since we can't describe it with 100% certainty, my goal here is to put a few red signs. Like, "this is a bug and you shouldn't rely on it".

reference/features/field_properties.md
guimachiavelli:
```suggestion
Due to an implementation bug, manually updating `searchableAttributes` will change the displayed order of document fields in the JSON response. This behavior is inconsistent and will be fixed in a future release.
```

PR #1060, opened 2021-08-09
Update .code-samples.meilisearch.yaml

CaroFG:
According to the docs `If using the GET route to perform a search, all parameters must be URL-encoded`, plus this code sample isn't working (the `--data` part is not being passed at all), that's why I am changing it to a POST request


guimachiavelli (approved):
Looks great, thanks so much 🤗 

I will incorporate this into prepare-v0.21 instead of master to avoid merge conflicts.

PR #1065, opened 2021-08-17
Remove wrong telemetry sentence

curquiza:
This will come in v0.22.0 😇

PR #1067, opened 2021-08-18
Fix missing telemetry nodes

gmourier:


learn/what_is_meilisearch/telemetry.md
```diff
@@ -17,7 +17,8 @@ We use two data-collection tools: [Amplitude](https://amplitude.com/amplitude-an
 - Amount of time the MeiliSearch instance has been running
 - MeiliSearch version
 - Technical specs of the device running MeiliSearch
-- Number of documents per index
+- Database size per instance
+- Number of documents per instance
```
guimachiavelli, line 21: According to @curquiza's PR #1065, `Number of documents per index` will only be available on v0.22.

Is `Number of documents per instance` available on v0.21? Or is the proposed addition only correcting exchanging `index` with `instance` but refer to the same field?

gmourier: Hey @guimachiavelli!

An instance can contain several indexes.

So the @curquiza fix is ok! In 0.21 we already have `Number of documents per instance`.

guimachiavelli: Got it, thanks for clarifying!

guimachiavelli (approved):
Looks great, thanks for this! 🧨

PR #1074, opened 2021-08-24
 Documentation changes for v0.22

guimachiavelli:
Preliminary list of changes required to make the docs up-to-date with v0.22:

## Telemetry page
_Issue reference:_ https://github.com/meilisearch/MeiliSearch/issues/1592
_Spec reference:_ https://github.com/meilisearch/specifications/blob/main/text/0034-telemetry-policies.md

- [ ] learn/what_is…/telemetry: add new collected data
- [ ] Create PR
- [ ] Request review
- [ ] Merge

## Indexer
Issue reference:  https://github.com/meilisearch/MeiliSearch/issues/1591
Spec reference: N/A

- [ ] Identify references to memory crashes due to heavy RAM usage by indexer
- [ ] reference/under_the_hood/storage: review page

## Sort
Issue reference:  https://github.com/meilisearch/MeiliSearch/issues/1581
Spec reference: https://github.com/meilisearch/specifications/blob/develop/text/0055-sort.md

- [ ] reference/features/search_parameters: add `sort` parameter
- [ ] reference/api/settings: add `sortableAttribute`
- [ ] learn/what_is…/telemetry: add new measurements
- [ ] Create PR
- [ ] Audit PR
- [ ] Request review
- [ ] Address feedback
- [ ] Merge

## New file formats: CSV and NDJSON
Issue reference:  https://github.com/meilisearch/MeiliSearch/issues/1593
Spec reference: https://github.com/meilisearch/specifications/blob/develop/text/0028-indexing-csv.md and https://github.com/meilisearch/specifications/blob/develop/text/0029-indexing-ndjson.md

- [ ] Identify places where we mention file formats 
- [ ] Clarify we CSV and NDJSON are now accepted

_Note: this PR is a duplicate of #1066, which was automatically closed when we merged #924._

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -1,6 +1,6 @@
 const ogprefix = 'og: http://ogp.me/ns#'
 module.exports = {
-  title: 'MeiliSearch Documentation v0.21',
+  title: 'MeiliSearch Documentation v0.22',
   description: 'Open source Instant Search Engine',
   themeConfig: {
     repo: 'meilisearch/MeiliSearch',


curquiza:
## New indexer

❤️ Only suggestions ❤️

### Add a question in FAQ: "MeiliSearch crashes when I try to add documents".

This is due to the RAM consumption. It means your machine size might be not adapted to the size of your dataset. We recommend increasing the size of your machine in term of RAM. This is a known problem the core team always try to improve.

(In the future, another solution could be to push using CSV or NDJSON format)

###  Add a question in FAQ: "MeiliSearch is really slow when adding documents"

There is two ways to increase the speed of the indexation:
- The indexation is RAM consuming and multi-threaded: the bigger is the RAM and the number of cores, and the faster will be the indexation.
- We always recommend pushing your documents to MeiliSearch by batch and not one by one. For example, if you have a dataset containing 100 000 documents you can try to push it in two batches of 50 000. The API routes to add or update documents are designed to push several documents at the same time. The bigger is the HTTP payload (so the batch of your document), the faster will be the indexation.
Also, the maximum payload size of MeiliSearch is 100Mb and [can be optionally increased](https://docs.meilisearch.com/reference/features/configuration.html#payload-limit-size). It means most of the dataset can be pushed with one call to the API. However, if the size of your machine is too small in term of RAM, it might lead MeiliSearch to crash. Cf question above.

(In the future, another solution could be to push using CSV or NDJSON format)

### Add a section (maybe "Under the hood") to explain the indexation

The MeiliSearch indexation is RAM consuming and multi-threaded.

1. RAM

MeiliSearch detects the memory size of the machine and adapts its behavior to optimize the indexation by taking a maximum of 2/3 of the RAM of the machine. This prevents from the crashes without over-consuming the RAM. Also, we always want to keep some RAM to perform the search, that's why we don't use 100% of the RAM.
If MeiliSearch cannot guess the size of the RAM, it still adapts the behavior but some crashes could happen. It's impossible to guess the RAM in rare cases for specific OS. For example, FreeBSD is not supported yet, see the opened issue in the sysinfo lib we use to detect the RAM.
Here is the exhaustive list of the supported OS (in the Rust docs of sysinfo): https://docs.rs/sysinfo/0.20.0/sysinfo/#supported-oses

Some crashes can still happen even if MeiliSearch can guess the OS. This means the size of your machine is not adapted at to the size of your dataset. See this question in the FAQ.

2. Multi-thread

MeiliSearch's indexation is multi-threaded and used half the available cores. If the machine has 12 cores, only 6 are used. For machines having only 1 core, of course, this single core is used. This way, there are always cores available for the search: using all the cores during indexation would mean disable the search during indexation.

.code-samples.meilisearch.yaml
```diff
@@ -424,3 +431,61 @@ phrase_search_1: |-
...
+    -X DELETE 'http://localhost:7700/indexes/books/settings/sortable-attributes'
+search_parameter_guide_sort_1: |-
+  curl \
+    -X POST 'http://localhost:7700/indexes/books/search' \
+    --data '{
+      "q": "science fiction",
+      "sort": [
+        "price:asc"
+      ]
+    }'
```
curquiza, line 491: You should not put a new line at the end of the file. It does not impact any usage, but it's a style rule to avoid unrelated changes between two contributors in the future PRs

To avoid this kind of error, you can download the editorconfig plugin, depending on your text editor. After downloading it, every time you will save your file, it will automatically apply the basic style rules to your file (removing trailing whitespaces, adding a new line at the end of file...) thanks to the .editorconfig file in this repo.
See more about editorconfig: https://editorconfig.org/#overview
And the editorconfig file of this repo: https://github.com/meilisearch/documentation/blob/master/.editorconfig

reference/under_the_hood/storage.md
```diff
@@ -58,10 +58,16 @@ This dataset is a JSON file of 8.6 MB and has 19,553 documents. When we index th
 
 | Raw JSON | MeiliSearch database size on disk | RAM usage | Virtual memory usage |
 | -------- | --------------------------------- | -------------------- | ------------------- |
-| 9.1 MB   | 320 MB                            | ≃ 426 MB             | 205 Gb (memory map) |
+| 9.1 MB   | 224 MB                            | ≃ 305 MB             | 205 Gb (memory map) |
 
 > This means the database is using 426 MB of RAM and 122 MB of disk space. Note that [virtual memory](https://www.enterprisestorageforum.com/hardware/virtual-memory/) refers only to disk space allocated by your computer for MeiliSearch—it does not mean that it's actually in use by the database. See [Memory Usage](#memory-usage) for more details.
```
curquiza, line 63:
```suggestion
> This means the database is using 305 MB of RAM and 224 MB of disk space. Note that [virtual memory](https://www.enterprisestorageforum.com/hardware/virtual-memory/) refers only to disk space allocated by your computer for MeiliSearch—it does not mean that it's actually in use by the database. See [Memory Usage](#memory-usage) for more details.
```

curquiza: oops my fault, I forgot it in my PR

.code-samples.meilisearch.yaml
guimachiavelli: ah, sorry! I got carried away in the rush of things because we have a markdown lint rule that enforces new lines at the end of every .md (which is something I really want to change but haven't had the time so far).

Anyway, thanks, I'll take care of it =)

curquiza (approved):
🚀 

What a doc!!!

Issue #953, opened 2021-06-01
Make Quick Start's JS code samples consistent


guimachiavelli:
[The last section of the Quick Start tutorial](https://docs.meilisearch.com/learn/getting_started/quick_start.html#integrate-with-your-project) contains three examples of integrations between Meilisearch and vanilla JS, Vue and React. 

This is great, but each sample is following slightly different coding styles. It's a minor thing, but it might be good to keep things consistent.

The three main discrepancies I noticed:
- `<html>` has a `lang` attribute on vanilla JS, but none in React and Vue;
- `<script>` tags are inserted within the `<body>` on vanilla JS, but outside it in React and Vue;
- `<head>` outside `<html>` on React. I'm sure browsers understand it just fine, but the HTML spec dictates that `<head>` must be a child of `<html>`.

What do you think @bidoubiwa?

PR #1084, opened 2021-08-26
Add curl header authentification example

bidoubiwa:
related to: #953


guimachiavelli (approved):
LGTM 🔥 

Thanks for this =)

PR #1087, opened 2021-08-26
v0.22: Sort

guimachiavelli:
v0.22 adds a new feature, Sort, which allows users to change search result order at query time. This can be useful when a user might want to, e.g., see an eshop's products in ascending or descending price order.

Previously, this functionality could be awkwardly replicated by creating several indexes with custom ranking rules.

Reference spec: https://github.com/meilisearch/specifications/blob/develop/text/0055-sort.md
Issue on core: https://github.com/meilisearch/MeiliSearch/issues/1581

Tasks:

- [x] Create new reference/features page: Sorting
- [x] reference/features/search_parameters: add `sort` parameter
- [x] reference/api/settings: add `sortableAttribute`
- [x] lear/what_is…/features: add 2 sentences explaining Sorting
- [x] learn/core_concepts/relevancy: add `sort` to default ranking rules
- [x] reference/api/ranking_rules: add `sort` to default ranking rules


curquiza:
I don't know if it's planned for this PR, but don't forget the new ranking rule syntax (cf the spec)

reference/api/sortable_attributes.md
```diff
@@ -0,0 +1,98 @@
...
+
+Attributes that can be used together with the [`sort` search parameter](/reference/features/search_parameters.md#sort). [You can learn more about sorting in our dedicated guide.](/reference/features/sorting.md)
+
+Sortable attributes can also be updated through the [global settings route](/reference/api/settings.md#update-settings).
+
+## Get sortable attributes
+
+<RouteHighlighter method="GET" route="/indexes/:index_uid/settings/sortable-attributes" />
+
+Get an index's [`sortableAttributes`](/reference/features/sorting.md).
```
curquiza, line 13:
```suggestion
Get the index's [`sortableAttributes`](/reference/features/sorting.md).
```

curquiza: Since `sortableAttributes` is plural, but I'm not sure!

reference/features/sorting.md
```diff
@@ -0,0 +1,200 @@
...
+
+To allow your users to sort results at search time you must:
+
+1. Decide which attributes you want to use for sorting
+2. Add those attributes to the `sortableAttributes` index setting
+3. Update MeiliSearch's [ranking rules](/learn/core_concepts/relevancy.md) (optional)
+
+### Select attributes for sorting
+
+MeiliSearch allows you to sort results based on document fields. Only fields  containing numbers or strings can be used for sorting.
```
curquiza, line 21:
```suggestion
MeiliSearch allows you to sort results based on document fields. Only fields containing numbers or strings can be used for sorting.
```

learn/core_concepts/relevancy.md
```diff
@@ -97,11 +107,11 @@ The `attribute` rule sorts the results by [attribute importance](/learn/core_con
...
-To add a custom ranking rule, you have to communicate either `asc` for ascending order or `desc` for descending order followed by the field name in parentheses.
+To add a custom ranking rule, you have to communicate the attribute name followed by a colon (`:`) and either `asc` for ascending order or `desc` for descending order.
 
-- To apply an **ascending sort** (results sorted by increasing value of the attribute): `asc(attribute_name)`
+- To apply an **ascending sort** (results sorted by increasing value of the attribute): `attribute_name:asc`
 
-- To apply a **descending sort** (results sorted by decreasing value of the attribute): `desc(attribute_name)`
+- To apply a **descending sort** (results sorted by decreasing value of the attribute): `attribute_name:desc`
 
 **The attribute must have a numeric value** in all of the documents contained in that index. **If any value is not a numeric type, the sorting rule won't be applied**.
```
curquiza, line 116: Not true anymore, you can apply custom ranking rules on string :)

curquiza (changes requested):

alallema:
This one also needs an `update_settings_1: |-` in the `.code-samples.meilisearch.yaml` ❤️

learn/core_concepts/relevancy.md
```diff
@@ -114,35 +124,51 @@ Let's say you have a movie dataset. The documents contain the fields `release_da
...
-  "asc(release_date)",
-  "desc(movie_ranking)"
+  "release_date:asc",
+  "movie_ranking:desc"
 ]
 ```
 
+### Sorting and custom ranking rules
+
+MeiliSearch allows users to define [sorting order at query time](/reference/features/sorting.md) by using the [`sort` search parameter](/reference/features/search_parameters.md#sort). There is some overlap between sorting and ranking rules, but the two do have different uses.
```
gmourier, line 153:
```suggestion
MeiliSearch allows users to define [sorting order at query time](/reference/features/sorting.md) by using the [`sort` search parameter](/reference/features/search_parameters.md#sort). There is some overlap between sorting and custom ranking rules, but the two do have different uses.
```

reference/api/settings.md
```diff
@@ -40,11 +41,12 @@ Get the settings of an index.
...
+| **filterableAttributes**  | [Strings] | Attributes to use as [filters and facets](/reference/features/filtering_and_faceted_search.md)         | `[]`                                                                         |
 | **distinctAttribute**     | String    | Search returns documents with distinct (different) values of the given field     | `null`                                                                                            |
 | **searchableAttributes**  | [Strings] | Fields in which to search for matching query words sorted by order of importance | `["*"]` (all attributes)                                                                          |
-| **displayedAttributes**   | [Strings] | Fields displayed in the returned documents                                       | `["*"]` (all attributes)                                                                                         |
+| **displayedAttributes**   | [Strings] | Fields displayed in the returned documents                                       | `["*"]` (all attributes)                                                                          |
```
gmourier, line 48:
```suggestion
| **displayedAttributes**   | [Strings] | Fields allowed to be displayed in the returned documents                                     | `["*"]` (all attributes)                                                                          |
```

reference/features/search_parameters.md
```diff
@@ -22,6 +22,7 @@ This is not necessary when using the `POST` route or one of our [SDKs](/learn/wh
 | **[cropLength](/reference/features/search_parameters.md#crop-length)**                        | Maximum field value length                         | `200`         |
 | **[attributesToHighlight](/reference/features/search_parameters.md#attributes-to-highlight)** | Highlight matching terms contained in an attribute | `null`        |
 | **[matches](/reference/features/search_parameters.md#matches)**                               | Return matching terms location                     | `false`       |
+| **[sort](/reference/features/search_parameters.md#sort)**                                     | Sort queries by an attribute's value                        | `null`        |
```
gmourier, line 25:
```suggestion
| **[sort](/reference/features/search_parameters.md#sort)**                                     | Sort search results by an attribute's value                        | `null`        |
```

reference/features/sorting.md
```diff
@@ -0,0 +1,200 @@
...
+
+::: warning
+If a field has values of different types across documents, MeiliSearch will give precedence to numbers over strings. This means documents with numeric field values will be ranked higher than those with string values.
+
+This can lead to unexpected behavior when sorting, so we strongly recommend you only allow sorting at query time on fields containing the same type of value.
```
gmourier, line 26: Nice one! 🤓

gmourier (changes requested):
Thanks, @guimachiavelli 💪  I changed a few small details.

learn/core_concepts/relevancy.md
gmourier: Was it initially written for custom ranking rules or more generally for ranking rules?

learn/core_concepts/relevancy.md
guimachiavelli: Great, thanks for clarifying that. I thought custom ranking rule behavior had remained the same.

learn/core_concepts/relevancy.md
guimachiavelli: You're absolutely right that it was written with custom ranking rules in mind. Leaving `custom` out was an oversight, thanks for spotting this!

gmourier (approved):
🏎

reference/api/sortable_attributes.md
guimachiavelli: Both could work because the article (`a`/`the`) needs to agree with `index`, not `sortableAttributes`. I also generally tend to think of settings as singular: "get an index's `sortableAttributes` setting" vs "get all sortable attributes belonging to an index."

I have preferred `a/an` in cases like this because an instance can have multiple indexes, each of which can have different settings. What do you think?

curquiza (approved):

reference/features/search_parameters.md
```diff
@@ -366,3 +367,31 @@ You would get the following response with **information about the matches in the
...
 }
 ```
+
+## Sort
+
+**Parameter**: `sort`
+**Expected value**: a list of attributes written as an array or as a comma-separated string
+**Default value**: `null`
+
+Sorts search results at query time according to the specified attributes and indicated order.
```
bidoubiwa, line 377: It may be missing some usage information like the one you can find in [facets distribution](https://docs.meilisearch.com/reference/features/search_parameters.html#matches).

It lets the user quickly understand how the string is constructed without having to look at the examples.

```
Usage: `attribute:order`
A search formula is composed of an attribute and an order. Both are separated by a colon `:`. For example, `price:asc`. 
```

reference/features/sorting.md
```diff
@@ -0,0 +1,200 @@
...
+
+To allow your users to sort results at search time you must:
+
+1. Decide which attributes you want to use for sorting
+2. Add those attributes to the `sortableAttributes` index setting
+3. Update MeiliSearch's [ranking rules](/learn/core_concepts/relevancy.md) (optional)
+
+### Select attributes for sorting
+
+MeiliSearch allows you to sort results based on document fields. Only fields containing numbers or strings can be used for sorting.
```
bidoubiwa, line 21: Fields containing arrays also work. If I understood correctly if you have two different documents: 

```json
{
   "id": 1,
   "something": [4, 5]
},
{
   "id": 2,
   "something": [3, 6]
}
```

and you search on `something:asc`, then document with id `1` will be more relevant as the lowest number in both array is `3`.
@gmourier Could you confirm?

reference/features/sorting.md
```diff
@@ -0,0 +1,200 @@
...
+
+### Adding attributes to `sortableAttributes`
+
+After you have decided which fields you will allow your users to sort on, you must add their attributes to the [`sortableAttributes` index setting](/reference/api/sortable_attributes.md).
+
+`sortableAttributes` accepts an array of strings, each corresponding to one attribute. Note that the attribute order in `sortableAttributes` has no impact on sorting.
+
+#### Example
+
+Suppose you have collection of books containing the following fields:
```
bidoubiwa, line 37: I think only two documents may be enough! But feel free to ignore if you disagree :)

reference/features/sorting.md
```diff
@@ -0,0 +1,200 @@
...
+#### Example
+
+If your users care more about finding cheaper books than they care about finding specific matches to their queries, you can place `sort` much higher in the ranking rules:
+
+<CodeSamples id="sorting_guide_update_ranking_rules_1" />
+
+## Sorting results at search time
+
+After configuring `sortableAttributes`, you can use the [`sort` search parameter](/reference/features/search_parameters.md#sort) to control the sorting order of your search results.
+
```
bidoubiwa, line 114: Same as comment in searrch parameters. 

It lets the user quickly understand how the string is constructed without having to look at the examples.

```
Usage: `attribute:order`
A search formula is composed of an attribute and an order. Both are separated by a colon `:`. For example, `price:asc`. 
```

What do you think?

bidoubiwa (changes requested):
🔥🔥🔥

.code-samples.meilisearch.yaml
```diff
@@ -421,3 +428,63 @@ get_dump_status_1: |-
...
+  curl \
+  -X POST 'http://localhost:7700/indexes/books/settings/ranking-rules' \
+  --data '[
+    "words",
+    "sort",
+    "typo",
+    "proximity",
+    "attribute",
+    "exactness"
+  ]'
```
bidoubiwa, line 450: Some times you use the subroute of a setting and sometime the main setting route. Is its willingly? Its not important I'm just curious ^^

reference/features/sorting.md
gmourier: This is exact! Only the first level is taken into account.

bidoubiwa (changes requested):
The naming of the code samples are not consistent with the other namings: 

```
sortable_attributes_get_1: |-
sortable_attributes_update_1: |-
sortable_attributes_reset_1: |-
```

example
```
get_filterable_attributes_1
update_filterable_attributes_1
reset_filterable_attributes_1
```

Its not a big problem for the SDK's to change them

reference/features/sorting.md
guimachiavelli: Hmmm…I think two might be too little, but you're probably right we can cut one or two of these. I'll try it out, thanks!

reference/features/sorting.md
guimachiavelli: Yeah, that's great. Will add this.

guimachiavelli:
> The naming of the code samples are not consistent with the other namings:
> 
> ```
> sortable_attributes_get_1: |-
> sortable_attributes_update_1: |-
> sortable_attributes_reset_1: |-
> ```
> 
> example
> 
> ```
> get_filterable_attributes_1
> update_filterable_attributes_1
> reset_filterable_attributes_1
> ```

As discussed elsewhere: for this release, we'll stick to the existing convention. Thanks for pointing this out and my apologies to the integrations team for the confusion.

I will update the `sortableAttributes` setting code samples shortly. Their final IDs will be:

> ```
> get_sortable_attributes_1
> update_sortable_attributes_1
> reset_sortable_attributes_1
> ```

.code-samples.meilisearch.yaml
bidoubiwa: At the time I thought it may be less confusing for the user to only have on route. But I don't realise if it creates complexity or if it reduce it. So it's really up to your guts :')

reference/features/search_parameters.md
bidoubiwa: Good for me

.code-samples.meilisearch.yaml
guimachiavelli: After thinking a bit more about it, I think it actually creates complexity and takes attention away from what's important in the guide.

I will use the direct subroutes.

.code-samples.meilisearch.yaml
```diff
@@ -114,6 +115,10 @@ update_settings_1: |-
             "a",
             "an"
         ],
+        "sortableAttributes": [
+          "price",
+          "author"
```
alallema, line 120:
```suggestion
          "title",
          "release_date"
```
For consistency, if this example is supposed to feat with the movies dataset I'm not sure there is `price` and `author` fields available

curquiza: Not sure it fits with the movie dataset since the index is `books` and not `movies`

bidoubiwa (approved):
Great guide! 🚀

.code-samples.meilisearch.yaml
guimachiavelli: @alallema is right, very good catch! 

That's the example for the main settings route, which indeed uses the movies index. I changed it without thinking that it was not in the isolated context of sorting.

PR #1091, opened 2021-08-30
Fix filtering/faceted search redirects

guimachiavelli:
As reported on meilisearch/MeiliSearch/issues/1648, old links pointing to `reference/features/filtering` do not work and lead the user to a 404.

The reason for this is that the current redirect rule sends users to the `/reference/api/filtering_and_faceted_search.html`  instead of `/reference/features/filtering_and_faceted_search.html`

### DIFF
--- a/.vuepress/public/_redirects
+++ b/.vuepress/public/_redirects
@@ -77,5 +77,5 @@
 
 # Group filters and facets content
 /reference/api/attributes_for_faceting.html         /reference/api/filterable_attributes.html
-/reference/feature/filtering.html                   /reference/api/filtering_and_faceted_search.html
-/reference/feature/faceted_search.html              /reference/api/filtering_and_faceted_search.html
+/reference/feature/filtering.html                   /reference/feature/filtering_and_faceted_search.html
+/reference/feature/faceted_search.html              /reference/feature/filtering_and_faceted_search.html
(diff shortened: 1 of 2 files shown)


.vuepress/public/_redirects
```diff
@@ -77,5 +77,5 @@
 
 # Group filters and facets content
 /reference/api/attributes_for_faceting.html         /reference/api/filterable_attributes.html
-/reference/feature/filtering.html                   /reference/api/filtering_and_faceted_search.html
-/reference/feature/faceted_search.html              /reference/api/filtering_and_faceted_search.html
+/reference/features/filtering.html                   /reference/features/filtering_and_faceted_search.html
```
Kerollmops, line 80: Is this url supposed to link to https://docs.meilisearch.com/reference/features/filtering.html, because it lead to a  404.

Kerollmops (changes requested):

.vuepress/public/_redirects
guimachiavelli: Links on the errors page are hard-coded, so they link to the live docs. I only updated the redirects as a quick patch to catch all instances of the wrong link, so the 404 is expected.

That said, it's probably a good idea to update the errors link directly as well. I've done that in my latest commit to this PR.

Kerollmops (approved):
Thank you very much, LGTM! I checked the links

curquiza (approved):

PR #1097, opened 2021-09-02
v0.22: Sentry

guimachiavelli:
MeiliSearch no longer uses Sentry to collect usage data. 

According to `ack`, the telemetry page is the only remaining location where we still mention it.

### DIFF
--- a/learn/what_is_meilisearch/telemetry.md
+++ b/learn/what_is_meilisearch/telemetry.md
@@ -8,7 +8,7 @@ MeiliSearch collects anonymized data from users in order to improve our product.
 
 ## What kind of data do we collect?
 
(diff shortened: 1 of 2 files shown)


gmourier:
There is also a mention of Sentry at the end of the [FAQ page](https://docs.meilisearch.com/faq.html#why-does-meilisearch-send-data-to-amplitude-does-meilisearch-track-its-users) 😇

guimachiavelli:
Duh, I was running `ack` on case-sensitive mode. Thanks for catching this before I merged =)

curquiza (approved):
🚀

gmourier (approved):

gmourier:
🚀

PR #1103, opened 2021-09-07
v0.22: Indexer

guimachiavelli:
v0.22 will feature a new, reworked indexer. This has implications for indexation performance and is a good opportunity for a more in-depth article on the subject.

- [x] /faq: add new question "Why does Meilisearch crash when I try to add documents?"
- [x] /faq: add new question "How can I speed up the indexation process?"
- ~~reference/under_the_hood/storage: review page~~ This page is being updated directly by the core team on #1106
- [x] reference/under_the_hood/indexation: add new page "Indexation and performance"

### DIFF
--- a/reference/under_the_hood/indexation.md
+++ b/reference/under_the_hood/indexation.md
@@ -18,6 +18,6 @@ Crashes caused by memory overconsumption can still happen in two cases:
 
 ## Multi-threading
 
-The indexer never uses more than half of the available processor cores. For example, if your machine has twelve cores, the indexer will use six of them at most. This ensures that MeiliSearch is always ready to perform searches, even while you are updating an index.
+In machines with multi-core processors, the indexer avoids using more than half of the available processing units. For example, if your machine has twelve cores, the indexer will use six of them at most. This ensures MeiliSearch is always ready to perform searches, even while you are updating an index.
 
 Multi-threading is unfortunately not possible in machines with only one processor core.
--- a/reference/under_the_hood/indexation.md
+++ b/reference/under_the_hood/indexation.md
@@ -1,0 +1,24 @@
+# Indexation and performance
+
+MeiliSearch indexation is a memory-intensive multi-threaded operation. It is one of the most important operations and at the core of what makes our search engine fast, relevant and reliable. In this article, we will go over some of the details regarding RAM consumption and multi-threading.
+
+
+## RAM
+
+Our indexer uses the `sysinfo` Rust library to calculate a machine's total memory size and adapts its behavior so indexation uses a maximum two thirds of available resources. This process helps us with two very important things:
+
+1. By making sure indexation does use more resources than available in a machine, we minimize crashes due to RAM over-consumption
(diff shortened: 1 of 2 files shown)


curquiza (approved):
LGTM 😊 
I asked @ManyTheFish as reviewer since he implemented the new indexer :)

resources/faq.md
```diff
@@ -245,3 +245,19 @@ In general, we recommend the former. However, if you need to reduce the size of
 **MeiliSearch will never track or identify individual users**. That being said, we do use Amplitude and Sentry to collect anonymous data about user trends and bug reports.
 
 You can read more about what metrics we collect, why we collect them, and how to disable it on our [telemetry page](/learn/what_is_meilisearch/telemetry.md). Transparency is very important to us, so if you feel we are lacking in this area please [open an issue](https://github.com/meilisearch/documentation/issues/new/choose) and let us know! ❤️
+
+## Why does MeiliSearch crash when I try to add documents?
```
ManyTheFish, line 249: during indexing, we are using disk too, we could crash if there is no space left on device.

reference/under_the_hood/indexation.md
```diff
@@ -0,0 +1,23 @@
...
+
+Crashes caused by memory overconsumption can still happen in two cases:
+
+1. `sysinfo` may not be able to calculate the amount of available RAM for certain OSes. MeiliSearch still makes an educated estimate and adapts its the behavior based on that, but crashes may still happen in this case. [Follow this link for an exhaustive list of OSes supported by `sysinfo`](https://docs.rs/sysinfo/0.20.0/sysinfo/#supported-oses).
+
+2. Lower-end machines might struggle when processing huge datasets. Splitting your data payload into smaller batches can help in this case. [For more information, consult the FAQ.](/resources/faq.md#why-does-meilisearch-crash-when-i-try-to-add-documents)
+
+## Multi-threading
+
+The indexer never uses more than half of the available processor cores. For example, if your machine has twelve cores, the indexer will use six of them at most. This ensures that MeiliSearch is always ready to perform searches, even while you are updating an index.
```
ManyTheFish, line 21: if we only have 1 thread, the indexing uses 1 thread which is more than half of the available threads, I'm not comfortable with the `never uses more` expression

ManyTheFish (commented):
It's a pretty clear explanation! Thanks! 👍

resources/faq.md
guimachiavelli: Thanks! I added that to the FAQ 👍

reference/under_the_hood/indexation.md
guimachiavelli: I definitely got carried away with that `never` 😅 

I rephrased it so it's a bit more nuanced. It would be great if you could check it again just to be sure it's fine now (completely fine if you don't, though)

guimachiavelli:
Merging this into the v0.22 branch. Thanks for the feedback, folks ❤️

PR #1114, opened 2021-09-14
Add trace in log level

bidoubiwa:
To keep the doc exhaustive and follow the moto 

> the documentation is always complete but never finished

I think we should not hide any feature. 

I've added the missing log level as this may be used by external contributors but also for futur meili developers that uses the doc to know the possibilities MeiliSeach offers.

### DIFF
--- a/reference/features/configuration.md
+++ b/reference/features/configuration.md
@@ -199,6 +199,7 @@ MeiliSearch currently supports four log levels, listed in order of increasing ve
 - `'WARN:'` log all unexpected events, regardless of their severity
 - `'INFO:'` log all events. This is the default value of `--log-level`
 - `'DEBUG'`: log all events and including detailed information on MeiliSearch's internal processes. Useful when diagnosing issues and debugging
+- `'TRACE'`: log all events, details, and even more detailed information on MeiliSearch's internal processes. We do not advise using this level as it is overkill. Consider using `DEBUG` instead.
 
 ### Max index size
 
--- a/reference/features/configuration.md
+++ b/reference/features/configuration.md
@@ -199,7 +199,7 @@ MeiliSearch currently supports four log levels, listed in order of increasing ve
 - `'WARN:'` log all unexpected events, regardless of their severity
 - `'INFO:'` log all events. This is the default value of `--log-level`
 - `'DEBUG'`: log all events and including detailed information on MeiliSearch's internal processes. Useful when diagnosing issues and debugging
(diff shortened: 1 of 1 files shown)


guimachiavelli:
Thanks for this! I didn't include it originally because of the info on this `transplant` issue: https://github.com/meilisearch/transplant/issues/193, where it's specified `TRACE` is not currently used.

@curquiza, has `TRACE` been implemented? What would users get if they set the log level configuration option to that value?

curquiza:
`TRACE` is implemented but should not be used by the users: it will add too many logs. It's only for MeiliSearch contributors for debugging.

reference/features/configuration.md
```diff
@@ -199,6 +199,7 @@ MeiliSearch currently supports four log levels, listed in order of increasing ve
 - `'WARN:'` log all unexpected events, regardless of their severity
 - `'INFO:'` log all events. This is the default value of `--log-level`
 - `'DEBUG'`: log all events and including detailed information on MeiliSearch's internal processes. Useful when diagnosing issues and debugging
```
guimachiavelli, line 201: Oops, spotted something wrong on the text for the DEBUG level.

```suggestion
- `'DEBUG'`: log all events and include detailed information on MeiliSearch's internal processes. Useful when diagnosing issues and debugging
```

reference/features/configuration.md
```diff
@@ -199,6 +199,7 @@ MeiliSearch currently supports four log levels, listed in order of increasing ve
...
+- `'TRACE'`: log all events, and even more detailed information on MeiliSearch's internal processes. We do not advise using this level as it is overkill and overwhelming. Use `DEBUG` before considering `TRACE`.
```
guimachiavelli, line 202: `overkill and overwhelming` has a very different tone when compared to the text in the rest of the page and is not explicit on what makes this level overwhelming. What do you think we go for this instead?

```suggestion
- `'TRACE'`: log all events and include even more detailed information on MeiliSearch's internal processes. We do not advise using this level as it is extremely verbose. Use `DEBUG` before considering `TRACE`.
```

guimachiavelli (changes requested):
Thanks for this! 🍰 

I noticed one mistake on my part in the `DEBUG` level description, so I'm adding a suggestion fixing that.

Also, a minor stylistic change to your edit—let me know if you think it's fine!

guimachiavelli:
bors merge

PR #1123, opened 2021-09-20
core concept changes

maryamsulemani97:
made minor changes to all the files under Core Concepts.


guimachiavelli (changes requested):
Thanks for this!

I've made a few comments before giving final approval, take a look and let me know.

Additionally, in the future try to qualify PR names so reviewers can know what the PR is about. To use this one as an example, something like `Core concept style and wording changes` would make the main goal of the PR clearer.

PR #1150, opened 2021-09-30
V23 update for ndjson and csv payloads

maryamsulemani97:


errors.yaml
```diff
@@ -41,7 +41,7 @@ errors:
   - code: unretrievable_document
     description: "The document exists in store, but there was an error retrieving it. This probably comes from an inconsistent state in the database."
   - code: unsupported_media_type
-    description: "The payload content type is not supported by MeiliSearch. Currently, MeiliSearch supports only JSON payloads."
+    description: "The payload content type is not supported by MeiliSearch. Currently, MeiliSearch only supports JSON, CSV, and NDJSON payloads."
```
gmourier, line 44: There are new errors described in the specs related to the format handling 🤓

- https://github.com/meilisearch/specifications/blob/develop/text/0028-indexing-csv.md#errors
- https://github.com/meilisearch/specifications/blob/develop/text/0029-indexing-ndjson.md#errors

learn/core_concepts/documents.md
```diff
@@ -38,6 +46,30 @@ In the above example, `"id"`, `"title"`, `"genre"`, `"release-year"`, and `"cast
...
+CSV objects are separated by delimiters that could be a space, semicolon, or some other character, though the comma is most common. Any [rules that apply to formatting CSV](https://datatracker.ietf.org/doc/html/rfc4180) also apply to MeiliSearch documents.
+
+The above JSON document would look like this in CSV:
+
+```csv
+  "id","title","genre","release-year","cast"
+  "1564saqw12ss","Kung Fu Panda","Children's Animation",2008,""
+```
+
+Since CSV does not support arrays, the cast cannot be converted to CSV.
```
gmourier, line 71:
```suggestion
Since CSV does not support arrays or nested objects, the cast cannot be converted to CSV.
```

learn/core_concepts/documents.md
```diff
@@ -156,7 +188,7 @@ Take note that the document addition request in MeiliSearch is <clientGlossary w
 
 ## Upload
 
-By default, MeiliSearch limits the size of `JSON` payloads—and therefore document uploads—to 100MB.
+By default, MeiliSearch limits the size of all payloads (JSON, CSV, NDJSON) —and therefore document uploads—to 100MB.
```
gmourier, line 191: I am not sure if it is useful to specify all the formats here 🤔

learn/core_concepts/documents.md
```diff
@@ -38,6 +46,30 @@ In the above example, `"id"`, `"title"`, `"genre"`, `"release-year"`, and `"cast
...
+```
+
+#### CSV objects
+
+CSV objects are separated by delimiters that could be a space, semicolon, or some other character, though the comma is most common. Any [rules that apply to formatting CSV](https://datatracker.ietf.org/doc/html/rfc4180) also apply to MeiliSearch documents.
+
+The above JSON document would look like this in CSV:
+
+```csv
+  "id","title","genre","release-year","cast"
```
gmourier, line 67: Something important is missing about the CSV format. Maybe it should be explained on a dedicated page?

Indeed, the explanation of the typing within the header is missing to help the search engine to select the right type for indexing since the cell values are contained between two quotes.

CSV data format needs to contain a first line representing the list of attributes with the optionally chosen type separated from the attribute name with the `:` character. The type is case insensitive.
An attribute can be specified with two types: string or number. A missing type will be interpreted as a string by default.

Valid headline example: `"id:number","title:string","author","price:number"`

gmourier (changes requested):

gmourier:
Ping @curquiza

learn/core_concepts/documents.md
maryamsulemani97: @gmourier 
We can have a dedicated page as well, but we do have a link directing the user to the rules for CSV. Is that not enough?

gmourier: Oh, sorry! I didn't see the link. That should be enough! 

We can improve it later if we realize that the information is not as easy to find as we expected. 🤓

I think that it is necessary to have a part that explains the typing of the fields to guide the user who will want to index documents from a CSV file.

learn/core_concepts/documents.md
```diff
@@ -38,6 +46,30 @@ In the above example, `"id"`, `"title"`, `"genre"`, `"release-year"`, and `"cast
 Each attribute must be associated with a **value**, e.g. `"Kung Fu Panda"` is the value of `"title"`.
 At minimum, the document must contain one field with the **[primary key][primary-key]** attribute and a unique **[document id][document-id]** as its value. Above, that's: `"id": "1564saqw12ss"`.
 
+#### NDJSON objects
+
+NDJSON objects consist of individual lines where each individual line is valid JSON text and each line is delimited with a newline character. Any [rules that apply to formatting NDJSON](http://ndjson.org/) also apply to MeiliSearch documents.
+
+The above JSON document would look like this in NDJSON:
+
+```ndjson
```
guimachiavelli, line 55: Would this work?

```suggestion
```json
```

I think that example should count as valid JSON and get some syntax highlighting.

learn/core_concepts/documents.md
```diff
@@ -38,6 +46,30 @@ In the above example, `"id"`, `"title"`, `"genre"`, `"release-year"`, and `"cast
...
+CSV objects are separated by delimiters that could be a space, semicolon, or some other character, though the comma is most common. Any [rules that apply to formatting CSV](https://datatracker.ietf.org/doc/html/rfc4180) also apply to MeiliSearch documents.
+
+The above JSON document would look like this in CSV:
+
+```csv
+  "id","title","genre","release-year","cast"
+  "1564saqw12ss","Kung Fu Panda","Children's Animation",2008,""
+```
+
+Since CSV does not support arrays or nested objects, the cast cannot be converted to CSV.
```
guimachiavelli, line 71: It took a few moments for me to understand what "the cast" was referring to here. 

First of all, "cast" should be marked as code with backticks (`\``). We are not referring to cast, the word, here, but `cast`, the column header, which is in certain ways like a variable name or a an object's key.

Second, that definite article (`the`) is unnecessary. If you want to keep it, you should qualify `cast`: "the `cast` field" could work, for instance.

reference/features/configuration.md
```diff
@@ -230,7 +230,7 @@ The `update` database handles [pending updates](/learn/advanced/asynchronous_upd
 **CLI option**: `--http-payload-size-limit`
 **Default value**: `104857600` (~100MB)
 
-Sets the maximum size of accepted JSON payloads. Value must be given in bytes or explicitly stating a base unit. For example, the default value can be written as `107374182400`, `'107.7Gb'`, or `'107374 Mb'`.
+Sets the maximum size of accepted JSON, CSV, or NDJSON payloads. Value must be given in bytes or explicitly stating a base unit. For example, the default value can be written as `107374182400`, `'107.7Gb'`, or `'107374 Mb'`.
```
guimachiavelli, line 233: What do you think we find a different way of saying this that doesn't specify accepted file formats? Trying to think ahead and avoiding having to update this when/if we start supporting other formats.

reference/api/README.md
```diff
@@ -14,9 +14,13 @@ Check out [the FAQ](/resources/faq.md) for answers to some common questions 💡
 
 #### Content type
 
-Every request and response body is in `JSON`.
+ Requests can be in JSON, CSV, or NDJSON but responses are always in JSON.
```
guimachiavelli, line 17: It's not always pretty, but in contexts like API docs I always prefer to use short and dry sentences.

```suggestion
 Requests can be in JSON, CSV, or NDJSON. Responses are always in JSON.
```

resources/faq.md
```diff
@@ -72,7 +72,7 @@ If you are curious about how the asynchronous part of MeiliSearch works, you can
 
 ## I am trying to add my documents but I keep receiving a `400 - Invalid data` response.
 
-MeiliSearch API only accepts JSON format.
+MeiliSearch API accepts JSON, CSV, and NDJSON formats.
```
guimachiavelli, line 75: @gmourier, do we still get `400 — Invalid data` errors? According to the CSV spec, 400 should correspond to `Bad Request`. Or are these the same and largely dependent on stuff like browser implementation?

guimachiavelli (changes requested):
Requesting a few changes.

In general, I'm wondering if we can't remove a lot of mentions to objects (e.g. `CSV objects`). Seems like a lot of these don't add much relevant information and make things more complex than they actually are.

learn/core_concepts/documents.md
```diff
@@ -13,14 +13,22 @@ Documents function as **containers for organizing data**, and are the basic buil
 - **Document**: an object which contains data in the form of one or more fields.
 - **[Field][fields]**: a set of two data items that are linked together: an **attribute** and a **value**.
 - **Attribute**: the first part of a field. Acts as a name or description for its associated value.
-- **Value**: the second part of a field, consisting of data of any valid `JSON` type.
+- **Value**: the second part of a field, consisting of data of any valid JSON type.
 - **[Primary Field][primary-field]**: A special field that is mandatory in all documents. It contains the primary key and document identifier.
 - **[Primary Key][primary-key]**: the attribute of the primary field. **All documents in the same index must possess the same primary key.** Its associated value is the document identifier.
 - **[Document Identifier][document-id]**: the value of the primary field. **Every document in a given index must have a unique identifier**.
 
 ### Formatting
```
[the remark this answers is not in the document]
maryamsulemani97: Agreed

reference/features/configuration.md
maryamsulemani97: Sets the maximum size of accepted payloads. Link this to the Documents page where we mention the acceptable formats?

learn/core_concepts/documents.md
maryamsulemani97: My bad, will update this

reference/api/README.md
```diff
@@ -14,9 +14,13 @@ Check out [the FAQ](/resources/faq.md) for answers to some common questions 💡
...
 
-Every request and response body is in `JSON`.
+ Requests can be in JSON, CSV, or NDJSON. Responses are always in JSON.
 
-It is **not required** to have `Content-Type: application/json` in the header. Any content-type is accepted.
+You **must** specify a Content-Type header. MeiliSearch currently supports the following formats:
+
+- `Content-Type: application/json` for JSON
+- `Content-Type: application/x-ndjson` for NDJSON
+- `Content-Type: text/csv` for CSV
```
curquiza, line 23: CSV and NDJSON are only supported for `/indexes/:uid/documents` route (`POST` and `PUT`).
You should not be able to push your setting in CSV via the `/indexes/:uid/settings` routes that only accept JSON.
@gmourier can you confirm?

errors.yaml
gmourier: Absolutely @guimachiavelli! `invalid_content_type` and `missing_content_type`  and not `media_type` 😇

There are also two new `malformed_payload` and `missing_payload` errors that should be listed. I guess you were aware of it and expected to ship it with another PR?

`malformed_payload` can occur for several reasons:

- Sending a different payload type than the `Content-Type` header
- Sending an invalid CSV or NDJSON format (For now, if the CSV header does not match what we expect, the payload is considered to be invalid. We will probably have a dedicated error in the future)

`missing_payload` can occur if a payload is expected but empty

learn/core_concepts/documents.md
gmourier: It could be useful in dedicated guides in the future. e.g. "How to send data from a SQL database to MeiliSearch?". (SQL database exports are often in CSV format).

In addition to providing new formats, the goal is also to speed up indexing times. NDJSON and CSV are streamable. There is no need to load the whole file in RAM to validate it.

At the moment it could be a sentence explaining it succinctly! What do you think?

reference/api/README.md
gmourier: Yes, I confirm!

resources/faq.md
gmourier: This is legacy! Nice spot @guimachiavelli. I don't know when it changed probably before v0.20 🤓

The current error is 

```
{
    "message": "invalid type: map, expected a sequence at line 1 column 1. The json payload provided is malformed.",
    "errorCode": "bad_request",
    "errorType": "invalid_request_error",
    "errorLink": "https://docs.meilisearch.com/errors#bad_request"
}
```

and it will become a `malformed_payload` error code.

learn/core_concepts/documents.md
guimachiavelli: Hmmm…this article is under the `learn` section, so explaining the intricacies of file formats feels a bit too advanced and out-of-scope.

I propose we:

1. leave comments regarding why a user would prefer a file format over another (e.g. CSV over JSON) out of this article;
2. open a new issue proposing a new guide that is specifically about file formats;
3. once the guide is done and published, we'll update `documents.md` with a link to it.

resources/faq.md
guimachiavelli: Just to confirm we understand this: Is it possible a user gets a `400 — Bad Request` when adding documents due to a cause other than `malformed_payload`?

reference/api/README.md
guimachiavelli: It is mandatory to specify `Content-Type` headers for every `PUT` and `POST` routes, correct?

errors.yaml
guimachiavelli: Ha, sorry, was a bit tired when I wrote the error type names!

Are these errors parsed in the order you mentioned them here? For example, if I send an update request without `Content-Type` headers and a payload, I will get `missing_content_type` or `missing_payload`?

That aside, confirming errors and tasks:

1. remove `unsupported_media_type` ;
2. add `invalid_content_type`:  `Content-Type` header specifies an unsupported file format (e.g. SQL or zip)
3. add `missing_content_type`: payload does not contain a `Content-Type` header
4. add `malformed_payload`:  payload is not formatted according to the specified content type (i.e. user specifies e.g. JSON but the dataset is malformed or is actually something else which cannot be read as valid JSON)
5. add `missing_payload`: `Content-Type` was specified, but no file was sent to the server or the file itself is empty

reference/api/README.md
curquiza: Nope, only if we pass a payload. For example, the route to create a dump is `POST` but does not require any payload so no need to enforce the content type. I think this is the only exception, but I might be wrong.

guimachiavelli: Got it, thanks, @curquiza!

create/how_to/meilisearch_react.md
```diff
@@ -112,7 +112,7 @@ const MeiliSearch = require("meilisearch");
 
 ## Index documents
 
-MeiliSearch receives documents in JSON format and stores them for searching purposes. These documents are composed of fields that can hold any type of data.
+MeiliSearch receives documents in JSON, CSV, and NDJSON format and stores them for searching purposes. These documents are composed of fields that can hold any type of data.
```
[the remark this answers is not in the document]
maryamsulemani97: Updated

errors.yaml
gmourier: "Are these errors parsed in the order you mentioned them here? For example, if I send an update request without Content-Type headers and a payload, I will get missing_content_type or missing_payload?"

Yes, according to my point of view the `missing_content_type`error should come before `missing_payload`. I'll do some tests on the RC to see if this is the case. For the time being it's not very critical if it's not the case because the user will always find his way through the errors no matter which way.

I also confirm your last point 🤓

resources/faq.md
gmourier: Yes! `missing_payload` is also a `400 — Bad Request` HTTP code.

reference/api/README.md
maryamsulemani97: updated

resources/faq.md
maryamsulemani97: updated

errors.yaml
maryamsulemani97: updated

errors.yaml
```diff
@@ -40,10 +40,16 @@ errors:
...
-  - code: unsupported_media_type
-    description: "The payload content type is not supported by MeiliSearch. Currently, MeiliSearch supports only JSON payloads."
+  - code: invalid_content_type
+    description: "The Content-Type header is not supported by MeiliSearch. Currently, MeiliSearch only supports JSON, CSV, and NDJSON."
+  - code: missing_content_type
+    description: "The payload does not  contain a Content-Type header. Currently, MeiliSearch only supports JSON, CSV, and NDJSON."
+  - code: malformed_payload  
+    description: "The Content-Type header does not match the payload or the format is invalid."
+  - code: missing_payload
+    description: "The Content-Type was specified, but no file was sent to the server or the file itself is empty."
```
gmourier, line 50:
```suggestion
    description: "The Content-Type header was specified, but no request body was sent to the server or the request body is empty."  
```

errors.yaml
```diff
@@ -40,10 +40,16 @@ errors:
...
   - code: unretrievable_document
     description: "The document exists in store, but there was an error retrieving it. This probably comes from an inconsistent state in the database."
-  - code: unsupported_media_type
-    description: "The payload content type is not supported by MeiliSearch. Currently, MeiliSearch supports only JSON payloads."
+  - code: invalid_content_type
+    description: "The Content-Type header is not supported by MeiliSearch. Currently, MeiliSearch only supports JSON, CSV, and NDJSON."
+  - code: missing_content_type
+    description: "The payload does not  contain a Content-Type header. Currently, MeiliSearch only supports JSON, CSV, and NDJSON."
+  - code: malformed_payload  
+    description: "The Content-Type header does not match the payload or the format is invalid."
```
gmourier, line 48:
```suggestion
    description: "The Content-Type header does not match the request body payload format or the format is invalid."
```

gmourier (changes requested):
Since the "payload" term can represent the whole request, I think is much clearer to say "Request body payload" to help the user. That's just my 2 cents! 😇

learn/core_concepts/documents.md
```diff
@@ -38,6 +46,40 @@ In the above example, `"id"`, `"title"`, `"genre"`, `"release-year"`, and `"cast
...
+#### CSV
+
+CSV files express data as a sequence of values separated by a delimiter character. Though the comma is the most common delimiter, spaces, tabs, and semicolons are valid alternatives. Any [rules that apply to formatting CSV](https://datatracker.ietf.org/doc/html/rfc4180) also apply to MeiliSearch documents.
+
+Compared to JSON, CSV has better writing performance and is less CPU and memory intensive.  
+
+The above JSON document would look like this in CSV:
+
+```csv
+  "id:string","title:string","genre:string","release-year:number"
```
gmourier, line 77: If `:string` or `:number` is not specified, it will fallback to `:string` by default. I think it's worth mentioning 😇

gmourier (changes requested):

learn/core_concepts/documents.md
maryamsulemani97: Updated to add this as tip

learn/core_concepts/documents.md
```diff
@@ -38,6 +46,44 @@ In the above example, `"id"`, `"title"`, `"genre"`, `"release-year"`, and `"cast
...
+  "title": "Kung Fu Panda", 
+  "genre": "Children's Animation", 
+  "release-year": 2008, 
+  "cast": [ {"Jack Black": "Po"}, {"Jackie Chan": "Monkey"} ]
+  }
+```
+
+#### CSV
+
+CSV files express data as a sequence of values separated by a delimiter character. Though the comma is the most common delimiter, spaces, tabs, and semicolons are valid alternatives. Any [rules that apply to formatting CSV](https://datatracker.ietf.org/doc/html/rfc4180) also apply to MeiliSearch documents.
```
gmourier, line 70: For the moment MeiliSearch only uses the  `,` separator. We don't support other separators but it could be a possibility in the future.

I'm sorry @maryamsulemani97, it's entirely my fault because it's not mentioned in the specification. I will update it right away.

maryamsulemani97: No worries! Updated the CSV part to mention this

gmourier (approved):
Thanks, @maryamsulemani97! 🚀🚀🚀

reference/features/configuration.md
```diff
@@ -230,7 +230,7 @@ The `update` database handles [pending updates](/learn/advanced/asynchronous_upd
 **CLI option**: `--http-payload-size-limit`
 **Default value**: `104857600` (~100MB)
 
-Sets the maximum size of accepted JSON payloads. Value must be given in bytes or explicitly stating a base unit. For example, the default value can be written as `107374182400`, `'107.7Gb'`, or `'107374 Mb'`.
+Sets the maximum size of [accepted payloads](https://docs.meilisearch.com/learn/core_concepts/documents.html#dataset-format). Value must be given in bytes or explicitly stating a base unit. For example, the default value can be written as `107374182400`, `'107.7Gb'`, or `'107374 Mb'`.
```
guimachiavelli, line 233: Very important thing here: always prefer relative links when linking to stuff within the docs website. I think the only place in the docs where we can't use relative links is on `errors.yaml`.

In case you need a refresher:

- absolute links specify the full path and protocol to a certain resource (e.g. an image, a page): `https://example.com`;
- relative links specify the path relative to the resource you're writing: `./an_image.jpg`. To make things easier, you can make them relative to the root of the website: `/learn/core_concepts/documents.md#dataset_format`. Note that we need to link to the `.md` file and not to the final compiled html resource.

There are a few reasons for this, but one of the most immediately useful for us is that our dead link checker only inspects relative links.

```suggestion
Sets the maximum size of [accepted payloads](/learn/core_concepts/documents.md#dataset-format). Value must be given in bytes or explicitly stating a base unit. For example, the default value can be written as `107374182400`, `'107.7Gb'`, or `'107374 Mb'`.
```

resources/faq.md
```diff
@@ -70,12 +70,12 @@ In addition, MeiliSearch delivers a global route to [track all your update statu
 
 If you are curious about how the asynchronous part of MeiliSearch works, you can find more information [here](/learn/advanced/asynchronous_updates.md).
 
-## I am trying to add my documents but I keep receiving a `400 - Invalid data` response.
+## I am trying to add my documents but I keep receiving a `400 - Bad Request` response.
 
-MeiliSearch API only accepts JSON format.
+MeiliSearch API accepts JSON, CSV, and NDJSON formats.
 In case of a [document addition](/reference/api/documents.md#add-or-replace-documents), only an array of objects is expected.
```
guimachiavelli, line 76: Only highlighting this to give you another example of a relative link in the docs.

PR #1168, opened 2021-10-04
Update dumps directory

alallema:
The dumps would be now create in a specific directory: `/var/opt/meilisearch/dumps`
This will be available only for MeiliSearch v0.23.0
See: meilisearch/cloud-scripts#55


alallema:
I close this PR in agreeing with @guimachiavelli. This branch part from master instead of v0.23 ... my bad

PR #1169, opened 2021-10-04
Update dumps directory

alallema:
The dumps would be now create in a specific directory: `/var/opt/meilisearch/dumps`
This will be available only for MeiliSearch v0.23.0
See: meilisearch/cloud-scripts#55

### DIFF
--- a/create/how_to/updating.md
+++ b/create/how_to/updating.md
@@ -157,7 +157,7 @@ Now that all fields are displayed, proceed to the next step.
 Before creating your dump, make sure that your [dump directory](/reference/features/configuration.md#dumps-destination) is somewhere accessible. By default, dumps are created in a folder called `dumps` at the root of your MeiliSearch directory.
 
 ::: note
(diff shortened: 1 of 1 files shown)


guimachiavelli (approved):
Looks great, thanks! ❤️ 

One question, which is perhaps of larger scope than this PR: would this path look the same in a non-UNIX environment? For example, if a user is on a windows machine?

alallema:
> Looks great, thanks! ❤️
> 
> One question, which is perhaps of larger scope than this PR: would this path look the same in a non-UNIX environment? For example, if a user is on a windows machine?

Not at all. Normally, the `dumps` folder should be found in the folder where you run the binary.
This script is made to work on the machine create in DO, AWS, and GCP with the tools we made for. The `dumps` folder is only here because we add the CLI-option: `--dumps-dir` when we are launching Meilisearch with `systemd`.

guimachiavelli:
I think it's alright. I only read the bit you edited at first, but I think it's clear in the context of the whole guide.

Thanks again! Will merge it later today or tomorrow morning.

PR #1173, opened 2021-10-06
users should check the index status before they start searching

### LINKED ISSUE #638: Clarify Getting Started
There should be a warning in Getting Started that attempting to search before all documents are indexed will result in undefined behavior.
OR:
Add checking the index status as a full additional step before search (this is better--beginner tutorial should teach checking the index status before search as a best practice)

[1133](https://github.com/meilisearch/MeiliSearch/issues/1133)

maryamsulemani97:
…ning of undefined behavior if they search before indexing finishes

closes #638

### DIFF
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -74,6 +74,9 @@ search_get_1: |-
 get_update_1: |-
   curl \
     -X GET 'http://localhost:7700/indexes/movies/updates/1'
+get_update_2: |-
+  curl \
+    -X GET 'http://localhost:7700/indexes/movies/updates/0' 
 get_all_updates_1: |-
   curl \
     -X GET 'http://localhost:7700/indexes/movies/updates'
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -74,7 +74,7 @@ search_get_1: |-
 get_update_1: |-
   curl \
     -X GET 'http://localhost:7700/indexes/movies/updates/1'
-get_update_2: |-
+getting_started_get_update: |-
   curl \
     -X GET 'http://localhost:7700/indexes/movies/updates/0' 
 get_all_updates_1: |-
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -74,9 +74,6 @@ search_get_1: |-
 get_update_1: |-
   curl \
     -X GET 'http://localhost:7700/indexes/movies/updates/1'
-getting_started_get_update: |-
-  curl \
-    -X GET 'http://localhost:7700/indexes/movies/updates/0' 
 get_all_updates_1: |-
   curl \
     -X GET 'http://localhost:7700/indexes/movies/updates'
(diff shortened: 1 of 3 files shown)


learn/getting_started/quick_start.md
```diff
@@ -81,7 +81,15 @@ See our guide on [asynchronous updates](/learn/advanced/asynchronous_updates.md)
 
 ## Search
 
-Now that your documents have been ingested into MeiliSearch, you are able to search them.
+Before you start querying, we recommend that you check the status of your index. You can do that using:
+
+<CodeSamples id="get_update_1" />
```
dichotommy, line 86: So, I'm not sure about this.

I believe that at this point in the quick start the user has made only one request triggering creation of an update object: the document addition request.

I also believe that the first update created for a MeiliSearch index will have the `updateId` of `0`. This code sample gives the value `1`, so I'm concerned that it might not work for users copy+pasting this command. Could you test it out yourself @maryamsulemani97 ?

dichotommy (changes requested):
This needs to be tested, since supplying the user with an incorrect command could create a lot of problems.

In general, the lack of clarity on what the user has or has not done at this point in the quick start guide speaks to how overwritten this guide has become. I think that cleaning it up should be one of our first priorities in the following weeks.

learn/getting_started/quick_start.md
maryamsulemani97: You're right, it should be `0`.

.code-samples.meilisearch.yaml
```diff
@@ -74,6 +74,9 @@ search_get_1: |-
 get_update_1: |-
   curl \
     -X GET 'http://localhost:7700/indexes/movies/updates/1'
+get_update_2: |-
```
dichotommy, line 77: Could we change the code sample name and position in `.code-samples.meilisearch.yaml` to reflect that it's used in the quick start guide? Something like `getting_started_get_update` would work. The current name + location make it seem like it's used in the API references.

dichotommy (changes requested):

learn/getting_started/quick_start.md
```diff
@@ -81,7 +81,15 @@ See our guide on [asynchronous updates](/learn/advanced/asynchronous_updates.md)
...
 
-Now that your documents have been ingested into MeiliSearch, you are able to search them.
+Before you start querying, we recommend that you check the status of your index. You can do that using:
+
+<CodeSamples id="getting_started_get_update" />
+
+Once the update shows `processed`, you can start searching.
+
+::: warning
+Attempting to search before all documents are indexed **will result in undefined behavior**.
```
dichotommy, line 91: It's already in a big warning box, so probably not necessary to bold this.

```suggestion
Attempting to search before all documents are indexed will result in undefined behavior.
```

.code-samples.meilisearch.yaml
```diff
@@ -74,6 +74,9 @@ search_get_1: |-
 get_update_1: |-
   curl \
     -X GET 'http://localhost:7700/indexes/movies/updates/1'
+getting_started_get_update: |-
```
dichotommy, line 77: Could you move this code sample to be next to the other `getting_started` code samples instead of being among the API reference code samples?

Sorry it's a bit nitpicky 😬

.vuepress/public/sample-template.yaml
```diff
@@ -17,6 +17,7 @@ delete_one_document_1: |-
 delete_documents_1: |-
 search_post_1: |-
 get_update_1: |-
+getting_started_get_update: |-
```
dichotommy, line 20: Same comment as above in `.code-samples.meilisearch.yaml`. It helps us keep these files tidy if code samples from the same page or site section are grouped together.

dichotommy (changes requested):

dichotommy (approved):
Looks great, thanks!

dichotommy:
bors merge

PR #1177, opened 2021-10-06
Added `Expected value` fields to configuration options

### LINKED ISSUE #1039: Add `expected value` fields to configuration options
#976 adds an explicit `default value` field for each configuration option. This is helpful, but can still be improved by adding an `expected value` field stating what sort of value each option requires.

For example:

```
Log level

Environment variable: MEILI_LOG_LEVEL
CLI option: --log-level
Default value: 'INFO'
```

Could become 

```
Log level

Environment variable: MEILI_LOG_LEVEL
CLI option: --log-level
Default value: 'INFO'
Expected value: one of `ERROR`, `WARN`, `INFO`. or `DEBUG`
```

This must be done for all configuration options.

_Note: currently all options require a value, even those whose behaviour is closer to a flag than actual option (e.g.: `--no-analytics` must be accompanied by a value in order to be valid, but it doesn't matter what that value actually is: `--no-analytics=true` and `--no-analytics=false` have, confusingly, the same effect)._

aniketh-varma:
Closes #1039

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -89,7 +89,7 @@ module.exports = {
           collapsable: false,
           children: [
             {
-              title: 'Contributing to the Docs',
+              title: 'Contributing to the docs',
               path: '/learn/contributing/contributing_to_docs',
             },
           ],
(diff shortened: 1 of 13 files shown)


maryamsulemani97:
Hey,

You should create a new branch from master. Hope that helps.

PR #1181, opened 2021-10-11
Add java code samples to sdks.json

ansavanix:
I have added the java code samples file to sdks.json.

### DIFF
--- a/.vuepress/code-samples/sdks.json
+++ b/.vuepress/code-samples/sdks.json
@@ -6,7 +6,7 @@
   },
   {
     "language": "javascript",
-    "label": "JavaScript",
+    "label": "JS",
     "url": "https://raw.githubusercontent.com/meilisearch/meilisearch-js/main/.code-samples.meilisearch.yaml"
   },
   {


dichotommy:
I tried testing this out by running `yarn build` and serving it locally with `http-server`. I don't see any Java samples present in the code tabs, so I think something isn't quite working. The error is presumably caused by something in the `.code-samples.meilisearch.yaml` in the Java repo, which may not have been prepared yet for integration with the docs site.

What do you think @alallema @bidoubiwa ? Are Java code samples ready to be added to the docs site?

Also, the code sample related to Gradle seems to be a troublemaker.

alallema:
I let you know when it's working

bidoubiwa:
@alallema It appears that there is one missing code sample in the `java` file, if the template file is up to date. Could you add [`list_all_indexes_1`](https://github.com/meilisearch/documentation/blob/c1e0d5d0522848f845890ce34888a88a89dbcef0/.code-samples.meilisearch.yaml#L9) in your code-samples file?

curquiza:
@ansavanix it means you have to change the order in the `.vuepress/code-samples/sdks.json` file

curquiza:
> @dichotommy, it's working fine on local now. And we have the arrow 😃 not sure I like it.

Maybe replacing `JavaScript` by `JS`

PR #1211, opened 2021-10-19
Rename markdownlint-cli config file

guimachiavelli:
This repo currently uses markdownlint-cli to ensure there's something like a consistent usage of markdown syntax across documents. Configuration for markdownlint-cli can take several shapes; the one we currently use is storing everything on a `markdownlintrc.jsonc` file.

This PR renames `markdownlintrc.jsonc` to `markdownlint.jsonc`. It also updates the `package.json` file so markdownlint behaviour (and thus our CI pipeline) remains unaffected.

## Rationale

We use `markdownlint-cli` in two ways: as a `yarn` script we manually run before committing changes and as part of our CI tasks.

VSCode, the de facto standard editor in our workflow, gives us the option of installing [an extension](https://github.com/DavidAnson/vscode-markdownlint) that also lints documents on write. This extension, however, does not support `markdownlintrc.jsonc` files, only (confusingly) `markdownlint.jsonc` files (notice the missing `rc`?)


guimachiavelli:
Just yesterday I was imagining someone being angry at you and calling you Thomas Longbottom Melvin (don't ask, I'm not responsible for the random stuff that pops up when my brain is in screensaver mode). I think that commit message supports this scenario.

guimachiavelli:
bors merge

PR #1222, opened 2021-10-21
Update `attribute` description

curquiza:
A part of the `attribute` criterion role was missing


curquiza:
Nope `words` is not related to `attribute` at all.

Issue #1190, opened 2021-10-12
v0.25: Task API


guimachiavelli:
The Update API is used to handle asynchronous processes throughout MeiliSearch. It is particularly common when dealing with an index's settings or adding new documents — since both operations can take a significant amount of time, they are processed separately and leave the main thread free for other tasks.

This API's name is a common source of confusion, since a user can update an index's key, its documents, or just use the word to refer to changing something.

For this reason, v0.25 renames Updates to Task. Full information can be found on the spec linked below.

---

## Tasks

- [x] Update API reference (all routes)
- [x] Update `/learn/asynchronous_updates.md`
- [x] Search for other mentions of `Update` throughout the documentation and update them when necessary
- [x] Search for `Update` responses (e.g. https://docs.meilisearch.com/create/how_to/updating.html)
- [x] Check whether the `updateId` object returned when triggering an async update remains the same. Very likely it changes to something like `taskId`. See above example for an example of `updateId`
- [x] Update `update` objects so they match the `task` schema
- [x] Add new error to `errors.yaml`: `task_not_found`
- [x] Update configuration option: `--max-udb-size` to `--max-task-db-size` (don't forget its corresponding env var)
- [x] Update telemetry page with new collected data
- [x] Confirm with product/core that none of these changes influence the Dump API

---

## Reference
- https://github.com/meilisearch/specifications/blob/develop/text/0060-refashion-updates-apis.md
- https://github.com/meilisearch/MeiliSearch/issues/1780

maryamsulemani97:
@guimachiavelli @dichotommy 
For [this](https://docs.meilisearch.com/learn/advanced/asynchronous_updates.html#terminate-meilisearch-while-a-task-is-being-processed) section of the Asynchronous updates article:
- the term "task" refers to any task and not just updates
- the term "update task" refers to what is now "task"
What would be a better alternative for the old "task" term? "Process" or "Operation"

dichotommy:
> the term "task" refers to any task and not just updates
> the term "update task" refers to what is now "task"

I'm not sure that this was an intentional distinction, but I think the way you've defined it is useful nonetheless.

>   What would be a better alternative for the old "task" term? "Process" or "Operation"

I like "asynchronous operation" as an umbrella term. To my knowledge, the only asynchronous operation we have that isn't a "task" is dump creation.

PR #1224, opened 2021-10-21
v0.25: Async Indexes API

### LINKED ISSUE #1191: v0.25: Index API's POST, PUT, and DELETE are now async
The `POST`, `PUT`, and `DELETE` routes of the index API respectively handle index creation, configuration of undefined primary keys, and deletion of indexes. These routes are currently synchronous, which can cause errors and unexpected behavior if used in combination with asynchoronous processes (e.g. trigger indexation and then use the `DELETE` endpoint).

v0.24 changes this behavior and makes all three routes asynchronous. `GET` remains unaltered.

---

## Tasks

- [ ] Update `/reference/API/indexes.md`
- [ ] Check whether this affects other documents (`/learn/getting_started/quick_start.md`, perhaps?)

---

## Reference
- https://github.com/meilisearch/MeiliSearch/issues/1786
- some information is available on the Task API spec: https://github.com/meilisearch/specifications/blob/develop/text/0060-refashion-updates-apis.md#8-asynchronous-write-operations-on-index-resource

guimachiavelli:
closes #1191


reference/api/indexes.md
```diff
@@ -64,15 +64,19 @@ Get information about an [index](/learn/core_concepts/indexes.md).
 
 <RouteHighlighter method="POST" route="/indexes"/>
 
-Create an [index](/learn/core_concepts/indexes.md).
+Create an [index](/learn/core_concepts/indexes.md). This route accepts two arguments: `index_uid` and `primaryKey`.
 
-This route takes as parameter an unique `uid` and **optionally** the [primary key](/learn/core_concepts/indexes.md#primary-key).
+`index_uid` is mandatory. It should be a unique string identifying your new index.
```
gmourier, line 69:
```suggestion
`uid` is mandatory. It should be a unique string identifying your new index.
```

When creating an index resource, the field name for naming the index (his unique identifier) is still named `uid`.

The `202` response does not contain an `index` resource but a summarized `task` resource. 

That's why the `task` resource contains an `indexUid` to give information that a `task` has been enqueued to create that index and will belong to this index.

_**Generally speaking**_, when we design an API resource, we avoid putting its name in the fields that concern it.

On the other hand, it is advisable to contextualize fields linked to another resource by a prefix of the resource name or an object bearing the name of the linked resource. As in the case of a `task` that is linked to an `index`, it could have been: 

```json
{
    "uid": 0, //task uid
    "index": {
        "uid": "movies"
    },
    ...
}
```

But we chose this format since we don't expect to give more information from the `index` related to a `task`.

```json
{
    "uid": 0,
    "indexUid": "movies",
    ...
}
```

reference/api/indexes.md
```diff
@@ -64,15 +64,19 @@ Get information about an [index](/learn/core_concepts/indexes.md).
...
+
+`primaryKey` is optional and indicates the field containing your documents' [primary key](/learn/core_concepts/indexes.md#primary-key). If you do not supply a value for `primaryKey`, MeiliSearch will do its best to infer it once you add your first document to the index.
```
gmourier, line 71:
```suggestion
`primaryKey` is optional and indicates the field containing your documents [primary key](/learn/core_concepts/indexes.md#primary-key). If you do not supply a value for `primaryKey`, MeiliSearch will do its best to infer it once you add your first document to the index.
```

reference/api/indexes.md
```diff
@@ -90,62 +94,73 @@ An index is automatically created when adding [documents](/reference/api/documen
...
-  "updatedAt": "2019-11-20T10:16:42.761858Z"
+  "uid": 1,
+  "indexUid": "movie_review",
+  "status": "enqueued",
+  "type": "indexUpdate",
+  "enqueuedAt": "2021-08-12T10:00:00.000000Z"
 }
 ```
 
+You can use the response's `uid` to track the status of your request.
```
gmourier, line 153: It would be nice to give a link to the API reference to get the task and watch the status.

gmourier (changes requested):
Some minor changes, the only important thing is the API expects `uid` instead `index_uid` for the request payload for `POST - /indexes` 🤓

reference/api/indexes.md
```diff
@@ -64,15 +64,19 @@ Get information about an [index](/learn/core_concepts/indexes.md).
...
+
+`primaryKey` is optional and indicates the field containing your documents' [primary key](/learn/core_concepts/indexes.md#primary-key). If you do not supply a value for `primaryKey`, MeiliSearch will do its best to infer it once you add your first document to the index.
 
 ::: note
-An index is automatically created when adding [documents](/reference/api/documents.md) or [settings](/reference/api/settings.md) to an index that does not already exist.
+If you try to add [documents](/reference/api/documents.md) or [settings](/reference/api/settings.md) to an index that does not exist, MeiliSearch will automatically create it for you. This is called implicit index creation.
 :::
 
-#### Body
+Creating an index is an asynchronous task. [You can read more about asynchronous processes in our dedicated guide.](/learn/advanced/asynchronous_updates.md)
```
maryamsulemani97, line 77: Can we go with the term "[asynchronous operation](https://github.com/meilisearch/documentation/issues/1190#issuecomment-947604634)" instead of "asynchronous process"? It's what I used used in the Task API PR

reference/api/indexes.md
guimachiavelli: Yeap. I'm until that PR is approved so I can point the link to the correct page/heading.

reference/api/indexes.md
guimachiavelli: Sure, no problem 👍

reference/api/indexes.md
gmourier: No, my English level just thought it was a typo. 💩

dichotommy, line 71: Personally I'm not quite convinced that these paragraphs add enough value to justify the screen real estate; we already have the `Body` table below. Instead of ballooning the description of the endpoint could we incorporate whatever information isn't already present into that table?

(FWIW, in the long run, I am anti-table and hoping to transform a lot of these into paragraphs, much like [Stripe options](https://stripe.com/docs/js/elements_object/create#stripe_elements-options))

dichotommy (changes requested):
Great work 💯  Hope you don't mind my few nitpicks.

reference/api/indexes.md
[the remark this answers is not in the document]
guimachiavelli: Chibi MeiliSearch is great. I name you head vice-director of branding and marketing-adjacent subjects.

Jokes aside, though I do get your point, making primary key inferral come across as unreliable was actually sort of intentional. My understanding is that we are not supposed to encourage people to rely too much on inferring the primary key since we cannot guarantee it will work.

reference/api/indexes.md
```diff
@@ -64,19 +64,21 @@ Get information about an [index](/learn/core_concepts/indexes.md).
 
 <RouteHighlighter method="POST" route="/indexes"/>
 
-Create an [index](/learn/core_concepts/indexes.md).
+Create an [index](/learn/core_concepts/indexes.md). This endpoint accepts two arguments: `index_uid` and `primaryKey`.
```
maryamsulemani97, line 67:
```suggestion
Create an [index](/learn/core_concepts/indexes.md). This endpoint accepts two arguments: `uid` and `primaryKey`.
```

maryamsulemani97 (approved):
🍪

guimachiavelli:
bors merge

guimachiavelli:
bors merge

Issue #1193, opened 2021-10-12
v0.24: Update errors


guimachiavelli:
When an error occurs in MeiliSearch, users will be notified via error messages in e.g. their command line.

v0.24 updates the error object, most error descriptions and a few error codes. We need to update the docs to reflect these changes.

---

## Tasks

- [x] Update error codes and descriptions on `errors.yaml`
- [x] Search for and update any error object samples with the new schema

---

## Reference
- https://github.com/meilisearch/specifications/blob/474bfba29088346232e087c1569eae757ab19ffd/text/0061-error-format-and-definitions.md

PR #1227, opened 2021-10-25
v0.24-update errors

maryamsulemani97:
closing #1193

### DIFF
--- a/errors.yaml
+++ b/errors.yaml
@@ -25,7 +25,7 @@ errors:
   - code: invalid_document_id
     description: "The provided document identifier does not meet the format requirements. A document identifier must be of type integer or string, composed only of alphanumeric characters (a-z A-Z 0-9), hyphens (-), and underscores (_)."
   - code: invalid_filter
-    description: "The filter provided with the search is invalid. This maybe due to syntax errors in the `filter` parameter, using reserved fields as filter expressions, or not adding the filtering attributes to `filterableAttributes`. Check our [guide on filtering](https://docs.meilisearch.com/reference/features/filtering_and_faceted_search.html) to troubleshoot."
+    description: "The filter provided with the search is invalid. This may be due to syntax errors in the `filter` parameter, using reserved fields as filter expressions, or neglecting to add the filtering attributes to `filterableAttributes`. For troubleshooting, check our [guide on filtering](https://docs.meilisearch.com/reference/features/filtering_and_faceted_search.html)."
   - code: invalid_sort
     description: "The sort value is invalid. This maybe due to syntax errors in the `sort` parameter, using reserved fields as sort expressions, or not adding the sorting attributes to `sortableAttributes`. Check our [guide on sorting](https://docs.meilisearch.com/reference/features/sorting.html) for more information."
   - code: bad_request
--- a/errors.yaml
+++ b/errors.yaml
@@ -15,7 +15,7 @@ errors:
   - code: missing_primary_key
     description: "All documents must be identified with a unique key. MeiliSearch will try to infer the document field to use it as a primary key. If it can't infer it, the primary key must be specified manually, and this error is returned. See  our [guide on primary keys](https://docs.meilisearch.com/learn/core_concepts/documents.html#structure)."
   - code: primary_key_inference_failed 
-    description: "The provided document does not contain any fields with the substring `id`. [Manually designate the primary key](https://docs.meilisearch.com/learn/core_concepts/documents.md#setting-the-primary-key) or add `id` to one of your fields so it can be used as the primary key during inference."
+    description: "The first provided document contains no fields with the substring `id`. [Manually designate the primary key](https://docs.meilisearch.com/learn/core_concepts/documents.md#setting-the-primary-key) or add `id` to one of your fields so it can be used as the primary key during inference. We recommend manually setting the primary key."
   - code: primary_key_already_present
     description: "This error is thrown when trying to modify the primary key of an index, when it has been already set. See [guide on primary keys](https://docs.meilisearch.com/learn/core_concepts/documents.html#structure)."
   - code: document_fields_limit_reached
--- a/errors.yaml
+++ b/errors.yaml
@@ -12,12 +12,8 @@ errors:
     description: "An internal error occurred while trying to access the requested index."
   - code: invalid_state
     description: "The database is in an invalid state. Deleting the database and re-indexing should solve the problem."
-  - code: missing_primary_key
-    description: "All documents must be identified with a unique key. MeiliSearch will try to infer the document field to use it as a primary key. If it can't infer it, the primary key must be specified manually, and this error is returned. See  our [guide on primary keys](https://docs.meilisearch.com/learn/core_concepts/documents.html#structure)."
   - code: primary_key_inference_failed 
     description: "The first provided document contains no fields with the substring `id`. [Manually designate the primary key](https://docs.meilisearch.com/learn/core_concepts/documents.md#setting-the-primary-key) or add `id` to one of your fields so it can be used as the primary key during inference. We recommend manually setting the primary key."
-  - code: primary_key_already_present
-    description: "This error is thrown when trying to modify the primary key of an index, when it has been already set. See [guide on primary keys](https://docs.meilisearch.com/learn/core_concepts/documents.html#structure)."
   - code: document_fields_limit_reached
     description: "A document exceeds the maximum limit of 65,535 fields."
   - code: missing_document_id
@@ -34,8 +30,6 @@ errors:
     description: "The requested document can't be retrieved. Either it doesn't exist, or the database was left in an inconsistent state."
   - code: internal
     description: "MeiliSearch experienced an internal error. Check the error message, and open an issue if necessary."
-  - code: invalid_token
-    description: "The provided token is invalid."
   - code: missing_authorization_header
     description: "The requested resources are protected with an API key, that was not provided in the request header. Check our guide on [authentication](https://docs.meilisearch.com/reference/features/authentication.html) for more information."
   - code: not_found
--- a/errors.yaml
+++ b/errors.yaml
@@ -70,4 +70,6 @@ errors:
     description: "The provided document does not contain any field with the substring `id`. Please add `id` to one of your fields so it can be used as the primary key during inference."
   - code: index_primary_key_already_exists
     description: "The requested index already has a primary key that cannot be changed."
+  - code: invalid_document_id
+    description: "The provided document identifier does not meet the format requirements. A document identifier can be of type integer or string, only composed of alphanumeric characters (a-z A-Z 0-9), hyphens (-) and underscores (_)."
 ...
--- a/errors.yaml
+++ b/errors.yaml
@@ -59,7 +59,7 @@ errors:
   - code: dump_process_failed
     description: "An error occurred during dump creation process, task aborted."
   - code: invalid_ranking_rule
-    description: "The provided ranking rules are invalid. This maybe due to syntax errors in your ranking rules or specifying custom ranking rules on reserved fields like `_geo` and `_geoDistance` or reserved expressions like `_geoPoint` and `_geoRadius`. For more information on ranking rule configuration, check our [relevancy guide](https://docs.meilisearch.com/learn/core_concepts/relevancy.html#ranking-rules)"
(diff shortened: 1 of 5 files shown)


gmourier (commented):
The definition of this [error](https://github.com/meilisearch/specifications/blob/develop/text/0061-error-format-and-definitions.md#task_not_found) is missing but otherwise, everything looks good from my side. Thanks, @maryamsulemani97 !

maryamsulemani97:
> The definition of this [error](https://github.com/meilisearch/specifications/blob/develop/text/0061-error-format-and-definitions.md#task_not_found) is missing but otherwise, everything looks good from my side. Thanks, @maryamsulemani97 !

@gmourier Weren't we supposed to exclude this as it will be part of the next release?

gmourier:
> > The definition of this [error](https://github.com/meilisearch/specifications/blob/develop/text/0061-error-format-and-definitions.md#task_not_found) is missing but otherwise, everything looks good from my side. Thanks, @maryamsulemani97 !
> 
> @gmourier Weren't we supposed to exclude this as it will be part of the next release?

Thank you @maryamsulemani97. I don't see any problem shifting its presence to the next release. We'll just have to be careful not to forget to update the errors page 😇.

gmourier (approved):
Thanks, @maryamsulemani97. LGTM 🚀

errors.yaml
```diff
@@ -1,29 +1,33 @@
...
   - code: index_not_accessible
     description: "An internal error occurred while trying to access the requested index."
   - code: invalid_state
-    description: "An error that occurs when the database is in an invalid state. Deleting the database and re-indexing should solve the problem."
+    description: "The database is in an invalid state. Deleting the database and re-indexing should solve the problem."
   - code: missing_primary_key
     description: "All documents must be identified with a unique key. MeiliSearch will try to infer the document field to use it as a primary key. If it can't infer it, the primary key must be specified manually, and this error is returned. See  our [guide on primary keys](https://docs.meilisearch.com/learn/core_concepts/documents.html#structure)."
+  - code: primary_key_inference_failed 
+    description: "The provided document does not contain any field with the substring `id`. Please add `id` to one of your fields so it can be used as the primary key during inference, or [manually designate the primary key](/learn/core_concepts/documents.md#setting-the-primary-key)."
```
guimachiavelli, line 18: Honest question: should the first `field` be plural or singular? Plural feels more correct to my ear, but I'm really not sure.

A second comment: isn't it better to recommend users to manually set the primary key field first? Seems like the easier solution since it doesn't require the dev to mess around with the dataset.

A third and very important change: as far as I understand, `errors.yaml` is not parsed in the same way as `*.md`, meaning that links need to be hardcoded URLs to the compiled vuepress output. Can you confirm that, @dichotommy?

```suggestion
    description: "The provided document does not contain any fields with the substring `id`. [Manually designate the primary key](https://docs.meilisearch.com/learn/core_concepts/documents.md#setting-the-primary-key) or add `id` to one of your fields so it can be used as the primary key during inference."
```

errors.yaml
```diff
@@ -48,12 +52,24 @@ errors:
     description: "The Content-Type header does not match the request body payload format or the format is invalid."
   - code: missing_payload
     description: "The Content-Type header was specified, but no request body was sent to the server or the request body is empty."  
-  - code: dump_already_in_progress
-    description: "A dump creation is already in progress and a new one can't be triggered until the previous dump creation is not finished."
+  - code: dump_already_processing
+    description: "Dump creation is already in progress. A new dump creation process can be triggered when the current one is complete."
```
guimachiavelli, line 56: Minor rewording suggestion, don't feel strongly about it:

```suggestion
description: "Dump creation is already in progress. A new dump creation process can be triggered after the current one has been completed."
```

errors.yaml
```diff
@@ -48,12 +52,24 @@ errors:
...
   - code: invalid_geo_field
     description: "The provided `_geo` field of one or more documents is invalid. Read more about `_geo` and how to troubleshoot it in [our dedicated guide](https://docs.meilisearch.com/reference/features/geosearch.html)."
+  - code: invalid_api_key
+    description: "The requested resources are protected with an API key. The provided API key is invalid. Read more about it at in our [dedicated guide](https://docs.meilisearch.com/reference/features/authentication.html)."
+  - code: invalid_store_file
+    description: "The `data.ms` folder is in an invalid state. Your `.mdb` file is corrupted or the `data.ms` folder has been replaced by a file."
+  - code: no_space_left_on_device
+    description: "The host system partition has reached its maximum capacity and can no longer accept writes." 
+  - code: database_size_limit_reached
+    description: "The requested database has reached its maximum limit; no further documents can be added. For information on customizing the maximum database size accepted by MeiliSearch, check out this [guide](https://docs.meilisearch.com/reference/features/configuration.html#max-udb-size)."
```
guimachiavelli, line 72: I know it's on the error code, but perhaps it's good to specify it's the size limit we're talking about?

```suggestion
    description: "The requested database has reached its maximum size; no further documents can be added. For information on customizing the maximum database size accepted by MeiliSearch, check out this [guide](https://docs.meilisearch.com/reference/features/configuration.html#max-udb-size)."
```

errors.yaml
```diff
@@ -48,12 +52,24 @@ errors:
...
+  - code: index_primary_key_already_exists
+    description: "The requested index already has a primary key that cannot be changed."
```
guimachiavelli, line 74: Are there primary keys that can be changed?

resources/faq.md
```diff
@@ -105,6 +105,16 @@ cat your_file.json | jq
...
+## What do the different error types mean?
+
+MeiliSearch has the following types of errors:
+
+| Type            | Description                                                                                       |
+|-----------------|---------------------------------------------------------------------------------------------------|
+| invalid_request | This is due to an error in the user input. It is accompanied by the HTTP code `4xx`. |
+| internal        | This is due to machine or configuration constraints. The most common cause is reaching or exceeding hard limits, such as the size of the disk, the size limit of an index, etc. It is accompanied by the HTTP code `5xx`.  |
+| auth   | This type of error is related to authentication and authorization. It is accompanied by the HTTP code `4xx`. |
+
```
guimachiavelli, line 117: Trying to format the table so all `|` line up

```suggestion
| Type            | Description                                                                                       |
|-----------------|---------------------------------------------------------------------------------------------------|
| invalid_request | This is due to an error in the user input. It is accompanied by the HTTP code `4xx`. |
| internal        | This is due to machine or configuration constraints. The most common cause is reaching or exceeding hard limits, such as the size of the disk, the size limit of an index, etc. It is accompanied by the HTTP code `5xx`.  |
| auth            | This type of error is related to authentication and authorization. It is accompanied by the HTTP code `4xx`. |
```

guimachiavelli (commented):
Thanks for this!

Just a few minor suggestions!

errors.yaml
maryamsulemani97: Not that I know of.
```suggestion
    description: "The requested index already has a primary key and cannot be changed."
```

errors.yaml
maryamsulemani97: > isn't it better to recommend users to manually set the primary key field first? Seems like the easier solution since it doesn't require the dev to mess around with the dataset.

Do we need to suggest that in the error message? Or can we do that [here](https://docs.meilisearch.com/learn/core_concepts/documents.html#primary-field)?

dichotommy: 1. I believe the word field here can be singular or plural. Could be wrong OFC. Probably better just to rephrase it:

```suggestion
    description: "The first provided document contains no fields with the substring `id`. [Manually designate the primary key](https://docs.meilisearch.com/learn/core_concepts/documents.md#setting-the-primary-key) or add `id` to one of your fields so it can be used as the primary key during inference."
```

2. I could be wrong here, but I took the addition of this error as part of a move towards primary key inference.

The current distinction between manual and inferred primary keys is difficult to explain to users. It's unclear what the benefits are of manually setting a primary key vs. allowing it to be inferred. There are two possible solutions to this that I see: either committing to inference, or committing to manual primary keys.

I believe we are heading in the direction of removing manual primary keys, and instead going full inference, i.e. you're fine if your dataset already contains a field with the substring `id`, and otherwise you should add one for MeiliSearch to use. This obviously would introduce some friction to using MeiliSearch (requires the extra step of adding an `id` field to some datasets) but would greatly clarify the documentation/understanding side, since users would know exactly what to expect and, when sharing datasets (e.g. for troubleshooting), the primary key would be much easier to identify.

@gmourier Which direction is the core product team leaning right now?

3. I believe you are correct Gui! I haven't tested it myself, but the fact that all links currently in `errors.yaml` are hardcoded suggests to me that it is the case.

errors.yaml
gmourier: Currently, if an index already has a primary key, it cannot be updated. 

In v0.24.0, this primary key can be updated if the index contains no document. In the rare case where there are 0 documents, this will reduce the current friction which is to delete the index and recreate it to change the primary key.

This is a change that was made at the end of last week very easily but which was not foreseen in the specification. However, as it can improve the user's life in a specific case, we decided to keep this change.

We didn't want to redefine a more precise error message for the moment on the core side but it will be done later. For the moment the message here is consistent because this error will occur only in case the primary key can't be changed.

Depending on what we decide to do about inference, waiting for a specific field like `id`, or being able to change the value of the primary key, it can change.

errors.yaml
gmourier: Very good question @dichotommy, thanks for asking! 

I personally believe that our inference process is not good today for some specific cases. Analytics will allow us to verify the hypotheses that we have. 

For example, if the majority of people use `id` we could completely remove the inference and default to the `id` field when no primary key has been set on the index in the first place.

We could also strictly expect an `id` attribute as most databases do today and generate a unique value when the users do not provide it in their documents.

This adds friction in a way but this friction will be quickly reached and allows to solve other hidden frictions.

For example: 

The user has no `id` field in his document, we generate it automatically and he can have a first coherent and functional search whereas today if the inference is missed and we choose a field containing `id` before `id` (which it's the primary key that he wanted at first), he ends up with a completely inconsistent first search.

The Aha moment is gone and he won't see all the value we could have brought directly in his first use of the product.

If he hasn't gone elsewhere and insists, this is what happens: (The condition is that he must have understood by himself that the search is not in a good shape because of a bad inference)

He will have to check the document, find how to delete an index, delete it, recreate it, find that he can set a primary key to disable the inference. Long way to go, isn't it?

I have very strong hypotheses about this but I wait to be able to check them globally before making a decision about this subject.

guimachiavelli: Ok, so, first, let's summarise this into actionable things:

1. I like Tommy's suggestion, so if you agree with my comment please ignore my suggestion and apply his;
2. Given inference behaviour is likely to change in the future, I'd say it is simpler to emphasise manually setting the primary key right now. The user already had to deal with one error, best to give them an unambiguous instruction that will definitely solve the problem;
3. I just tested this and the link on the preview leads to a 404. Absolute URLs it is (only inside `errors.yaml`, `md` files should continue using relative links.

Do we have consensus on these three items? This is absolutely not a rhetorical question, please correct me if I got anything wrong.

---

> Do we need to suggest that in the error message? Or can we do that [here](https://docs.meilisearch.com/learn/core_concepts/documents.html#primary-field)?

@maryamsulemani97, I personally like the idea of giving users a hand and pointing them to a solution already on the errors page. If you prefer to do it on `core_concepts/documents`, though, I  think it's best to open another PR since this one is about error formats.

---

@gmourier thanks for the super detailed answer! It will be interesting to see what will come out of the analytics. I personally always find inference the type of thing that can help beginners but gets in the way as soon as you're trying to get some work done. As I mentioned, as a front-end dev I often received datasets over which I had no control; the thought of having to add a preprocessing step just so a third-party tool can try to infer primary keys would be really unappealing to me in that situation. Of course, these are my 0.02¥ and do not necessarily reflect Meili's users' needs.

gmourier: Indeed @guimachiavelli, we could let the user choose his primaryKey if he doesn't want to use `id`, this way, he is not forced to modify the dataset he receives and which could have an identifier with a different name than `id`. There are several possible solutions that I think are sufficient to guide the user in the right way when he first uses it, and thus eliminate the somewhat twisted cases that we can encounter today.

guimachiavelli (approved):
🦄

gmourier (approved):
Thanks, @maryamsulemani97 🚀

guimachiavelli:
bors merge

guimachiavelli:
bors merge

PR #1238, opened 2021-10-27
V0.24: attributesToHighlight

guimachiavelli:
In v0.24, `attributesToHighlight` works on both string and numeric values.

This PR updates the text to reflect that change and generally tries to improve wording on this search parameter's reference.

### DIFF
--- a/reference/features/search_parameters.md
+++ b/reference/features/search_parameters.md
@@ -299,18 +299,18 @@ If `attributesToCrop` is not configured, `cropLength` has no effect on the retur
 **Expected value**: an array of <clientGlossary word="attribute" label="attributes" /> or `["*"]`
 **Default value**: `null`
 
-Highlights matching query terms in the specified attributes. Both string and numeric values, whether inside arrays and objects or not, are highlighted.
-
-Parameter values must be given as an array of attributes: `attributesToHighlight=["attributeA", "attributeB"]`.
-
-Alternatively, you can provide `["*"]` as a value: `attributesToHighlight=["*"]`. In this case, all the attributes present in `attributesToRetrieve` will be assigned to `attributesToHighlight`.
+Highlights matching query terms in the specified attributes by enclosing them in `<em>` tags. `attributesToHighlight` highlights matches in fields containing strings, numeric values, arrays, and objects.
 
 When this parameter is set, returned documents include a `_formatted` object containing the highlighted terms.
 
-::: tip
-The highlighting performed by this parameter consists of wrapping matching query terms in `<em>` tags. It is not possible to change the tag or its attributes.
+Parameter values must be given as an array of attributes: `attributesToHighlight=["attributeA", "attributeB"]`.
 
-If you need finer control over the highlights and their format, we recommend using [the `matches` search parameter](#matches).
+Alternatively, you can provide `["*"]` as a value: `attributesToHighlight=["*"]`. In this case, all the attributes present in [`attributesToRetrieve`](/reference/features/search_parameters.md#attributes-to-retrieve) will be assigned to `attributesToHighlight`.
+
+::: tip
+It is not possible to change the `<em>` tag or its attributes.
+
+If you need finer control over the formatted output, we recommend using [the `matches` search parameter](#matches).
 :::
 
 ### Example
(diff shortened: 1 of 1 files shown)


reference/features/search_parameters.md
```diff
@@ -299,25 +299,27 @@ If `attributesToCrop` is not configured, `cropLength` has no effect on the retur
 **Expected value**: an array of <clientGlossary word="attribute" label="attributes" /> or `["*"]`
 **Default value**: `null`
 
-Highlights matching query terms in the given attributes. When this parameter is set, the `_formatted` object is added to the response for each document, within which you can find the highlighted text.
+Highlights matching query terms in the specified attributes by enclosing them in `<em>` tags. `attributesToHighlight` highlights matches in fields containing strings, numeric values, arrays, and objects.
```
dichotommy, line 302: "Numeric value" is inconsistent IMO since in a MeiliSearch context these are all "values".

```suggestion
Highlights matching query terms in the specified attributes by enclosing them in `<em>` tags. `attributesToHighlight` only works on values of the following types: string, number, array, object.
```

reference/features/search_parameters.md
```diff
@@ -299,25 +299,27 @@ If `attributesToCrop` is not configured, `cropLength` has no effect on the retur
...
 
-Values can be supplied as an array of attributes: `attributesToHighlight=["attributeA", "attributeB"]`.
+When this parameter is set, returned documents include a `_formatted` object containing the highlighted terms.
 
-Alternatively, you can provide `["*"]` as a value: `attributesToHighlight=["*"]`. In this case, all the attributes present in `attributesToRetrieve` will be assigned to `attributesToHighlight`.
+Parameter values must be given as an array of attributes: `attributesToHighlight=["attributeA", "attributeB"]`.
```
dichotommy, line 306: Is it necessary to include this information given the "Expected value" field above?

reference/features/search_parameters.md
```diff
@@ -299,25 +299,27 @@ If `attributesToCrop` is not configured, `cropLength` has no effect on the retur
...
 
 ### Example
 
-If you wanted to highlight query matches that appear within the `overview` attribute:
+The following query highlights matches present in the `overview` attribute:
 
 <CodeSamples id="search_parameter_guide_highlight_1" />
 
-You would get the following response with the **highlighted version in the `_formatted` object**:
+The highlighted version can then be found in the returned document's `_formatted` object:
```
dichotommy, line 322:
```suggestion
The highlighted version of the text would then be found in the `_formatted` object included in each returned document:
```

dichotommy (changes requested):

reference/features/search_parameters.md
guimachiavelli: Not strictly necessary, no, but perhaps useful anyhow because we present an example? Don't feel strongly about it either way—if you think it's best to remove it, I'm happy to do it.

reference/features/search_parameters.md
```diff
@@ -299,25 +299,25 @@ If `attributesToCrop` is not configured, `cropLength` has no effect on the retur
 **Expected value**: an array of <clientGlossary word="attribute" label="attributes" /> or `["*"]`
 **Default value**: `null`
 
-Highlights matching query terms in the given attributes. When this parameter is set, the `_formatted` object is added to the response for each document, within which you can find the highlighted text.
+Highlights matching query terms in the specified attributes by enclosing them in `<em>` tags. `attributesToHighlight` only works on values of the following types: string, number, array, object.
```
dichotommy, line 302: We should probably make this a list, right?
```suggestion
Highlights matching query terms in the specified attributes by enclosing them in `<em>` tags. 

`attributesToHighlight` only works on values of the following types:

- string
- number
- array
- object
```

guimachiavelli: Isn't this giving a disproportionate amount of attention to a list of types not seen in other places in the docs?

Perhaps that's because I'm not on team "always use lists when you have more than two items"?

dichotommy: Yeah, you could be right. I'm happy to leave this change out.

dichotommy (approved):
LGTM 🚀

guimachiavelli:
bors merge

Issue #858, opened 2021-03-10
Clarify attribute ranking order example


dichotommy:
In the [attribute ranking order example](https://docs.meilisearch.com/learn/core_concepts/relevancy.html#example-2) it isn't clear that the impact of the rule is determined by ITS position in the ranking-rules list, e.g. if attribute is last in the ranking rules list, then attribute ranking order will have effectively no impact.

In other words, we need to draw a clearer connection between the "attribute" ranking-rule and the concept of the attribute ranking order.

maryamsulemani97:
closed in PR #1239

PR #1239, opened 2021-10-28
Clarify attribute ranking order example

maryamsulemani97:
closing #858

### DIFF
--- a/learn/core_concepts/relevancy.md
+++ b/learn/core_concepts/relevancy.md
@@ -196,5 +196,5 @@ For a more detailed look at this subject, see our reference page for [the search
 With the above attribute ranking order, matching words found in the `title` field would have a higher impact on relevancy than the same words found in `description` or `release_date`. If you searched "1984", for example, results like Michael Radford's film "1984" would be ranked higher than movies released in the year 1984.
 
 :::note
-Please keep in mind that **the order of the results depends on the position of the `attribute` rule in the [`rankingRules`](#default-order) object.** Meaning, **if `attribute` is at the bottom of the ranking rules list, it will have no impact.**
+The `attribute` rule's position in [`rankingRules`](#default-order) determines how the results are sorted. Meaning, **if `attribute` is at the bottom of the ranking rules list, it will have almost no impact on your search results.**
 :::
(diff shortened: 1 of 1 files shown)


learn/core_concepts/relevancy.md
```diff
@@ -194,3 +194,7 @@ For a more detailed look at this subject, see our reference page for [the search
 ```
 
 With the above attribute ranking order, matching words found in the `title` field would have a higher impact on relevancy than the same words found in `description` or `release_date`. If you searched "1984", for example, results like Michael Radford's film "1984" would be ranked higher than movies released in the year 1984.
+
+:::note
+Please keep in mind that **the order of the results depends on the position of the `attribute` rule in the [`rankingRules`](#default-order) object.** Meaning, **if `attribute` is at the bottom of the ranking rules list, it will have no impact.**
```
guimachiavelli, line 199: In my opinion, too much emphasis is almost the same as no emphasis. Like, it should be something that tells the reader what is the most important thing about a specific section of text, so if we say everything's important, well, we're not helping the reader find the crucial stuff more quickly.

Also suggesting a few minor wording changes, but they're not super important.

```suggestion
The order of the results depends on the position of the `attribute` rule in the [`rankingRules`](#default-order) array. Meaning, **if `attribute` is at the bottom of the ranking rules list, it will have almost no impact on your search results.**
```

guimachiavelli (changes requested):
Requesting a few changes.

Also, is it just me or is there an abundance of `the`s on the first sentence? Not sure we can do anything about it.

learn/core_concepts/relevancy.md
dichotommy: I think this wording change is important, since it will have an impact on your search results, just a very minor one (tie-breaking documents that were equally relevant under all the previous ranking rules, as I understand it).

dichotommy:
> Also, is it just me or is there an abundance of thes on the first sentence?

Now that you've pointed it out, I can't un-hear it. The word "the" is beginning to sound like strange nonsense. plz send help

guimachiavelli (approved):
LGTM!
Great job on that sentence 🥇

dichotommy (approved):
🥳

dichotommy:
bors merge

PR #1244, opened 2021-11-02
Update storage.md

jellevdvliet:
I assume the virtual memory usage is only 205Mb instead of 205Gb?


guimachiavelli:
@curquiza and @gmourier, can either of you confirm this is a typo? 205GB does sound…excessive 😅

curquiza:
Checked with @Kerollmops 

This is not a typo: this is the virtual memory and not the real physic memory that is allocated. MeiliSearch could take 205Gb of memory but does not really immediately: your OS will manage the physical memory it really allocates for MeiliSearch on its own.

curquiza (changes requested):
👆

jellevdvliet:
Good to know. I think the docs should be expanded then to explain the 205Gb because it definitely looks like a typo 😅 .

PR #1245, opened 2021-11-02
Add dotnet code samples

curquiza:


guimachiavelli (approved):
I'm pre-emptively approving this, but we might have to start re-thinking the code sample widget.

With C# we are officially unable to display all SDKs at the same time and rely on users noticing they can use the tiny `>` button to scroll horizontally.

Pinging @dichotommy because he recently handled an addition to the code samples.

curquiza:
FYI, I make sure the less used ones (we supposed) are the last ones -> the swift and dart SDKs.

I know it's not really nice to have the `>` but I don't think it should prevent merging this PR since it's a real feature for the dotnet users, that are quite many 🙂

curquiza:
Also, currently the dotnet code samples are only present on the Getting Started.

dichotommy:
Hello hello @curquiza , super excited to be adding C# code samples to the docs. I also agree completely that this minor display problem should not stop us from merging this PR.

Still, if possible, it would be great to find a simple CSS solution for the pesky arrow. Theoretically, we don't need to completely resolve the problem—just prevent it from impacting users until we migrate to Docusaurus. At that time we will certainly have to recreate the code samples plugin in React, and may in fact prefer to completely redesign it since, as @cmaneu pointed out, the trend is towards page-level language selection.

VuePress's default `content-width` is `740px`. I have experimented with modifying the screen space used by the default VuePress theme, adding the following line to `.vuepress/styles/palette.styl`:

```
$contentWidth = 820px
```

Maximized on my 16" MacBook Pro, it looks like this:

Screen Shot 2021-11-03 at 4 41 28 PM

Of course, users with really low-resolution screens or those who keep the docs in smaller windows could still see the arrows, but I would consider this as solving the problem for a large portion of users. We could even go a little bit smaller, if we are concerned about increasing the line size too much.

@guimachiavelli @bidoubiwa @tpayet What do you think?

guimachiavelli:
I'm not a big fan of increasing the main content's max-width because long lines can be difficult to read, but I can live with it if it's a temporary thing.

An alternative solution would be to reduce the padding inside each tab element. Both regular tab and code sample components seem to share the style rules here, so it should be possible to change everything easily:

Screen Shot 2021-11-03 at 17 49 28

The above image shows what the quick start would look like if we reduced horizontal padding on `.el-tabs__item` by 5px (that is, from `padding: 0 20px` to `padding: 0 15px`). For reference, this is what it looks like now:

Screen Shot 2021-11-03 at 17 55 23

tpayet:
I do like gui's solution, but we should merge this PR and open another issue if we need to discuss it further

dichotommy (approved):

dichotommy:
bors merge

Issue #886, opened 2021-03-29
Add information about fieldDistribution


CaroFG:
There is currently no documentation on what fieldDistribution means when calling the stats routes. We don't know if it's dependent or not on the searchableAttributes : https://meilicommunity.slack.com/archives/CP9DVS1RQ/p1616765363060700

react-learner:
A more detailed explanation of fieldDistribution should be added in the [Stats API reference](https://docs.meilisearch.com/reference/api/stats.html).

We still need further clarification from the Core team (@MarinPostma @LegendreM ): is fieldDistribution affected in any way by searchableAttributes or displayedAttributes?

maryamsulemani97:
Fixed in PR #1250

PR #1250, opened 2021-11-04
explain fieldDistribution

maryamsulemani97:
closing #886

### DIFF
--- a/reference/api/stats.md
+++ b/reference/api/stats.md
@@ -34,6 +34,12 @@ Get stats of an index.
 }
 ```
 
+### Field distribution
+
+The `fieldDistribution` object is returned when you call the stats route. In the above code sample, it shows every field in the `movies` index along with the total number of documents in `movies` that contain that field.
+
+`fieldDistribution` is not impacted by `searchableAttributes` or `displayedAttributes`. Meaning, if one of the fields is not displayed or searchable, it will still be displayed in the `fieldDistribution` object.  
+
 ## Get stats of all indexes
 
 <RouteHighlighter method="GET" route="/stats"/>
@@ -78,3 +84,9 @@ Get stats of all indexes.
   }
 }
 ```
+
+### Field distribution
+
+The `fieldDistribution` object is returned when you call the stats route. In the above code sample, it shows every field in the `movies` and `rangemovies` indexes along with the total number of documents in the database that contain that field.
+
+`fieldDistribution` is not impacted by `searchableAttributes` or `displayedAttributes`. Meaning, if one of the fields is not displayed or searchable, it will still be displayed in the `fieldDistribution` object.  


CaroFG (commented):
I don't really like the fact that the content is duplicated, but I don't have a better idea where to put it...

CaroFG (dismissed):
LGTM!

CaroFG (dismissed):

reference/api/stats.md
```diff
@@ -1,6 +1,20 @@
 # Stats
 
-`Stats` gives extended information and metrics about indexes and the MeiliSearch database.
+`Stats` gives extended information and metrics about indexes and the MeiliSearch database. It returns the following details:
```
dichotommy, line 3:
```suggestion
The `/stats` route gives extended information and metrics about indexes and the MeiliSearch database. It returns the following details:
```

reference/api/stats.md
```diff
@@ -1,6 +1,20 @@
...
+
+#### 1. Number of documents
```
dichotommy, line 5: There's a MarkDown style rule about incrementing headers progressively that we try to follow. Basically we don't want to jump from an H1 to an H4 if we can avoid it. We also probably don't want to use a numbered list if the order doesn't matter.

I wonder what's the best way to present this information. A table could be good, since we could include a column for whether it's returned by "get stats of one index" or by "get stats of all indexes", or both. On the other hand it could look a bit messy.

Perhaps it would be better to emulate [Stripe's approach](https://stripe.com/docs/js/initializing#init_stripe_js-publishableKey) to method parameters. Something like this:

## Fields returned

### Get stats for one index

`GET /indexes/:index_uid/stats` returns the following fields:

#### `numberOfDocuments`

blah blah description blah

#### `isIndexing`

yadda yadda description here

#### `fieldDistribution`

description

### Get stats for all indexes

In addition to all fields returned by [`GET /indexes/:index_uid/stats`](#get-stats-for-one-index), `GET /stats` returns the following instance-level fields:

#### `databaseSize`

description

#### `lastUpdate`

description

What do you think @guimachiavelli ?

reference/api/stats.md
```diff
@@ -78,3 +94,5 @@ Get stats of all indexes.
   }
 }
 ```
+
+In the above code sample, `fieldDistribution` shows every field in the `movies` and `rangemovies` indexes along with the total number of documents in the database that contain that field.
```
dichotommy, line 98: > along with the total number of documents in the database that contain that field.

This is not accurate; if it was, the example would show 19654 + 19654 = 39308 for `overview`, `id`, and `title`.

reference/api/stats.md
```diff
@@ -34,6 +48,8 @@ Get stats of an index.
 }
 ```
 
+ In the above code sample, `fieldDistribution` shows every field in the `movies` index along with the total number of documents in `movies` that contain that field.
```
dichotommy, line 51: I don't think either of these (the two pieces of content that are almost identical) are necessary if we have a full explanation of the returned fields above. We can just link to that instead.

dichotommy (changes requested):
What you wrote on the displayed and searchable API references is true, but seems overly specific. Instead, let's clarify that these should only affect search results, nothing else.

For listing descriptions of all returned fields, I think the format inspired by Stripe's approach is a good one.

reference/api/stats.md
guimachiavelli: Sorry, completely missed this notification.

Being extremely nitpicky about this type of thing, I agree with @dichotommy: we shouldn't use ordered lists when the order of the list items is not important and headings should follow a logic progression from `#` to `##` to `###` and so on. Having a clear hierarchy that we always follow can be very useful down the road.

I'm also not a big fan of tables because they can don't degrade well when conditions are not ideal (too much content, too little content, different types and lengths of content in the same row, etc), so I like Stripe's approach. Keep in mind I worked for a long time as a front-end dev, so I'm naturally biased against tables, though.

Stripe's approach degrades nicely and I like it very much, but wouldn't using it here make the stats API reference inconsistent with all other endpoints?

maryamsulemani97: Agreed with this being inconsistent with all other endpoints but this is the only one where we need to explain what the returned fields mean.

guimachiavelli: I guess we can just move forward with @dichotommy's suggestion, then 👍

dichotommy: I think introducing a little inconsistency is acceptable if we take the decision together, and agree to use this as a model for future API reference revisions.

maryamsulemani97: Can we not use the field names as the heading?
I think Number of documents looks better as a heading compared to `numberOfDocuments`

image
image

I don’t think Is Indexing is a good alternative to `isIndexing`

dichotommy: I personally prefer the first screenshot, it feels like I have instant visual clarity that I'm looking at definitions of fields. What do you think @guimachiavelli ?

guimachiavelli: I'm team first screenshot too. I think I prefer it because I have an immediate visual cue that the each heading is describing a key/field, whereas "Number of documents" requires me to read the text to be sure.

reference/api/stats.md
```diff
@@ -1,12 +1,28 @@
 # Stats
 
-`Stats` gives extended information and metrics about indexes and the MeiliSearch database.
+The `/stats` route gives extended information and metrics about indexes and the MeiliSearch database.
 
-## Get stat of an index
+## Fields returned
+
+### Get stat of an index
```
dichotommy, line 7:
```suggestion
### Get stats of an index
```

reference/api/stats.md
```diff
@@ -34,15 +50,19 @@ Get stats of an index.
 }
 ```
 
-## Get stats of all indexes
+### Get stats of all indexes
 
 <RouteHighlighter method="GET" route="/stats"/>
 
-Get stats of all indexes.
+Get stats of all indexes. In addition to all fields returned by `GET /indexes/:index_uid/stats`, `GET /stats` returns the following instance-level fields:
```
dichotommy, line 57:
```suggestion
Get stats of all indexes. In addition to [all fields returned by `GET /indexes/:index_uid/stats`](#fields-returned), `GET /stats` returns the following instance-level fields:
```

Add link to all the fields returned by `GET /indexes/:index_uid/stats`.

dichotommy (changes requested):
I like the way you integrated the field descriptions into the API reference. However making the two endpoint sections children of "Fields returned" feels strange, and makes it so that you can't jump to a specific endpoint using the sidebar.

Screen Shot 2021-11-29 at 1 17 35 PM

Let's flip it. Get rid of `## Fields returned`. Put the two former H2s back as they were. Then, before both `#### numberOfDocuments` and `#### databaseSize`, add in `### Fields returned`. I think this will look really nice.

guimachiavelli (approved):
LGTM ✨

dichotommy:
Hey this looks beautiful! Seriously I think we'll use this as a model for future API reference pages.

Could we just move the "Path variables" table above the "Fields returned" header? That way it can be viewed at the same time as the route, without having to scroll back and forth.

dichotommy (approved):
Looks great to me 😎 🚀

maryamsulemani97:
bors merge

PR #1252, opened 2021-11-08
Revert "users should check the index status before they start searching"

dichotommy:
Reverts meilisearch/documentation#1173

We have an unknown error causing this new code sample to not be found. In addition, it should perhaps be standardized with the other getting started code samples (use `127.0.0.1` instead of `localhost`; wrapped in three upticks and marked as bash).

We also still need to request the integrations team to create code samples in other languages for this. So let's take a little longer to work on this PR.


maryamsulemani97 (approved):

dichotommy:
bors merge

PR #1254, opened 2021-11-08
Add state manager to store last active code language tab.

bidoubiwa:
This PR adds a state manager that upon clicking on a different label in the code-samples tabs, stores the language.

On language update, all other code-samples are updated accordingly. 

Meaning that if I click on `ruby`, the `ruby` tabs becomes not only the active one in the tabs I'm using, but in all others across the documentation. 

**Additionally** The store manager stores the language inside the local storage. 

Meaning that if you reload the page, the `ruby` will still be the one active

### DIFF
--- a/.vuepress/code-samples/components/codeSamples.vue
+++ b/.vuepress/code-samples/components/codeSamples.vue
@@ -32,17 +32,28 @@ export default {
   data() {
     return {
       samples: undefined,
+      preferedLanguage: undefined,
     }
   },
   computed: {
     tabsLanguage() {
-      return this.$store.state.tabsLanguage
+      this.resolvePreferedTabsLanguage()
+      return this.preferedLanguage
     },
   },
   created() {
     this.samples = CODE_SAMPLES[this.id]
   },
   methods: {
+    /*
+    * Check if the prefered language found in the store has an existing matching code-sample-id.
+    * If not, defaults to the first language it finds in the list (often bash).
+    */
+    resolvePreferedTabsLanguage: function () {
+      const languages = this.samples.map(sample => sample.language)
+      const storedTabsLanguage = this.$store.state.tabsLanguage
+      this.preferedLanguage = languages.includes(storedTabsLanguage) ? storedTabsLanguage : languages[0]
+    },
     /*
     * Updates the tabs language state with the current active tab language.
     * Storage of the state is done in `./store.js`
--- a/.vuepress/code-samples/components/codeSamples.vue
+++ b/.vuepress/code-samples/components/codeSamples.vue
@@ -1,10 +1,14 @@
 <template>
-  <div>
-    <tabs v-if="samples" type="border-card">
+  <!-- The click listener updates the prefered tabs language state in ./store.js -->
+  <div @click="updateLanguage">
+    <!-- the active-name prop defines the current active tab -->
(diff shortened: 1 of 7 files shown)


dichotommy:
We need to check that using `localStorage` doesn't violate any privacy concerns. Does anyone know about that, maybe @gmourier ?

gmourier:
It's not personally identifiable information and it doesn't go to an external server, and on top of that, it enhances the user experience, so I think it's very consistent without being invasive.

bidoubiwa:
Like @gmourier said! And about the [cookie law](https://softwareengineering.stackexchange.com/questions/290566/is-localstorage-under-the-cookie-law):

> The cookie law is not actually about cookies (and its not actually called the cookie law). Its about tracking users, storing and sharing the information with third parties. Cookies are just the most popular method to track users.
>
> If you don't want to show the "cookie warning" then just don't track the users beyond the session and don't share traffic data with third parties.

dichotommy:
@bidoubiwa Nice, it feels so good to select a language and see all the other tabs on the site change 😊 

One problem I noticed, if you use the left or right arrows on the code sample tabs, it will deselect a tab and show nothing.
Also, if you select a language like C# for which we don't have all code samples, some code samples will display nothing.

A simple solution would be to add all code sample names to the corresponding .code-samples.meilisearch.yaml file, with a brief message, e.g. "This code sample has not been added yet".

.vuepress/code-samples/components/codeSamples.vue
```diff
@@ -30,8 +34,26 @@ export default {
       samples: undefined,
     }
   },
+  computed: {
+    count() {
+      return this.$store.state.count
+    },
```
bidoubiwa, line 40: This should be removed (memo for me)

bidoubiwa:
> One problem I noticed, if you use the left or right arrows on the code sample tabs, it will deselect a tab and show nothing.
Also, if you select a language like C# for which we don't have all code samples, some code samples will display nothing.

This has been fixed by checking that when you click you are indeed clicking on a tab and not any other element.

> A simple solution would be to add all code sample names to the corresponding .code-samples.meilisearch.yaml file, with a brief message, e.g. "This code sample has not been added yet".

For the moment the solution is that if the code-sample is empty I will add the message `This code sample has not been added yet`. But, the problem remains if the code-sample does not exists.

The best solution would have been to complete missing `code-samples` of a given language using the code-sample-template. Since it required too many changes, I did not do it to avoid unforeseen errors that we might not see due to a lack of tests.

bidoubiwa:
After discussing a bit with @guimachiavelli, 

For code-samples-id that are missing in many sdk's, adding the tab with the following `This code sample has not been added yet`  would be maybe more of an annoyance as the user would have to click tab by tab to find one sdk that has the code-sample.

Thus, in the case, for example of csharp, where the SDK is missing a specifig code-sample, the active tab is changed to the first SDK containing the code-sample. 

Test: 
1. Click on a c# tab in the `getting started` 
2. Travel to `reference/features/web_interface.html` . Curl is now the active tab
3. Go back to getting started and c# is again the active tab

dichotommy (approved):
This is wonderful! Thank you so much @bidoubiwa 😊 💐 🚀 🕺🏻 🥇 🔥

dichotommy:
bors merge

Issue #1246, opened 2021-11-03
`displayedAttributes` takes priority over `attributesToRetrieve`


dichotommy:
It should be clarified [here](https://docs.meilisearch.com/reference/features/search_parameters.html#attributes-to-retrieve) that a field not present in `displayedAttributes` cannot be retrieved by search, even if added to the `attributesToRetrieve` parameter.

maryamsulemani97:
fixed in #1258

PR #1258, opened 2021-11-10
displayedAttributes take priority over attributesToRetrieve

maryamsulemani97:
closing #1246

### DIFF
--- a/reference/features/search_parameters.md
+++ b/reference/features/search_parameters.md
@@ -238,6 +238,10 @@ Configures which attributes will be retrieved in the returned documents.
 
 If no value is specified, `attributesToRetrieve` uses the [`displayedAttributes` list](/reference/features/settings.md#displayed-attributes), which by default contains all attributes found in the documents.
 
+::: note
+If a field is not added to `displayedAttributes`, it will not be retrieved on search, even if it is present in the `attributesToRetrieve` parameter.
+:::
+
 ### Example
 
 To get only the `overview` and `title` fields, set `attributesToRetrieve` to `["overview", "title"]`.
--- a/reference/features/search_parameters.md
+++ b/reference/features/search_parameters.md
@@ -239,7 +239,7 @@ Configures which attributes will be retrieved in the returned documents.
 If no value is specified, `attributesToRetrieve` uses the [`displayedAttributes` list](/reference/features/settings.md#displayed-attributes), which by default contains all attributes found in the documents.
 
 ::: note
-If a field is not added to `displayedAttributes`, it will not be retrieved on search, even if it is present in the `attributesToRetrieve` parameter.
+If an attribute has been removed from `displayedAttributes`, `attributesToRetrieve` will silently ignore it and the field will not appear in your returned documents.
 :::
 
 ### Example


reference/features/search_parameters.md
```diff
@@ -238,6 +238,10 @@ Configures which attributes will be retrieved in the returned documents.
 
 If no value is specified, `attributesToRetrieve` uses the [`displayedAttributes` list](/reference/features/settings.md#displayed-attributes), which by default contains all attributes found in the documents.
 
+::: note
+If a field is not added to `displayedAttributes`, it will not be retrieved on search, even if it is present in the `attributesToRetrieve` parameter.
```
guimachiavelli, line 242: There's something funny going on here that makes the whole sentence very complex for me. I think it's tied to us giving the context on the first clause ("If a field is not added etc") and then adding more context on the final clause ("even if it is present etc").

Is it just me? Is this simpler?

```suggestion
If an attribute has been removed from `displayedAttributes`, `attributesToRetrieve` will silently ignore it and the field will not appear in your returned documents.
```

maryamsulemani97: I like this better too!

dichotommy: Huh. I've been staring at these two sentences for more than five minutes. I think I prefer Maryam's original—it appeals to my robot brain—but I also see the potential for confusion that Gui pointed out. There's a real double negative problem.

Here's my minimalistic attempt, FWIW:

```
Attributes must be present in `displayedAttributes` to be retrievable with `attributesToRetrieve`.
```

reference/features/search_parameters.md
```diff
@@ -238,6 +238,10 @@ Configures which attributes will be retrieved in the returned documents.
 
 If no value is specified, `attributesToRetrieve` uses the [`displayedAttributes` list](/reference/features/settings.md#displayed-attributes), which by default contains all attributes found in the documents.
 
+::: note
+If an attribute has been removed from `displayedAttributes`, `attributesToRetrieve` will silently ignore it and the field will not appear in your returned documents.
```
dichotommy, line 242: I prefer this, but if you and Gui prefer the current text then let's go with that 😄 

```suggestion
Attributes must be present in `displayedAttributes` to be retrievable with `attributesToRetrieve`.
```

maryamsulemani97: I think the current one is better as it also tells the user what happens if you don’t have your attribute in `displayedAttributes` 👀

guimachiavelli: I'm fine with either. Since this is @maryamsulemani97's PR, I defer to her judgement.

dichotommy (approved):
I still have some reservations about the language we're using here.

> If an attribute has been removed from `displayedAttributes`, `attributesToRetrieve` will silently ignore it[...]

`displayedAttributes` by default contains `"*"` which is equivalent to all fields in the index. If you set `displayedAttributes` to some new value using the POST endpoint, the wildcard behavior is deactivated. That means that someone changing the value of `displayedAttributes` for the first time can remove attributes from the list without intending to.

Obviously this is questionable API design, but what does it have to do with us? Well it's about how we present this feature to readers. From a user perspective, there is no endpoint to remove attributes from `displayedAttributes`. A user reading this might say, "well I've never removed an attribute from `displayedAttributes`, so this warning doesn't apply to me." But if they've ever set `displayedAtttributes` to any value besides `"*"`, they'd be wrong.

I don't know if this would be solved by my suggestion (I suspect it would still be unclear; there's only so much we can do to compensate for the API design). Anyway, I recognize that this is a slightly ridiculous reason to delay merging this small PR any longer. I'm approving it but after voicing my dissent 👍🏻

maryamsulemani97:
This applies to all the Settings right? We could add a new warning on every page about this

maryamsulemani97:
bors merge

dichotommy:
@maryamsulemani97 

> This applies to all the Settings right?

AFAIK, it only applies to `displayedAttributes` and `searchableAttributes`.

PR #1259, opened 2021-11-10
Remove Mermaid

dichotommy:
This PR removes [Mermaid](https://github.com/mermaid-js/mermaid) from the project.

Currently, Mermaid is used for a single diagram in the asynchronous updates guide. We determined that this diagram could be actively harmful since it could give users the impression that MeiliSearch is much more complicated than it is, and decided to remove Mermaid since it's not being used elsewhere. Ideally we would like to replace this diagram with something more clear + brand consistent at a later date.

Did my best to remove the package completely, but I'm not great w/ dependencies—can someone confirm?


tpayet:
:+1:

#### About removing the dependency

You forgot to push the `yarn.lock` file which should be impacted after removing a dependency in the `package.json` file. Just run `$ yarn` after removing the dependency from the `package.json`, this will update the `yarn.lock` file and you can push those changes to the branch.

guimachiavelli (changes requested):
We just need to address what @tpayet about updating `yarn.lock` and we're good to go.

guimachiavelli (approved):
I always forget how messed up .lock files are. Remove one line from package.json and suddenly there are 700 fewer in the lock.

Looks good to me, though 👍

dichotommy:
bors merge

PR #1285, opened 2021-11-17
Clarify that settings are index-level

dichotommy:
Also avoid confusion with configuration, which is instance-level


maryamsulemani97 (approved):
🤖

dichotommy:
bors merge

PR #1288, opened 2021-11-17
v0.25: version bump

guimachiavelli:


.code-samples.meilisearch.yaml
```diff
@@ -463,9 +509,9 @@ phrase_search_1: |-
   curl -X POST 'http://localhost:7700/indexes/movies/search' \
   -H 'Content-Type: application/json' \
   --data-binary '{ "q": "\"african american\" horror" }'
-authentication_header_1: |-
+authorization_header_1: |-
   curl -X GET 'http://127.0.0.1:7700/keys' \
-    -H 'X-Meili-API-Key: masterKey'
+    -H 'Authorization: Bearer masterKey'
```
curquiza, line 514: Wouldn't you be consistent with the other way you use to authenticate with the master key?
Seen below:
```
-H 'Authorization: Bearer [your_master_key]'
```

.code-samples.meilisearch.yaml
```diff
@@ -77,16 +77,62 @@ search_post_1: |-
...
+    -H 'Content-Type: application/json' \
+    --data-binary '{
+      "description": "Manage documents: Products/Reviews API key", // Update key description to specify its new scope
+      "actions": [
+        "documents.add",
+        "documents.delete" // Add access to document deletion
+      ],
+      "indexes": [
+        "products",
+        "reviews" // Add access to the "reviews" index
```
curquiza, line 128: Are you aware that copying/pasting this curl example won't work since the comments are not accepted in JSON?
The users will have to delete the comment before sending the request.

.code-samples.meilisearch.yaml
```diff
@@ -77,16 +77,62 @@ search_post_1: |-
...
+      ],
+      "expiresAt": "2021-12-31T23:59:59Z" // Extend expiration date to the end of the year
+    }'
+delete_a_key_1: |-
+  curl \
+    -X DELETE 'http://localhost:7700/keys/d0552b41536279a0ad88bd595327b96f01176a60c2243e906c52ac02375f9bc4' \
+    -H 'Authorization: Bearer yourMasterKey'
```
curquiza, line 135: Same here, should we use `masterKey`, `yourMasterKey` or `[your_master_key]`?

bidoubiwa: for consistency I would go with `masterKey`

dichotommy: `masterKey` sounds good to me.

.code-samples.meilisearch.yaml
dichotommy: Thanks for noticing that @curquiza . I think we always want users to be able to copy/paste API reference code samples, so I'm removing the comments.

reference/features/security.md
```diff
@@ -0,0 +1,243 @@
...
+```sh
+./meilisearch
+```
+
+:::
+
+::: tab Environment variable
+
+If your master key was configured with an environment variable, unset `MEILI_MASTER_KEY` and relaunch the instance:
+
```
bidoubiwa, line 232: Same as above, missing windows env

bidoubiwa (dismissed):

bidoubiwa:
Small feedback on the naming of the `security` guide. I was searching everywhere where the guide was on API keys. Couln't find it so I went through the API references to find the link to the guide. 

It appears it is in `security`. A name like Permissions or Authorizations could have made it more clear that we have that feature. Security is more about your password needs to be at least 8 char long with different type of characters than showing of features.

.code-samples.meilisearch.yaml
```diff
@@ -577,4 +655,4 @@ document_guide_add_document_primary_key: |-
         "poster": "https://image.tmdb.org/t/p/w1280/xnopI5Xtky18MPhK40cZAGAOVeV.jpg",
         "overview": "A boy is given the ability to become an adult superhero in times of need with a single magic word.",
         "release_date": "2019-03-23"
-     }]' 
+     }]'
```
curquiza, line 658: Not a big issue, but useless addition here in this PR. We would rather avoid useless changes in PRs since they can "spoil" a future PR. This is for a better readability :)

guimachiavelli: Apologies, I'm having a hard time letting go of old habits. Most JS style guides I used recommended never leaving newlines at the end of files.

```suggestion
     }]' 

```

reference/features/security.md
```diff
@@ -0,0 +1,243 @@
...
+It is particularly important to protect an instance when dealing with sensitive data. You can use API keys to ensure only authorized people can search through an index containing medical records:
+
+<CodeSamples id="security_guide_search_key_1" />
+
+### Using default API keys for authorization
+
+When you launch an instance for the first time, MeiliSearch will automatically generate two API keys: `Default Search API Key` and `Default Admin API Key`.
+
+As its name indicates, the `Default Search API Key` can only be used to access [the search route](/reference/api/search.md).
```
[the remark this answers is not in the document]
guimachiavelli: I was indeed referring to how I called them in the preceding paragraph.

I get your point and `description` would perhaps be more technically correct, but by this point in the text we haven't mentioned API keys have fields. I think it's best to use a generic term like name here to avoid introducing too much stuff at the same time for new users (and experienced programmers will probably not even notice).

reference/features/security.md
```diff
@@ -0,0 +1,243 @@
...
+
+The second automatically-generated key, `Default Admin API Key`, has access to all API routes except `/keys`. You should avoid exposing the default admin key in publicly accessible code.
+
+::: warning
+Both default API keys have access to all indexes in an instance and do not have an expiry date. They can be [retrieved](#listing-api-keys), [modified](#updating-an-api-key), or [deleted](#deleting-an-api-key) the same way as user-created keys.
+:::
+
+### Differences between the master key and API keys
+
+MeiliSearch currently has two types of security keys: one master key and any number of API keys.
```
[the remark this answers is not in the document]
guimachiavelli: The opposition we were making here was of type **and** quantity. That is:

`one master key`

vs

`a variable amount of API keys`

guimachiavelli:
Thanks for the feedback, @bidoubiwa. We changed the name of this guide based on @gmourier's remark from a few months ago. In essence, if I understood it correctly, what MeiliSearch does is more accurately called Authorization.

We decided to go for a generic Security title on this PR to avoid getting caught in discussions of which term is 100% accurate. Do you think Security and Permissions would be clearer?

gmourier:
> Thanks for the feedback, @bidoubiwa. We changed the name of this guide based on @gmourier's remark from a few months ago. In essence, if I understood it correctly, what MeiliSearch does is more accurately called Authorization.
> 
> We decided to go for a generic Security title on this PR to avoid getting caught in discussions of which term is 100% accurate. Do you think Security and Permissions would be clearer?

We could use synonyms to allow the search bar to find the guide more easily. 🤓

E.g: Api Key, security, permissions, authentication, authorization, etc..

.code-samples.meilisearch.yaml
```diff
@@ -76,16 +76,62 @@ search_post_1: |-
...
+    -H 'Content-Type: application/json' \
+    --data-binary '{
+      "description": "Add documents: Products API key",
+      "actions": [
+        "documents.add"
+      ],
+      "indexes": [
+      "products"
+      ],
+      "expiresAt": "2021-11-13T00:00:00Z"
```
alallema, line 112: This date has passed so if the user tries this command it will return an error.

.code-samples.meilisearch.yaml
```diff
@@ -76,16 +76,62 @@ search_post_1: |-
...
+      "description": "Manage documents: Products/Reviews API key",
+      "actions": [
+        "documents.add",
+        "documents.delete"
+      ],
+      "indexes": [
+        "products",
+        "reviews"
+      ],
+      "expiresAt": "2021-12-31T23:59:59Z"
```
alallema, line 129: Same for this one.

.code-samples.meilisearch.yaml
guimachiavelli: Hmmm…that's true. The fact is that any date will eventually generate an error, unless we go for something completely unreasonable like 2121-11-13. What do you think, @dichotommy?

guimachiavelli: Is there a way of interpolating the date with vuepress? Like, instead of writing `2021-11-13` we write something like `{{currentYear}}-11-13`?

guimachiavelli:
Ok, so based on feedback from @bidoubiwa and @gmourier we're changing the title of the new guide from "Security" to "Security and permissions."

.code-samples.meilisearch.yaml
guimachiavelli: To answer my own question, vuepress allows doing something like that, but it doesn't work within code samples (likely because code samples are not parsed the same way as regular content).

dichotommy: > Is there a way of interpolating the date with vuepress?

Perhaps, but it strikes me as an overly complex solution. @bidoubiwa , thoughts since you developed the plugin?

I would suggest we change this to some date within the next hundred years, ideally one with some kind of significance or joke attached so it's more obvious that users shouldn't pay too much attention to it.

dichotommy, line 112:
```suggestion
      "expiresAt": "2042-04-02T00:42:42Z"
```

.code-samples.meilisearch.yaml
dichotommy:
```suggestion
      "expiresAt": "2042-04-02T00:42:42Z"
```

.code-samples.meilisearch.yaml
guimachiavelli: you're probably right. Let's go with that.

dichotommy (approved):
LGTM!

guimachiavelli:
bors merge

guimachiavelli:
thanks, @dichotommy 👍 
now, bors, be a good bot

bors merge

PR #1294, opened 2021-11-18
Remove `Under the hood`

guimachiavelli:


.vuepress/config.js
```diff
@@ -81,10 +81,18 @@ module.exports = {
...
             '/learn/advanced/asynchronous_updates',
             '/learn/advanced/snapshots_vs_dumps',
+            '/learn/advanced/bucket_sort',
+            '/learn/advanced/concat',
+            '/learn/advanced/datatypes',
+            '/learn/advanced/indexation',
+            '/learn/advanced/prefix',
+            '/learn/advanced/storage',
+            '/learn/advanced/tokenization',
+            '/learn/advanced/typotolerance',
```
dichotommy, line 91: What do you think of keeping these within a subsection in the sidebar, so they're not grouped together with all other advanced guides? My reasoning is that I would consider these less essential for most users and more advanced than the other ones in this category.

```suggestion
            {
              title: 'Under the Hood',
              collapsable: false,
              children: [
                '/learn/advanced/bucket_sort',
                '/learn/advanced/concat',
                '/learn/advanced/datatypes',
                '/learn/advanced/indexation',
                '/learn/advanced/prefix',
                '/learn/advanced/storage',
                '/learn/advanced/tokenization',
                '/learn/advanced/typotolerance'
              ]
            }
```

guimachiavelli: I hesitated putting everything under Advanced, but in the end I wasn't sure about what we gain from separating them?

Trying to think from the user's perspective, what benefit would they get from having `Advanced` and `Under the hood` categories/subsections? Is the distinction you indicate (essential advanced vs non-essential advanced articles) important to the user and, if so, is `Under the hood` the best way to communicate that?

Perhaps I'm just not a fan of the `Under the hood` heading because I'm not 100% sure what I will find under it?

dichotommy: I think it is a useful distinction for users. It's also a useful reminder for us, since most of the articles in Under the Hood have not been updated in a long time and could really use some love (I'm looking at you `datatypes.md`), and some are just shorter and lower quality than others on the site (e.g. `bucket_sort.md`, which is basically an entire page for an extremely simplified definition of bucket sort).

Obviously the ultimate goal is to make all of these articles as high quality as anything else on the site, but I think that's out of the scope of the current restructure—we will already have our hands full with Feature References—so for now we should focus on what we can do purely w/ content organization. For me, that means we should try to keep them "together" and separate from most articles. (And please let's just get rid of that bucket sort page. It's not adding anything. Links to it could just as easily go to wikipedia.)

Totally open to renaming the section. Ideas? I've got:

- MeiliSearch internals
- How stuff works
- Nuts & bolts

maryamsulemani97: I agree with keeping all of this under a separate subsection. It's advanced but users don't need to know all of this to use MeiliSearch (maybe minus the [space separators](https://docs.meilisearch.com/reference/under_the_hood/datatypes.html#string)).  
Inner workings sounds great!

guimachiavelli: `Inner workings` subsection it is, then!

learn/advanced/prefix.md
```diff
@@ -2,7 +2,7 @@
 
 In MeiliSearch, **you can perform a search with only a single letter as your query**. This is because we follow the philosophy of **prefix search**.
 
-Prefix search is when document sorting starts by comparing the search query against the beginning of each word in your dataset. All documents with words that match the query term are added to the [bucket sort](/reference/under_the_hood/bucket_sort.md), before the [ranking rules](/learn/core_concepts/relevancy.md#ranking-rules) are applied sequentially.
+Prefix search is when document sorting starts by comparing the search query against the beginning of each word in your dataset. All documents with words that match the query term are added to the [bucket sort](/learn/advanced/bucket_sort.md), before the [ranking rules](/learn/core_concepts/relevancy.md#ranking-rules) are applied sequentially.
```
maryamsulemani97, line 5: I don't see the bucket sort article under Advanced topics

guimachiavelli: Yeah, I have just removed it, but forgot to sort out dead links. As suggested by @dichotommy, I will just point those to a wikipedia article.

maryamsulemani97: I'd suggest going for something other than Wikipedia. Idk if its just me but I don’t think its always reliable 👀

dichotommy: If we're looking for an alternative to Wikipedia, I like this one: https://www.bbc.co.uk/bitesize/guides/z2m3b9q/revision/3

dichotommy: The nice thing about wikipedia is that it's basically guaranteed to always be up, and links don't change frequently. Even if the info isn't necessarily reliable, it's a very reliable place to link to and is generally seen as "neutral" (even if that's not the case).

I think for now it's okay to do a quick google search and choose what we identify as the best resource. However, at some point in the future we may want to formalize our approach to external links.

.vuepress/public/_redirects
```diff
@@ -81,11 +81,13 @@
...
 /reference/under_the_hood/indexation            /learn/advanced/indexation
 /reference/under_the_hood/prefix                /learn/advanced/prefix
 /reference/under_the_hood/storage               /learn/advanced/storage
 /reference/under_the_hood/tokenization          /learn/advanced/tokenization
-/reference/under_the_hood/typotolerance         /learn/advanced/typotolerance
\ No newline at end of file
+/reference/under_the_hood/typotolerance         /learn/advanced/typotolerance
+
+# Bucket sort
+/reference/under_the_hood/bucket_sort           /learn/core_concepts/relevancy#behavior
```
dichotommy: I think you may need to add `.html` for these

guimachiavelli: yeap, you're right 🎀

.vuepress/public/_redirects
```diff
@@ -80,5 +80,17 @@
...
+/reference/under_the_hood/concat.html                /learn/advanced/concat.html
+/reference/under_the_hood/datatypes.html             /learn/advanced/datatypes.html
+/reference/under_the_hood/indexation.html            /learn/advanced/indexation.html
+/reference/under_the_hood/prefix.html                /learn/advanced/prefix.html
+/reference/under_the_hood/storage.html               /learn/advanced/storage.html
+/reference/under_the_hood/tokenization.html          /learn/advanced/tokenization.html
+/reference/under_the_hood/typotolerance.html         /learn/advanced/typotolerance.html
+
+# Bucket sort
+/reference/under_the_hood/bucket_sort                /learn/core_concepts/relevancy.html#behavior
```
dichotommy, line 93: Missed one 😉 

```suggestion
/reference/under_the_hood/bucket_sort.html                /learn/core_concepts/relevancy.html#behavior
```

dichotommy: Thanks for rebasing btw ❤️

guimachiavelli: We had a couple of conflicts due to the latest PR moving the Update guide

guimachiavelli: Also, thanks for catching that missing `.html`

maryamsulemani97 (approved):
💣

guimachiavelli:
bors merge

PR #1298, opened 2021-11-22
v0.25: API keys new errors

### LINKED ISSUE #1283: v0.25: API keys: New errors
The new API keys introduce a few new errors specific to the `/keys` endpoint plus one error that is likely to be used elsewhere in the future: `missing_parameter`.

---

## Tasks
- [x] Update `errors.yaml`
  - [x] `invalid_api_key_actions`
  - [x] `invalid_api_key_indexes`
  - [x] `invalid_api_key_expires_at`
  - [x] `invalid_api_key_description`
  - [x] `invalid_api_key`
  - [x] `api_key_not_found`
  - [x] `missing_parameter`

---

## References

- spec PR: https://github.com/meilisearch/specifications/pull/85

maryamsulemani97:
closes #1283


gmourier (changes requested):
That's great! 

A take-it-or-leave-it idea, use `given` rather than `requested` in this case because we are sending a payload that is supposed to represent an API Key resource to the API. 🤓

maryamsulemani97:
> we are sending a payload that is supposed to represent an API Key resource to the API

@gmourier I don’t get what you mean here 👀

gmourier:
> > we are sending a payload that is supposed to represent an API Key resource to the API
> 
> @gmourier I don’t get what you mean here 👀

It's just that for me the term request may not be clear between "making a request to retrieve information from an API" and "the purely technical fact to describe an HTTP communication".

To clarify here that we don't ask the API for information but we send it information but it is still a request.

It may not make much sense and may not matter much. Feel free to ignore it 😀  Request is probably very understandable for most people.

guimachiavelli (changes requested):
I do agree with @gmourier's reasoning: while we are **requesting** the server to do something with an API Key, I think it is correct to emphasise the error is in the values given in our request.

gmourier (approved):
LGTM 📦

errors.yaml
```diff
@@ -66,4 +66,16 @@ errors:
     description: "The requested database has reached its maximum size; no further documents can be added. For information on customizing the maximum database size accepted by MeiliSearch, check out this [guide](https://docs.meilisearch.com/reference/features/configuration.html#max-udb-size)."
   - code: index_primary_key_already_exists
     description: "The requested index already has a primary key that cannot be changed."
+  - code: invalid_api_key_description
+    description: "The `description` field for the provided API key resource is invalid. It should either be a string or be set to`null`."
```
guimachiavelli, line 70:
```suggestion
    description: "The `description` field for the provided API key resource is invalid. It should either be a string or `null`." 
```

guimachiavelli (changes requested):

guimachiavelli (approved):
LGTM 🐚

guimachiavelli:
bors merge

Issue #1249, opened 2021-11-04
Get rid of code samples arrow


dichotommy:
With #1245 , we officially passed 11 SDKs on the documentation. This has pushed the last tab of the code-samples plugin past the edge of our `max-width`, hiding it and requiring users to notice and click a small arrow to find it. This risks provoking erroneous issues from users who believe the Dart SDK (for example) is missing.

We will most likely go forward with [the solution](https://github.com/meilisearch/documentation/pull/1245#issuecomment-959729036) proposed by @guimachiavelli : decreasing the horizontal padding on [tabs](https://superbiger.github.io/vuepress-plugin-tabs/) to create enough space for the current number of coding languages.

If additional languages are added in the future before our [migration to a new platform](https://github.com/meilisearch/documentation/issues/932), we may need to consider another solution, such as increasing the main content's `max-width`.

dichotommy:
This problem was (temporarily) fixed by https://github.com/meilisearch/documentation/pull/1301. Closing this issue until adding another SDK to the docs forces us to reopen it.

PR #1301, opened 2021-11-22
Remove code samples arrow

guimachiavelli:
As described on #1249, the docs now have code samples 11 SDKs, which forces janky horizontal scrolling via easy-to-miss buttons.

This PR removes some horizontal padding to avoid horizontal scrolling.

### DIFF
--- a/.vuepress/styles/index.styl
+++ b/.vuepress/styles/index.styl
@@ -27,3 +27,7 @@ body
 
 {$contentClass} a code
   color $accentColor
+
+.el-tabs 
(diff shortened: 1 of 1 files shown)


guimachiavelli:
There's a small issue with unequal horizontal padding on the last tab item due to a weird rule forcing it to be 20px. It's a bit difficult to change that because we apparently cannot change styling directly, only override it by using higher-specificity selectors.

dichotommy:
bors merge

PR #1310, opened 2021-11-26
standardize authentication_header code sample

dichotommy:
For some reason this code sample uses double quotes in its header, where every other code sample on the site use single quotes, e.g. for the `Content-Type` field. It's also the only code sample on the site to use `--header` instead of `-H`. 

Of note, it is the only code sample on our site to demonstrate communicating w/ a protected instance using `X-Meili-API-Key`, so there may be a good reason for this distinction. I can't imagine what it would be though, so it's probably better to be consistent 👀


guimachiavelli (approved):
LGTM 📊

dichotommy:
bors merge

PR #1312, opened 2021-11-27
v0.25: Keys API reference

### LINKED ISSUE #1286: v0.25: API keys: API reference
With v0.25, the `/keys/` endpoint is being expanded to include a secondary `/keys/:key` route. 

The `/keys` endpoint can only be accessed when using the `main` key.

- [ ] Update `reference/api/keys`:
  - [ ] rewrite `GET /keys` route: lists all currently active API keys (`main` key is not included)
  - [ ] add `GET /keys/:key` route: get information on a specific API key
  - [ ] add `POST /keys` route: create a new API key
  - [ ] add `PATCH /keys/:key` route: update an API key (e.g. change its description or its expiry date)
  - [ ] add `DELETE /keys/:key` route: delete a key

**Updated 2021-11-26:** Previously this issue body stated that `POST /keys/:key` could be used to create a new key. The correct route is `POST /keys`. This has been corrected.

---

## Reference
- spec PR: https://github.com/meilisearch/specifications/pull/85

dichotommy:
Closes #1286 

## To note while reviewing

In our current format for API reference pages, information on path variables and body fields is listed in a table for each endpoint ([see example](https://docs.meilisearch.com/reference/api/indexes.html#update-an-index)). This is very convenient for users, but inefficient for us to update and prone to human error.

Since this PR essentially involves creating an API reference page from scratch, I used it as an opportunity to try a different approach. Instead of following the aforementioned convention, I chose to consolidate field descriptions in a single section (here called "API key object definition"). This has both advantages and disadvantages, which I hope to discuss below.

Things I'm wondering: 
- In an approach like this, where on the page should this section be located?
- Is it still necessary to include the path variables table?
- Is the convenience cost to users too great? (i.e. they have to click away from the specific endpoint they're looking at to go to the object definition, then find their way back).
- Should `actions` table be simplified?

**This is an experiment** to fuel discussion of our approach. If it doesn't work, I am happy to rewrite this page in the established format.


reference/api/keys.md
```diff
@@ -1,30 +1,264 @@
...
+
+| name             | description |
+|------------------|-------------|
+| search           | Provides access to `GET` and `POST` methods on `/indexes/:authorizedIndexes/search` routes. |
+| documents.add    | Provides access to `POST` and `PUT` on `/indexes/:authorizedIndexes/documents` routes. |
+| documents.get    | Provides access to `GET` methods on `/indexes/:authorizedIndexes/documents` and `/indexes/:authorizedIndexes/documents/:documentId` routes. |
+| documents.delete | Provides access to `DELETE` method on `/indexes/:authorizedIndexes/documents/:documentId`, `indexes/:authorizedIndexes/documents/:documentId` and `POST` method on `/indexes/:authorizedIndexes/documents/delete-batch` routes. |
+| indexes.add      | Provides access to `POST` `/indexes`. **⚠️ The newly created index is automatically added to the permissions for the API key making the call, as well as all others keys having `[*]` as a value for `indexes`**. |
+| indexes.get      | Provides access to `GET` `/indexes` and `/indexes/:authorizedIndexes`. **⚠️Non-authorized `indexes` are omitted from the response on `/indexes`**. |
```
guimachiavelli, line 57: Do we really need the `⚠️  etc` bits in the description? They feel a bit more like warnings for the devs than for documentation users.

.code-samples.meilisearch.yaml
```diff
@@ -83,10 +83,48 @@ get_update_1: |-
...
+get_one_key_1: |-
   curl \
-    -H "X-Meili-API-Key: 123" \
-    -X GET 'http://localhost:7700/keys'
+    -X GET 'http://localhost:7700/keys/d0552b41536279a0ad88bd595327b96f01176a60c2243e906c52ac02375f9bc4' \
+    -H 'Authorization: Bearer yourMasterKey'
+get_all_keys_1: |-
+  curl \
+    -X GET 'http://localhost:7700/keys' \
+    -H 'Authorization: Bearer yourMasterKey'
```
guimachiavelli, line 93: Possibly nitpicking, but I find this authorization key a bit distracting. I assume this is a reference to something, what with the blank space between `Bearer` and `yourmasterkey`, but I'm not sure what.

guimachiavelli: Ah, no, wait, I realised this is following the spec.

Perhaps we can add some `[]` to the master key value?

reference/api/keys.md
```diff
@@ -1,30 +1,264 @@
...
+#### Response: `201 Created`
+
+```json
+{
+    "description": "Add documents: Products API key",
+    "key": "d0552b41536279a0ad88bd595327b96f01176a60c2243e906c52ac02375f9bc4",
+    "actions": [
+        "documents.add"
+    ],
+    "indexes": ["products"],
```
maryamsulemani97, line 211:
```suggestion
    "indexes": [
    "products"
    ],
```

.code-samples.meilisearch.yaml
```diff
@@ -83,10 +83,48 @@ get_update_1: |-
...
+  curl \
+    -X POST 'http://localhost:7700/keys' \
+    -H 'Authorization: Bearer yourMasterKey' \
+    -H 'Content-Type: application/json' \
+    --data-binary '{
+      "description": "Add documents: Products API key",
+      "actions": [
+        "documents.add"
+      ],
+      "indexes": ["products"],
```
maryamsulemani97, line 104:
```suggestion
      "indexes": [
      "products"
      ],
```

maryamsulemani97 (changes requested):

reference/api/keys.md
```diff
@@ -1,30 +1,264 @@
...
+**Type:** string
+**Modifiable:** no
+
+Date and time the key was last updated, represented in **ISO-8601** format.
+
+## Get one key
+
+<RouteHighlighter method="GET" route="/keys/:key"/>
+
+Get information on the specified key.
```
guimachiavelli, line 102: Should we say something about what happens when you use `GET keys/:key` with a deleted/expired key?

reference/api/keys.md
```diff
@@ -1,30 +1,264 @@
...
+To learn more about the fields displayed here, take a look at the [API key object definition](#api-key-object-definition).
+
+## Update a key
+
+<RouteHighlighter method="PATCH" route="/keys/:key"/>
 
-::: warning
-You must have the master key to access this route.
-:::
+Update the description, permissions, and/or expiration date of an API key.
```
guimachiavelli, line 224: I personally feel "/or" is not necessary here. Also feels like we don't use that construct very often (or at all?) in the rest of the docs, though we can always decide to change that.

```suggestion
Update the description, permissions, and expiration date of an API key.
```

reference/api/keys.md
```diff
@@ -1,30 +1,264 @@
...
+
+You do not have to reproduce the API key object in the payload. **Provide only the fields you intend to update.**
```
guimachiavelli, line 226: I find the first sentence a bit confusing. Perhaps we can drop it?

Not sure this is any better:

```suggestion
This is a `PATCH` route, which means you can provide only the fields you intend to update.
```

guimachiavelli (changes requested):

dichotommy:
Really struggling with the decision of whether to keep the API key object definition separate or follow the current format (use tables for path variables, body variables, etc.).

I think the separate definition of fields is significantly prettier and less crowded than the tables, and obviously easier for us to update (SSOT). However, it requires users viewing an endpoint to click on an additional link. Additionally, if the `## API key object definition` header appears in the sidebar, that breaks the unspoken rule that all sidebar links in the API reference represent endpoints.

reference/api/keys.md
```diff
@@ -1,30 +1,264 @@
...
+| dumps.get        | Provides access to `GET` `/dumps/:dumpUid` route. **As dumps are not scoped by indexes, a restriction on `indexes` does not affect this action.** |
+| version          | Provides access to `GET` `/version` route.
+
+#### `indexes`
+
+**Type:** array
+**Modifiable:** yes
+**Mandatory:** yes
+
+A list of indexes permitted for the key. `["*"]` for all indexes.
```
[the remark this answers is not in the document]
guimachiavelli: gods, it took me five minutes to decipher what I meant.

In essence, I was wondering if using `for` with `permit` is correct English. I get that the meaning is "a list of the indexes a key has permission to access", but "permitted for the key" sounds a bit clunky. Or is it just me?

dichotommy:
As I said before, I have really struggled with some of the decision making on this one. I chose to integrate the API key object definition into the endpoints themselves, to avoid breaking our consistent rule of `API reference H2 == endpoint`. I also wanted to avoid copy and pasting this object definition 4-5 different times (once for each endpoint), so instead I copied it twice and slightly modified the content each time: once for get all keys, and once for create a key.

Let me know what you think about this. Please be honest! I really am not satisfied with this, but I'm not getting anywhere working on it on my own so I need an outside perspective.

maryamsulemani97:
Can we start with the Get all keys endpoint explain all the fields there and refer to it for other endpoints?

.code-samples.meilisearch.yaml
guimachiavelli:
```suggestion
    -H 'Authorization: Bearer [your_master_key]'
```

.code-samples.meilisearch.yaml
```diff
@@ -83,10 +83,50 @@ get_update_1: |-
...
+    -X GET 'http://localhost:7700/keys/d0552b41536279a0ad88bd595327b96f01176a60c2243e906c52ac02375f9bc4' \
+    -H 'Authorization: Bearer yourMasterKey'
+get_all_keys_1: |-
+  curl \
+    -X GET 'http://localhost:7700/keys' \
+    -H 'Authorization: Bearer yourMasterKey'
+create_a_key_1: |-
+  curl \
+    -X POST 'http://localhost:7700/keys' \
+    -H 'Authorization: Bearer yourMasterKey' \
```
guimachiavelli, line 97:
```suggestion
    -H 'Authorization: Bearer [your_master_key]'  \
```

.code-samples.meilisearch.yaml
```diff
@@ -83,10 +83,50 @@ get_update_1: |-
...
+      ],
+      "indexes": [
+      "products"
+      ],
+      "expiresAt": "2021-11-13T00:00:00Z"
+    }'
+update_a_key_1: |-
+  curl \
+    -X PATCH 'http://localhost:7700/keys/d0552b41536279a0ad88bd595327b96f01176a60c2243e906c52ac02375f9bc4' \
+    -H 'Authorization: Bearer yourMasterKey' \
```
guimachiavelli, line 112:
```suggestion
    -H 'Authorization: Bearer [your_master_key]'  \
```

.code-samples.meilisearch.yaml
```diff
@@ -83,10 +83,50 @@ get_update_1: |-
...
+      "indexes": [
+        "products",
+        "reviews" // Add access to the "reviews" index
+      ],
+      "expiresAt": "2021-12-31T23:59:59Z" // Extend expiration date to the end of the year
+    }'
+delete_a_key_1: |-
+  curl \
+    -X DELETE 'http://localhost:7700/keys/d0552b41536279a0ad88bd595327b96f01176a60c2243e906c52ac02375f9bc4' \
+    -H 'Authorization: Bearer yourMasterKey'
```
guimachiavelli, line 129:
```suggestion
    -H 'Authorization: Bearer [your_master_key]' 
```

reference/api/keys.md
```diff
@@ -1,30 +1,313 @@
 # Keys
 
-Each instance of MeiliSearch has three keys: a master, a private, and a public. Each key has a given set of permissions on the API routes.
+The `keys` route allows you to create, manage, and delete API keys. In order to use these endpoints, the [master key must be set](/reference/features/configuration.md#master-key) and supplied in the header of the request:
 
-You must have the master key to access the `keys` route.
+```bash
+curl \
+  -H 'Authorization: Bearer masterKeyGoesHere'
```
guimachiavelli, line 7:
```suggestion
  -H 'Authorization: Bearer [your_master_key]'
```

reference/api/keys.md
```diff
@@ -1,30 +1,313 @@
...
+**Modifiable:** yes
+
+A list of indexes permitted for the key. `["*"]` for all indexes.
+
+#### `expiresAt`
+
+**Type:** string
+**Modifiable:** yes
+
+Date and time when the key will expire, represented in **ISO-8601** format. `null` if the key never expires.
```
guimachiavelli, line 108: @maryamsulemani97 has been writing `ISO 8601` (without the dash).  I have my doubts about the usefulness of mentioning ISO formats by their code, but I think we should make sure we're consistent in how we call it.

This comment also applies to the two fields below, `createdAt` and `updatedAt`.

reference/api/keys.md
```diff
@@ -1,30 +1,313 @@
...
+}
+```
+
+## Update a key
+
+<RouteHighlighter method="PATCH" route="/keys/:key"/>
+
+Update the description, permissions, or expiration date of an API key. To learn more about the variables sent in the body of the request, see the [create key](#body) endpoint.
+
+Updates to keys are **partial**. This means that you should provide only the fields you intend to update, as any fields not present in the payload will remain unchanged.
```
guimachiavelli, line 271: Major nitpick: that `that` seems unnecessary. 

(Major as in extremely small and unimportant).

```suggestion
Updates to keys are **partial**. This means you should provide only the fields you intend to update, as any fields not present in the payload will remain unchanged.
```

guimachiavelli (changes requested):
Making some of the examples consistent with the guide code samples, plus minor comments here and there.

reference/api/keys.md
maryamsulemani97: I used ISO 8601 instead of `ISO 8601`.

guimachiavelli: Yeap, sorry, I accidentally used backticks instead of quotes for some reason.

Also, [it seems ISO also prefer not to use the hyphen](https://www.iso.org/iso-8601-date-and-time-format.html). I think it makes sense to follow their usage and format references to ISO standards as @maryamsulemani97 has been doing.

```suggestion
Date and time when the key will expire, represented in **ISO 8601** format. `null` if the key never expires.
```

dichotommy:
> Can we start with the Get all keys endpoint explain all the fields there and refer to it for other endpoints?

I moved the get all keys endpoint first and explained all the fields there, but still kept the body section in the create a key endpoint for a detailed explanation of the modifiable fields.

reference/api/keys.md
```diff
@@ -1,30 +1,277 @@
...
+
+The `*` character can also be used as a wildcard to get all permissions for a route, e.g. `documents.*` to authorize access to all document-related endpoints.
+
+#### `indexes`
+
+An array of indexes permitted for the key. `["*"]` for all indexes.
+
+#### `expiresAt`
+
+Date and time when the key will expire, represented in **ISO 8601** format. `null` if the key never expires.
```
maryamsulemani97, line 97: Do we want ISO 8601 to be bold every time?

guimachiavelli: Good catch. I'm against it, feels like we're giving it too much importance.

PR #1319, opened 2021-12-01
Small correction updating guide

dichotommy:
Just a small correction that probably no one will ever notice 😅


dichotommy:
Now that I'm taking a second look at it, not sure this is necessary or accurate. Don't dumps currently use the get documents endpoint? The only thing that changed in v0.21 is that the get documents endpoint no longer uses `displayedAttributes`, if I remember correctly.

Closing this.

PR #1322, opened 2021-12-02
Specify node version in CI jobs

bidoubiwa:
# Pull Request

## What does this PR do?

Fixes [CI environment that updated to node 16](https://github.com/actions/virtual-environments/issues/4446) while the documentation is not yet compatible with node 16.

## PR checklist
Please check if your PR fulfills the following requirements:
- [x] Have you read the contributing guidelines?
- [x] Have you made sure that the title is accurate and descriptive of the changes?

Thank you so much for contributing to MeiliSearch!

PR #1324, opened 2021-12-06
test our CIs

dichotommy:
We have a problem w/ CIs failing due to [GitHub changing the default version of Node for their VMs to v16](https://github.com/actions/virtual-environments/issues/4446), which is incompatible with our VuePress site. We attempted to resolve it with #1322 . This is a test to try to understand why we still see failing CIs.


dichotommy:
Test successful! We learned that GitHub CIs use the `.github/*.yaml` files present in the PR branch, not the ones present in master.

PR #1327, opened 2021-12-07
v0.25: Drop support for dump V1

### LINKED ISSUE #1326: v0.25: Drop support for dump V1
MeiliSearch > 0.25 will drop support for dumps version 1. This concerns MeiliSearch with versions before v0.21 (v0.21 not included).

# Tasks
- [x] update `/reference/features/dumps.md`
- [x] update `/learn/advanced/updating.md`

# Reference
https://github.com/meilisearch/MeiliSearch/issues/1972

maryamsulemani97:
closes #1326

### DIFF
--- a/learn/advanced/snapshots_vs_dumps.md
+++ b/learn/advanced/snapshots_vs_dumps.md
@@ -18,7 +18,8 @@ As a result, importing a dump requires MeiliSearch to re-index all of your docum
 
 ::: note
 We do not recommend using dumps from a new MeiliSearch version to import an older version.
-For example, you **should not import a dump from MeiliSearch v0.22.0 to MeiliSearch v0.21.0.** But importing a dump from MeiliSearch v0.21.0 to MeiliSearch v0.21.0 or higher will work.
+
+For example, you can import a dump from MeiliSearch v0.21 into v0.22 without any problems. Importing a dump generated in v0.22 into a v0.21 instance, however, can lead to unexpected behavior.
 :::
 
 ### Conclusion
--- a/learn/advanced/snapshots_vs_dumps.md
+++ b/learn/advanced/snapshots_vs_dumps.md
@@ -18,7 +18,7 @@ As a result, importing a dump requires MeiliSearch to re-index all of your docum
 
 ::: note
 We do not recommend using dumps from a new MeiliSearch version to import an older version.
-For example, you **should not** import a dump from MeiliSearch v0.22.0 to MeiliSearch v0.21.0. But importing a dump from MeiliSearch v0.21.0 to MeiliSearch v0.21.0 or higher will work.
+For example, you **should not import a dump from MeiliSearch v0.22.0 to MeiliSearch v0.21.0.** But importing a dump from MeiliSearch v0.21.0 to MeiliSearch v0.21.0 or higher will work.
 :::
 
 ### Conclusion
--- a/learn/advanced/updating.md
+++ b/learn/advanced/updating.md
@@ -254,7 +254,7 @@ Now that you’ve got your dump, [install the latest version of MeiliSearch](/le
 ```
 
 ::: warning
-If you are using MeiliSearch v0.20.0 or below, you will need to import your dump to a version between v0.21.0 and v0.25.0 and the re-export it to the newer version.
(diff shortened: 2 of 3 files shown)


learn/advanced/snapshots_vs_dumps.md
```diff
@@ -18,7 +18,7 @@ As a result, importing a dump requires MeiliSearch to re-index all of your docum
 
 ::: note
 We do not recommend using dumps from a new MeiliSearch version to import an older version.
-For example, you **should not** import a dump from MeiliSearch v0.22.0 to MeiliSearch v0.21.0. But importing a dump from MeiliSearch v0.21.0 to MeiliSearch v0.21.0 or higher will work.
+For example, you **should not import a dump from MeiliSearch v0.22.0 to MeiliSearch v0.21.0.** But importing a dump from MeiliSearch v0.21.0 to MeiliSearch v0.21.0 or higher will work.
```
guimachiavelli, line 21: Does it work if we invert the order of these examples? Meaning, we start from what is correct and then say what is wrong?

```suggestion
For example, you can import a dump from MeiliSearch v0.21 into v0.22 without any problems. Importing a dump generated in v0.22 into a v0.21 instance, however, can lead to unexpected behavior.
```

I would also simplify versioning here: `v0.21` is easier to parse than `v0.21.0` and gets the point across.

learn/advanced/updating.md
```diff
@@ -253,6 +253,10 @@ Now that you’ve got your dump, [install the latest version of MeiliSearch](/le
 ./meilisearch --import-dump /dumps/your_dump_file.dump
 ```
 
+::: warning
+If you are using MeiliSearch v0.20.0 or below, you will need to import your dump to a version between v0.21.0 and v0.25.0 and the re-export it to the newer version.
```
guimachiavelli, line 257: What do you think about this? Too wordy?

I think the important thing to keep in mind here is when a dump must be exported and when it must be imported.

```suggestion
If you are using MeiliSearch v0.20 or below, migration should be done in two steps. First, import your v0.20 dump into an instance running any version of MeiliSearch between v0.21 and v0.25. Second, export another dump from this instance and import it to a final instance running your targeted version.
```

reference/features/dumps.md
```diff
@@ -51,6 +51,15 @@ If you restart MeiliSearch after creating a dump, you will not be able to use th
 
 ## Importing a dump
 
+Dumps in v0.20.0 and below are no longer compatible with the new versions. Before you start importing, check your [MeiliSearch version](/reference/api/version.md#example) and proceed accordingly.
+
+::: note
+We do not recommend using dumps from a new MeiliSearch version to import an older version.
+For example, you **should not import a dump from MeiliSearch v0.22.0 to MeiliSearch v0.21.0.** But importing a dump from MeiliSearch v0.21.0 to MeiliSearch v0.21.0 or higher will work.
```
guimachiavelli, line 58: If you accept my suggestion on `learn/advanced/snapshots_vs_dumps.md`, this should be updated accordingly.

Also, be careful with that single line break between 57 and 58: these sentences either belong to the same paragraph and should be separated by a regular space or they are different paragraphs and should be separated by a blank line.

reference/features/dumps.md
```diff
@@ -59,10 +68,9 @@ As the data contained in the dump needs to be indexed, the process will take som
 ./meilisearch --import-dump /dumps/20200813-042312213.dump
 ```
 
-::: note
-We do not recommend using dumps from a new MeiliSearch version to import an older version.
-For example, you **should not** import a dump from MeiliSearch v0.22.0 to MeiliSearch v0.21.0. But importing a dump from MeiliSearch v0.21.0 to MeiliSearch v0.21.0 or higher will work.
-:::
+### Importing a dump for v0.20.0 or below
+
+If your MeiliSearch version is v0.20.0 or below, you will need to import your dump to a version between v0.21.0 and v0.25.0 and then re-export it to the newer version.
```
guimachiavelli, line 73: Same as my comment on `learn/advanced/updating.md`.

guimachiavelli (changes requested):

guimachiavelli (approved):
💣

guimachiavelli:
bors merge

guimachiavelli:
I believe in you, bors
bors merge

PR #1280, opened 2021-11-15
new getting started

maryamsulemani97:
closing #1279

PR #1343, opened 2021-12-15
Copy the complete installation panel in quickstart

ManyTheFish:
Many users don't create an index after launching a Meilisearch instance, moreover, we have many users that use Docker in development environment.

This low rate of index creation **could be** explained by a bad setup of the docker running. We may add the exhaustive installation panel in the quickstart guide to see if it increases the conversion rate.

poke @gmourier.

### DIFF
--- a/learn/getting_started/quick_start.md
+++ b/learn/getting_started/quick_start.md
@@ -8,10 +8,116 @@ All that is required is a [command line](https://www.learnenough.com/command-lin
 
 First of all, let's download and run MeiliSearch.
 
+:::: tabs
+
+::: tab cURL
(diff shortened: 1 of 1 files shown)


dichotommy:
We are already planning to remove the [installation](https://docs.meilisearch.com/learn/getting_started/installation.html) page when we add the new quick start + getting started (#1280 ). This PR would already make that page pretty redundant, so should we just do it now? @guimachiavelli @maryamsulemani97

ManyTheFish:
Oh! Thanks @dichotommy, I didn't see the PR about quickstart rewriting, I close mine.

I think we would be interested in the release date of this PR in order to watch if there is a change in the conversion rate. 👍

PR #1348, opened 2021-12-21
move snapshot and dump guides

guimachiavelli:
As part of our current efforts to simplify the reference page, I am slowly finding new places for content under reference/features.

This PR moves the guides for snapshots and dumps to /learn/advanced without any changes to these pages. The two guides are grouped under a "Data backup" subheading, together with "Data backup: dumps vs snapshots", which has itself been named to "Dumps vs snapshots".

As usual, this PR also sets up the required redirects and fixes broken links caused by moving files around.


.vuepress/config.js
```diff
@@ -79,11 +79,29 @@ module.exports = {
...
+              sidebarDepth: 0,
+              children: [
+                {
+                  title: 'Dumps',
+                  path: '/learn/advanced/dumps',
+                },
+                {
+                  title: 'Snapshots',
+                  path: '/learn/advanced/snapshots',
+                },
```
dichotommy, line 98: Unless I'm mistaken you've also got to remove these two from the `Feature references` sidebar.

reference/features/README.md
```diff
@@ -9,14 +9,14 @@ Understanding our reference documentation is easier with background knowledge of
 [Authentication](/reference/features/authentication.md)
 [Configuration](/reference/features/configuration.md)
 [Distinct attribute](/reference/features/distinct.md)
-[Dumps](/reference/features/dumps.md)
+[Dumps](/learn/advanced/dumps.md)
```
dichotommy, line 12: We should remove these links instead of updating them, since these pages are no longer in this site section.
(I know it's effectively pointless since these READMEs will be removed soon. Humor me 🥲 )

dichotommy (changes requested):
Looks pretty good. Let's remove the dump and snapshot guides from the `/reference/features` sidebar and see if that fixes the Netlify deploy preview. If it does then I think this PR is good to go 👍🏻

reference/features/README.md
guimachiavelli, line 12:
```suggestion
```

dichotommy (approved):
LGTM. A few notes/questions:

- Why set `sidebarDepth` to 0 for the new section? Did you test it and find that it looked strange (too nested)?
- Perhaps `Snapshots vs. Dumps` should come before the individual guides. Since it functions as a high-level overview of the differences between the two, I can see users consulting this first to decide which form of backup they want to go with and then looking at the corresponding guide for implementation.

dichotommy:
(Note that we will not merge this until the [v0.25 PR](http://github.com/meilisearch/documentation/pull/1288) has been merged, to avoid messy conflicts)

guimachiavelli:
> Why set sidebarDepth to 0 for the new section? Did you test it and find that it looked strange (too nested)?

Honestly, I copied that from another menu section and didn't give it much thought. Seems to make no difference in our current structure, so it's probably ok to just remove it.

---

> Perhaps Snapshots vs. Dumps should come before the snapshot and dump guides. Since it functions as a high-level overview of the differences between the two, I can see users consulting this first to decide which form of backup they want to go with and then looking at the corresponding guide for implementation.

I don't have a strong position on this. I think it's also possible to argue you must first know what each of these are before a comparison is useful, so to avoid the discussion I went with alphabetical order.

This does point towards a more widespread issue, imho: we have no rules for organising the sidebar menu. Some subsections seem to follow fairly arbitrary logic, others follow alphabetical order, and others…well, embrace chaos. I will open a discussion for that.

dichotommy (approved):
LGTM! 🎉

dichotommy:
bors merge

PR #1357, opened 2022-01-03
sample api reference template

maryamsulemani97:


dichotommy (commented):
- I think there is value in the HTTP verb endpoint templates you created (GET, POST, DELETE etc), but in order for it to become a functional resource we can use every time we’re creating a new API reference page, I think ideally they should be combined into a single page/file.
- Also, better if the template is not referring to an existing route (e.g. indexes) and is instead using something made up or comical, e.g. GET /foobar. This would make it easier to tell that they are templates and to for example use cmd+f to quickly find and replace all the template text (e.g. find and replace all foobar with keys, voila we have the skeleton of a keys API reference).
- In the template file, we may want to include multiple examples of the same HTTP verb such as “Get one” + “Get all”, or “Delete one” + “Delete all”

maryamsulemani97:
Close once https://github.com/meilisearch/documentation/pull/1549 is approved

PR #1369, opened 2022-01-13
Remove unnecessary param from the create key path

brunoocasali:
The param `/:key` is not necessary 🍕 

image


dichotommy:
bors merge

PR #1386, opened 2022-01-25
Add missing redirect

dichotommy:
Redirect from old authentication guide location to new security guide location.

We missed this one at the time 😞 Thanks for the reminder @bidoubiwa !

### DIFF
--- a/.vuepress/public/_redirects
+++ b/.vuepress/public/_redirects
@@ -125,6 +125,9 @@
 # New errors page
 /errors                                             /reference/api/error_codes.html
 
+# Rename Authentication guide
+/reference/features/authentication.html             /reference/features/security.html
+
 # Move Security guide
 /reference/features/security.html                   /learn/advanced/security.html
 


.vuepress/public/_redirects
```diff
@@ -125,6 +125,9 @@
 # New errors page
 /errors                                             /reference/api/error_codes.html
 
+# Rename Authentication guide
+/reference/features/authentication.html             /reference/features/security.html
```
guimachiavelli, line 129: Shouldn't this be redirecting to `/learn/advanced/security.html` since we just moved the security guide?

```suggestion
/reference/features/authentication.html             /learn/advanced/security.html
```

dichotommy: @guimachiavelli and I discussed this during a synchronous meeting. I am in favor of maintaining the `_redirects` file as a history of URL changes on the docs and assume that Netlify optimizes them (e.g. instead of jumping twice, `reference/features/authentication.html` should redirect directly to `learn/advanced/security.html`).

We both agreed that the `_redirects` file needs to be better organized in the future, but that it is a low priority and won't be done immediately.

guimachiavelli (approved):
🦅

dichotommy:
bors merge

PR #1389, opened 2022-01-25
Use sudo to run the apt commands

irevoire:
By not using `sudo` to run the apt commands it could give the impression that you're running all the commands in a root shell which is not the case.


guimachiavelli (dismissed):
LGTM! Thanks! ✨

guimachiavelli:
bors merge

guimachiavelli:
come on, bors, you can do it

bors merge

PR #1390, opened 2022-01-25
Fix the compiling from the source part

irevoire:
We forgot to the `target-cpu=native` flag when compiling meilisearch from the source, so it had no impact on the performances currently.

### DIFF
--- a/learn/getting_started/installation.md
+++ b/learn/getting_started/installation.md
@@ -114,7 +114,8 @@ To learn more about the Windows command prompt, follow this [introductory guide]
 ::::
 
 ::: tip Compile for Best Performance
-For best performance, compile MeiliSearch on the machine you intend to run it on. This way, the binary is optimized for your specific architecture.
+For best performance, compile MeiliSearch on the machine you intend to run it on with the following environment variable set `RUSTFLAGS="-C target-cpu=native"`.
+This way, the binary is optimized for your CPU, and rust will use more aggressive optimizations.
 :::
 
 ## Cloud deploy


learn/getting_started/installation.md
```diff
@@ -114,7 +114,8 @@ To learn more about the Windows command prompt, follow this [introductory guide]
 ::::
 
 ::: tip Compile for Best Performance
-For best performance, compile MeiliSearch on the machine you intend to run it on. This way, the binary is optimized for your specific architecture.
+For best performance, compile MeiliSearch on the machine you intend to run it on with the following environment variable set `RUSTFLAGS="-C target-cpu=native"`.
+This way, the binary is optimized for your CPU, and rust will use more aggressive optimizations.
```
JeffAbrahamson, line 118: Quick comment: this is definitely clearer.

Have you ever measured the performance gained?  If you have, it would be a great place to say "We've seen performance gains of 10-25%" or whatever.

bidoubiwa: If the performance gain is not huge, it may be worth removing this part as it basically says "if you don't compile from source you'll have a slower experience". 
Which, even if through, if it is barely noticeable I don't think it is worth mentioning. 

Or in a more advanced guide about performances!

irevoire: Honestly, after some tests, it's not noticeable, so I'm up for removing this part entirely. :grin:

guimachiavelli: Do users get anything else from setting these flags and compiling Meilisearch on the same machine that will run the binary? Besides marginal performance increase, I mean.

If not, then the @meilisearch/docs-team is also in favour of removing this bit entirely.

bidoubiwa: I don't think so, @irevoire what do you think?

irevoire: We should remove everything.
I’ll update the PR

guimachiavelli:
Closing this PR in favour of #1412

PR #1391, opened 2022-01-25
Explain how to uses tags

irevoire:
If a user needs to compile from the source we want him to use a release instead of using the latest commit that was pushed on main.
This commit adds a little explanation on how the tags works

### DIFF
--- a/learn/getting_started/installation.md
+++ b/learn/getting_started/installation.md
@@ -85,7 +85,15 @@ git clone https://github.com/meilisearch/MeiliSearch
 cd MeiliSearch
 ```
 
-In the cloned repository, compile MeiliSearch.
+In the cloned repository, you can choose which release you want to use.
+See the list of all release [here](https://github.com/meilisearch/MeiliSearch/releases).
+Then run the following command whith whichever tags you selected:
+
+```bash
+git checkout v0.25.2
+```
+
+The last part is to compile MeiliSearch.
 
 ```bash
 # Update the rust toolchain to the latest version


learn/getting_started/installation.md
```diff
@@ -85,7 +85,14 @@ git clone https://github.com/meilisearch/MeiliSearch
 cd MeiliSearch
 ```
 
-In the cloned repository, compile MeiliSearch.
+In the cloned repository, you can choose which release you want to use.
+See the list of all release [here](https://github.com/meilisearch/MeiliSearch/releases).
+Then run the following command whith whichever tags you selected:
+```bash
+git checkout v0.25.2
```
irevoire, line 92: It would be cool if we could update the tag automatically to the latest version here but I don't know if it's possible.

JeffAbrahamson: You can set symbolic tags and reset them when you want to change the commit to which they point.  Then you have tag `v0.25.2` and tag `release` pointing to the same commit and users can say "I want to be at `release` without worrying about which that is.

That's especially helpful in automated environments, where I might have a script that automatically looks for a well-known tag.

irevoire: Hey @JeffAbrahamson, I was thinking about that, and I was wondering, isn't it way more straightforward and less confusing for our user to get a link to the latest GitHub release: https://github.com/meilisearch/meiliSearch/releases/latest
And then, the user can download a simple tarball without even using git or a terminal.

What do you think?

guimachiavelli:
So, if I understand this PR correctly, there are two things going on:
1. Making it clear users compiling from source can (and probably should) use version-tagged commits;
2. A discussion on whether the core repo should have a tag pointing to Meilisearch's most recent version (e.g. `latest`, `release`, etc).

We are close to releasing a major rewrite of the "Getting started" section at #1280. Those changes will render this PR obsolete.

My suggestion is: 
- we address the first concern (compiling from source instructions) in #1280 
- creating the `latest` tag (and keeping it updated with every release) is something that directly concerns the core repo management. If we feel this is an idea worth exploring, we move this discussion there.

irevoire:
I don’t think the @meilisearch/core-team actually want to have a `latest` tag that we need to update for every release.

But I think we should totally link the GitHub latest release thingy (https://github.com/meilisearch/meilisearch/releases/tag/v0.25.2) and tell people to download the tarball so there is no way for them to end up with an unstable version.

Can I close the PR?

guimachiavelli:
Ok, we'll add that link on our side. I'll close this, then ☺️ 

Thanks for the help, @irevoire!

learn/getting_started/installation.md
irevoire: The tarball contain the source code.
To me people fluent with git will have no problem with the link I gave since you can see which tag is associated with the release.
And people who don’t want to clone a whole repo because only want this version will be able to download only what they need in a compressed manner without the burden of knowing how to use git.

To me this is a win-win situation.
And at least we won’t ever give the impression that you should just use latest commit on `main` which was the initial issue.

PR #1393, opened 2022-01-25
Change colors palette of doc

bidoubiwa:
# Color changes in the documentation

I started with the official color but it was hard to read in our white themed doc. 
Following are screenshot of both colors. In this PR you can see the result with the 2nd example. If you'd like me to change it so that you can have a preview with the first color feel free to tell me the changes are already ready locally. I considered opening two PR with both colors but I wouln't want to spam. Tell me if that seems a better option. 
Also, we might want the opinion of Lena! 

If you want to vote (not sure it will be taken into account by the person in charge of the decision)
- color 1: ❤️
- color 2: 🚀
- color 3: 🎉

**Additional question**: should I change the color of the left border of the `notes` ? See 3rd screen. 

## Color 1

The primary color gives the following result. It may be because of my sight problem but it is harder to read for me.

Screenshot 2022-01-25 at 21 03 08
Screenshot 2022-01-25 at 20 59 50
Screenshot 2022-01-25 at 21 02 21
Screenshot 2022-01-25 at 21 00 32
Screenshot 2022-01-25 at 21 36 34

## Color 2

I prefer this color as I find that it pops out more. But I can definitely change it if required. 

Screenshot 2022-01-25 at 20 55 30

which results in the following: 

Screenshot 2022-01-25 at 20 56 44
Screenshot 2022-01-25 at 20 56 38
Screenshot 2022-01-25 at 20 56 30
Screenshot 2022-01-25 at 21 30 38
I was hovering on PHP

## Color 3

Just for the fun 

Screenshot 2022-01-25 at 21 46 52

Screenshot 2022-01-25 at 21 46 11
Screenshot 2022-01-25 at 21 46 28

Screenshot 2022-01-25 at 21 45 58
Screenshot 2022-01-25 at 21 46 06


guimachiavelli:
Regarding the border of notes/boxes/warnings/etc: I think we either go all in and change all of them using colours from the brand guidelines, or leave them be. My vote is to leave them alone for two reasons:

1. The colours for these boxes are not neutral: for instance, if we go with the red proposal, the notes container becomes red, which can be very easily associated with errors;
2. If we change the colour of notes, we need to make sure it meshes well with the other containers (e.g. warning, danger) so it feels like they all still belong to the same palette.

dichotommy:
I prefer color 2 for visual style, color 3 for readability.

bidoubiwa:
I'll change that @guimachiavelli thanks for the feedback :)

dichotommy:
bors merge

Issue #1293, opened 2021-11-18
Remove subsection readmes


guimachiavelli:
Despite the title, this is more of a proposal/discussion.

Every subsection of the docs website (e.g.: `/learn/core_concepts`, `/reference/features`) leads to a page that is basically just a collection of links preceded by a fairly unhelpful text. There are a few exceptions to this, namely `learn/what_is_meilisearch` and `learn/contributing`, but generally speaking these subsection roots don't add much.

@dichotommy, @maryamsulemani97: any opinions? What do you think about removing these pages so the title of a subsection leads to the first article in that folder?

dichotommy:
Fully support this. Would add `/reference/api` as another exception—it contains some useful information, just poorly organized.

guimachiavelli:
Ok, let's do it then. 

The next question is what to do with the "good" readmes. I suggest we:

- what_is_meilisearch/readme.md: move it to an intro article, first of its subsection
- contributing/readme.md: move it to the first (and only) article in that subsection
- reference/api: move it to an intro article

maryamsulemani97:
Fixed in https://github.com/meilisearch/documentation/pull/1490 and https://github.com/meilisearch/documentation/pull/1499

PR #1400, opened 2022-01-27
Remove useless READMEs

dichotommy:
Inspired by #1293 

This PR may not be merged. Most likely, it demonstrates how NOT to solve this problem. 🥲 Why?

- First of all, it is already creating problems with our `config-path-checker`—but this can be solved
- More importantly, the redirects added in `enhanceApp.js` to avoid 404s [may not work in production](https://github.com/vuejs/vuepress/issues/239#issuecomment-384277232). We could probably replace them with Netlify redirects using `.vuepress/public/_redirects`, but that still leaves some problems.
- We want clicking on a sidebar section title to open the first child of that section, but I haven't found a way to do this automatically. Instead it's hardcoded, which feels bad and makes the site harder to maintain.
- We may be impacted by other unforeseen consequences of eliminating our `README.md` files. Vuepress docs state `Each sub-directory in your static site should contain a README.md. It will automatically be converted to index.html.`

All that aside, this PR removes every README.md file except the one at root (for GitHub), plus `learn/what_is_meilisearch/README.md` and `reference/api/README.md` because they contain some content that isn't duplicated anywhere else.


.vuepress/config.js
```diff
@@ -124,7 +124,7 @@ module.exports = {
         },
         {
           title: '📕 Cookbooks',
-          path: '/learn/cookbooks/',
+          path: '/learn/cookbooks/running_production',
```
bidoubiwa, line 127: I think cookbook can keep its landing page with a bullet point of every cookbook. 
Since when clicking on `cookbooks` it opens the sub-navbar of `Run in production` it become harder to see the other ones.

Issue #1330, opened 2021-12-09
Move `reference/features/field_properties.md` into `learn/core_concepts/documents.md`


guimachiavelli:
This is part of our current site restructure; more specifically, our effort to remove the Feature reference and move all its pages/content to more adequate places across the docs site.

Field properties is an important page with fundamental content, but as it is I believe it lacks context. It is intimately connected to the concept of documents and its fields, which is why I propose we:

1. remove `reference/features/field_properties.md`
2. place the majority of its content inside `learn/core_concepts/documents.md#fields

The biggest drawback I can think of is bloating a page inside core concepts, which is a page I believe we'd ideally like to be as accessible and easy as possible.

Thoughts, @meilisearch/docs-team?

dichotommy:
`reference/features/field_properties.md` is a guide for `searchableAttributes` and `displayedAttributes`, which are both part of the settings object. Personally, I don't consider it super related to the concept of documents—but I see how it could be if working from a different understanding of what the core concepts pages are for.

I see the core concepts pages as being, well, conceptual. Not a guide for everything related to documents/indexes/relevancy, but a conceptual overview of those subjects. In reality, however, they've gradually become dumping grounds for all kinds of information. `reference/core_concepts/documents.md` contains all of the important information on primary keys, `reference/core_concepts/relevancy.md` contains a lot of key information on relevancy that isn't repeated elsewhere. This isn't really consistent with the idea of "core concepts" IMO and starts to stray away from the stated intention of the articles, making it less intuitive to find information (e.g. why is the lengthy explanation of the primary key in the "documents" article and not the "indexes" one? There's no logic.)

But I digress. **I would vote for maintaining this as a separate guide rather than folding it into one of the core concepts pages.** I think we should aim to keep the core concepts pages true to their name, insofar as that is possible and useful.

I would however like to change the name, as "field properties" seems like a useless layer of abstraction placed on top of what this is actually a guide for, which is searchable and displayed fields.

WDYT?

guimachiavelli:
I think that's a fair assessment. I was more focussed on the task of getting stuff out of the reference than on the integrity of core concepts, but I agree with both your view of an ideal core concepts page (sort of a high-level summary of a concept) and what it currently is (everything relating to a concept e.g. documents that we feel is important and don't really know where to put).

Your suggestion of reframing this guide from something relating to fields and towards regular settings like `filterableAttributes` is great too.

[...]

PR #1411, opened 2022-02-03
Move configuration, distinct attribute, field properties, and synonym guides

guimachiavelli:
Part of our ongoing attempt to remove the "Feature reference" section.


guimachiavelli:
So, over at #1330, we discussed a few ways to deal with these four guides. I suggested creating a "Configuration" subheading under "Advanced topics".

@dichotommy rightly pointed out that we would be mixing index settings and instance options by doing this, which could be confusing in the long term.

@meilisearch/docs-team, what do you think about creating a new "Configuration" heading with the same level as "Advanced topics"? Too big a change? [Take a look at how the "Learn" section looks](https://deploy-preview-1411--dreamy-wiles-8fd85b.netlify.app/learn/getting_started/quick_start.html) and let me know.

P.S.: this is only a proof-of-concept thing at the moment so we can see how the menu looks and works. Don't mind weird file locations/URLs.

.vuepress/config.js
```diff
@@ -105,6 +105,23 @@ module.exports = {
             '/learn/core_concepts/relevancy',
           ],
         },
+        {
+          title: '⚙️ Configuration',
```
maryamsulemani97, line 109: The headings "Configuration" and "Index settings" aren't clickable. Adding the path should fix that
path: '/learn/advanced/instance_options', after `title: '⚙️ Configuration'`,
 path: '/learn/advanced/distinct', after ` title: 'Index settings'`

maryamsulemani97:
Looks good! Could we have a brief introduction for "Index settings" though? This will give the reader some context on what to expect 
The "Instance options" start of with a brief introduction and you have everything on the same page. The "Index settings" is just three pages on different settings. 🤔

.vuepress/config.js
dichotommy: Is it weird that I like the fact that it's not clickable? 🤔

guimachiavelli:
Following @dichotommy's comments on the index settings during our morning meeting, what do you think we use the [current settings feature reference](https://docs.meilisearch.com/reference/features/settings.html) as the intro page, @maryamsulemani97?

dichotommy:
Could we rename the guide currently called "Field properties"? I've always disliked that name. Something like "Searchable and displayed attributes" seems better to me (reverse the order if you prefer).

.vuepress/config.js
guimachiavelli: I think what we have here is a "worst of all possible worlds" scenario: these menu items have :hover effects, but are not actually clickable. So we're visually saying "hey, you can interact with this thing!" but not following up on our promise, as it were.

Though I align a bit more with @dichotommy on this one and don't think these items need to be clickable, I think @maryamsulemani97 is right. Clicking on every other main menu heading ("What is Meilisearch", "Core concepts", etc) leads to the first element of that menu section (e.g. "Core concepts" leads to "Documents"), so changing behaviour only for "Configuration" would make the interface less consistent and more confusing because the user.

dichotommy: > Clicking on every other main menu heading ("What is Meilisearch", "Core concepts", etc) leads to the first element of that menu section (e.g. "Core concepts" leads to "Documents")...

This isn't really the case currently though, is it? Clicking on those navbar links you mentioned takes you to the `readme` for that directory (which I strongly dislike).

But yes, I agree with what you wrote: presenting unclickable menu items w/ a hover highlight is pretty unconscionable. I'm currently looking at the `config.js` file for VuePress's own documentation site to get a better sense of the "right way" to make section titles unclickable and get rid of some of our readmes.

guimachiavelli:
@bidoubiwa, do you have any idea why I'm getting this error on `check-links`?

```
  Path in config.json file is missing a trailing slash: path: '[object Object]' (/home/runner/work/documentation/documentation/.vuepress/config-path-checker/config.json:110:17)
```

Not only did I not touch `.vuepress/config-path-checker/config.json`, it does not actually exist in the repo, as far as I can see. 😵‍💫

bidoubiwa:
I fixed the error hander! 

see #1424

.vuepress/config.js
```diff
@@ -105,6 +105,25 @@ module.exports = {
...
+          collapsable: false,
+          path: '/learn/configuration/instance_options',
```
guimachiavelli, line 111: @bidoubiwa: thanks to @maryamsulemani97's advice, we have realised that the errors go away if we add a `/` to the `path` values in this line and L#116. 

Is this an error on check-md/config-path-checker? `/learn/configuration/instance_options` points to a file, not a directory, so requiring a trailing slash seems like a bug. Or are we missing something?

bidoubiwa: At the time it created issues with docs-scraper. I'm not sure it still does. Maybe we could just remove it from the CI and if we realise some pages are not indexed well we can consider putting it back. What do you think?

maryamsulemani97 (commented):
I can’t comment on the unmodified part of the file 🤷🏻‍♀️
Can we name the "Settings" page "Index settings"?

.vuepress/config.js
```diff
@@ -105,6 +105,24 @@ module.exports = {
...
         },
+        {
+          title: '⚙️ Configuration',
+          collapsable: false,
+          path: '/learn/configuration/instance_options/',
+          children: [
+            '/learn/configuration/instance_options',
+            {
+              title: 'Index settings',
+              collapsable: false,
```
maryamsulemani97, line 116: The link still isn’t clickable 👀

You may need to change this to `index_settings` if you decide to rename the page to replace "Settings" with "Index settings"
```suggestion
              collapsable: false,
              path: '/learn/configuration/settings/',
```

guimachiavelli: yeah, I was going to add the link, but then realised that none of our subheadings on the sidebar menus are links. "Inner workings," "Data backup," etc are all regular text, so I'm opting for consistency here.

@bidoubiwa is working on a second docs rebrand PR which will also remove the `:hover` state from these headings, so they won't look like links.

guimachiavelli:
@maryamsulemani97 re: changing the "Settings" guide, don't you think it might be a bit redundant to have an "Index settings" heading immediately followed by an "Index settings" guide?

dichotommy:
> Don't you think it might be a bit redundant to have an "Index settings" heading immediately followed by an "Index settings" guide?

Good point. How about just "Overview" for the sidebar name? (I think we can keep the file name `settings.md`)

dichotommy (commented):
I'm having some last minute hesitation on "Instance options". I'm concerned that the word "instance" here might not be as obvious to a general audience as it is to us, and could intimidate some. What do you think about something like "Startup options", "Launch options", or "Configure Meilisearch at launch"?

dichotommy (approved):
If Maryam's feedback is addressed, this looks good for me 👍🏻 
We can merge it and make a decision afterwards about "instance options".

dichotommy (approved):
LGTM! How did this end up with 37 commits 😂 Git troubles?

I guess it's fine that `Index settings` in the sidebar isn't clickable since it's consistent with other subsections (`Data backup`, `Inner workings`, `Deployment`). We should try to find a way to stop the hover highlight though (in a separate PR).

Anyway, great work 🚀 🪐

maryamsulemani97 (approved):
🧟‍♀️

PR #1412, opened 2022-02-03
Remove the compiling from the sources part

irevoire:
closes #1390 
After a little bit of testing, I noticed no difference by compiling from the source.
So instead of updating that part, we decided to remove it entirely, you can see the discussion here: https://github.com/meilisearch/documentation/pull/1390#discussion_r791981840


guimachiavelli:
bors merge

PR #1424, opened 2022-02-08
Fix config path checker error handler

bidoubiwa:
`path` is the whole API dealing with resolving paths and not the path that raised the error. Changed the variable to correct one. It raised these errors for you @guimachiavelli 
```
  Path in config.json file is missing a trailing slash: path: '/learn/configuration/instance_options' (/.../config.json:110:17)
  Path in config.json file is missing a trailing slash: path: '/learn/configuration/settings' (...config.json:115:21)
```

I also changed the link to where the error is thrown. Not sure why it was config.json.

### DIFF
--- a/.vuepress/config-path-checker/find-and-check.js
+++ b/.vuepress/config-path-checker/find-and-check.js
@@ -2,12 +2,13 @@ const { isObject } = require('./utils')
 const { addError } = require('./result-report')
 const path = require('path')
 
-function checkPath(givenPath) {
+function checkPath(givenPath, elem) {
   if (givenPath.slice(-1) !== '/') {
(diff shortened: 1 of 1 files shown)


bidoubiwa:
Ah I see what I did wrong. 

Inside the `config.json` that is at the root of `.vuepress` there is a link that is defined the following way: 
`/learn/configuration/settings` This link is missing a trailing slash. 

The problem is that I do not know on which line of the `config.json` the link is added. 
The only thing I can do is actually show the correct link to the config.json!

guimachiavelli:
I think this is fine. We do need to address it, but I'm satisfied with knowing the immediate cause/trigger of the issue for now.

Can I merge this into the parent branch?

guimachiavelli:
bors merge

PR #1429, opened 2022-02-10
Use webarchive to link to the deleted LMDB blog post

Kerollmops:
A linked blog post about the LMDB memory usage redirects to a dead link, I would prefer that we link to the old version of it instead.

I remember a discussion we had about hosting our own version of the post on our blog but for now, it would be cool to just forward to Webarchive.


dichotommy (approved):
LGTM 🥳 ⚡

dichotommy:
bors merge

PR #1436, opened 2022-02-14
Remove feature reference section

### LINKED ISSUE #1435: Remove empty Feature reference section
By merging #1427, we have finished moving everything out of the Feature reference section, which is now completely empty.

We need to take the final step and remove the section.

guimachiavelli:
The feature reference is no more, long live the feature reference.

Closes #1435

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -270,13 +270,6 @@ module.exports = {
             '/reference/api/error_codes',
           ],
         },
-        {
-          title: '⭐ Feature references',
-          path: '/reference/features/',
-          collapsable: false,
-          children: [
-          ],
-        },
       ],
     },
   },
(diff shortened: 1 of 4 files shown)


dichotommy (commented):
LGTM, though perhaps we should add a redirect for `/reference/features/`?

maryamsulemani97 (approved):
🏃🏻‍♀️

guimachiavelli:
> LGTM, though perhaps we should add a redirect for /reference/features/?

I was thinking about that. Felt a bit useless, given what the readme's actually are, but if you think it's necessary I'm happy to do it.

dichotommy (approved):
> Felt a bit useless, given what the readme's actually are, but if you think it's necessary I'm happy to do it.

Indeed, but always better than a 404 in the admittedly hard-to-imagine case that somebody bookmarked it.

LGTM, time to light the fuse and run for cover 💣 💥

guimachiavelli:
bors merge

PR #1439, opened 2022-02-14
Remove vuex from dependabot updates

bidoubiwa:
fixes #1433


guimachiavelli:
LGTM!

We should close #1433  without merging, correct?

dichotommy:
bors merge

PR #1444, opened 2022-02-16
v0.26: instance options new flag behavior

### LINKED ISSUE #1418: v0.26: Instance options: flags must not be followed by a value
## Summary
Currently, all CLI options require a value. This is true even for options which don't have any logical value, so `--no-analytics` must be written as `--no-analytics=true`. Confusingly, the value after these options does not matter, so both `--no-analytics=true` and `--no-analytics=false` will disable analytics.

v0.26 will fix this in two ways:

- 7 CLI options are now flags, meaning, they do not take a value. Assigning a value to flag will cause Meilisearch to throw an error;
- Environment variables accept `n`, `no`, `f`, `false`, `off`,  and `0` as false literals. Meaning, no `MEILI_NO_ANALYTICS=false` will not disable analytics.

The affected instance options are:

- `--no-analytics`
- `--schedule-snapshot`
- `--ignore-missing-snapshot`
- `--ignore-snapshot-if-db-exists`
- `--ssl-require-auth`
- `--ssl-resumption`
- `--ssl-tickets`

## Tasks
- [ ] Update Instance options guide Usage section
- [ ] Update rest of Instance options guide where necessary
- [ ] Optional: add information regarding error when providing a value to a flag
- [ ] Scan the rest of the docs for mentions of these options (e.g. `--no-analytics` in the telemetry page, etc). Dedicate special attention to examples.

## References
Issue on core: https://github.com/meilisearch/MeiliSearch/issues/2097
Discussion: https://github.com/meilisearch/documentation/discussions/1365#discussioncomment-2045001
Related issue: #1420

maryamsulemani97:
closes #1418

### DIFF
--- a/learn/configuration/instance_options.md
+++ b/learn/configuration/instance_options.md
@@ -18,6 +18,16 @@ Server is listening on: http://127.0.0.1:7700
 
 In the previous example, `./meilisearch` is the command that launches a Meilisearch instance and `--http-addr` is the option that sets the URL and port this instance will use. **All command-line options are prepended with `--`.**
 
+Some of these options are command-line flags and unlike command-line options, **they don't take any values**.
+
+```bash
+./meilisearch --no-analytics
+```
+
+The above example disables analytics for the Meilisearch instance without accepting any values.
+
+We will refer to command-line with the :triangular_flag_on_post: emoji on this page.
+
 ## Configuring an instance with environment variables
 
 In order to configure a Meilisearch instance using environment variables, you have to set the environment variable prior to launching the instance. If it's your first time doing this you may want to read more about [setting and listing environment variables](https://linuxize.com/post/how-to-set-and-list-environment-variables-in-linux/), or [use a command-line option](#configuring-an-instance-with-command-line-options) instead.
@@ -59,7 +69,7 @@ error: The argument '--schedule-snapshot <schedule-snapshot>' requires a value b
 
 ### Advanced
 
-- [Disable analytics](/learn/configuration/instance_options.md#disable-analytics)
+- [Disable analytics](/learn/configuration/instance_options.md#disable-analytics) :triangular_flag_on_post:
 - [Dumps](/learn/configuration/instance_options.md#dumps-destination)
   - [Dumps destination](/learn/configuration/instance_options.md#dumps-destination)
   - [Import dump](/learn/configuration/instance_options.md#import-dump)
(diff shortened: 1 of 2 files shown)


maryamsulemani97:
I added a 🚩 next to links that are flags. Since I couldn’t add it to the actual heading, I used a warning there. Lmk if this isn’t  a good idea.

learn/configuration/instance_options.md
```diff
@@ -18,6 +18,16 @@ Server is listening on: http://127.0.0.1:7700
...
 
+Some of these options are command-line flags and unlike command-line options, **they don't take any values**. If you use the flag, it counts as `true` and if you don't, it counts as `false`.
+
+```bash
+./meilisearch --no-analytics
+```
+
+The above example disables analytics for the Meilisearch instance without accepting any values.
+
+We will refer to command-line flags with the 🚩 emoji on this page.
```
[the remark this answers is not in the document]
guimachiavelli: Reviewing this based on the confusion regarding flags vs options on #1469, perhaps it's unwise to rely on a visual flag to indicate whether an instance option requires a value or not?

learn/configuration/instance_options.md
```diff
@@ -145,12 +155,17 @@ If no master key is provided in a `development` environment, all routes will be
...
 **Environment variable**: `MEILI_NO_ANALYTICS`
 **CLI option**: `--no-analytics`
-**Default value**: `false`
-**Expected value**: `true` or `false`
+**Default**: Enabled
+
+Deactivates Meilisearch's built-in telemetry when enabled.
 
-Deactivates Meilisearch's built-in telemetry when set to `true`.
+The environment variable accepts `n`, `no`, `f`, `false`, `off`, and `0` as false literals. An absent environment variable will also be considered as false, everything else is considered `true`.
```
guimachiavelli, line 168: Do you think "false literals" might be too technical here? Perhaps we should rephrase it to something simpler?

guimachiavelli (commented):
I think we should also edit the text on L#51: 

> **All configuration options must specify a value.**

This is not completely true now, right?

Also, a possibly bad idea: instead of mentioning all the "falsy" values environment variable flags accept (0, f, no, etc), what if we just mentioned it once on or close to L#51?

maryamsulemani97:
> Also, a possibly bad idea: instead of mentioning all the "falsy" values environment variable flags accept (0, f, no, etc), what if we just mentioned it once on or close to L#51?

I added them under each flag assuming that most users will probably jump to the individual flag instead of reading the first paragraph

learn/configuration/instance_options.md
```diff
@@ -38,7 +46,7 @@ Server is listening on: http://127.0.0.1:7700
 
 Command-line options take precedence over environment variables. If the same configuration option is specified both as a command-line option and as an environment variable, Meilisearch will use the command-line option and its respective value.
 
-**All configuration options must specify a value.** Using a command-line option or environment variable without specifying a value will throw an error and interrupt the launch process.
+Some configuration options must specify a value. Using a command-line option or environment variable without specifying a value will throw an error and interrupt the launch process.
 
 ```bash
 ./meilisearch --schedule-snapshot
```
curquiza, line 52: the example with --schedule-snapshot is not true anymore

guimachiavelli: Agree with @curquiza. I'm not a fan of replacing "All" with "Some," though, as it is a bit vague.

Perhaps we could go with "Unless otherwise noted, all configuration options must specify a value"?

Or just delete the whole paragraph? This was necessary when users had to specify values even for options that did not really have an actual logical value (most of what we are calling flags now).

dichotommy (approved):
# Great work on a tricky task

First off: the changes from v0.26 are well documented here. From a release-oriented perspective, I consider this PR ready to be merged.

In terms of responding to the changes in this PR, however, I see a number of ways it can be improved, such as removing the `Default` field from every flag. My suggestions are not necessary for or directly related to the release, so it may be better for us organizationally to wait until afterwards and open a fresh PR implementing the changes.

## Remove "Default" from flag descriptions

My major gripe with `Default` is that it is essentially meaningless for flags, which can all basically be understood as booleans with a default value of `false` (you could say that that is [the definition of a command-line flag](https://superuser.com/questions/1070059/is-there-a-difference-between-a-command-line-flag-and-a-command-line-option)). If necessary, we can add this information to the section at the beginning covering command-line flags.

## Accepted values for environment variables

About this text, which is currently copy-pasted in the description of every flag w/ an environment variable:

> The environment variable accepts `n`, `no`, `f`, `false`, `off`, and `0` as `false`. An absent environment variable will also be considered as `false`, everything else is considered `true`.

I think we should relocate this explanation to the "Environment variables" section at the beginning of the article, then link to that section in each of the flag-related warning boxes. IMO this would provide a much clearer UX—as it is now, I suspect the copy-pasted bits will make it more difficult for users to identify the key information for each flag. 

## Bring back the flag emoji 🚩 

I loved the use of the flag emoji in an earlier version of this PR, and would like to see if there's a way to add that to flag headers without breaking the anchor links.

At the very least we can add a cute flag emoji to the warning boxes! But like my other suggestions, this can be done after v0.26 is live—it's not necessary for the release.

guimachiavelli:
bors merge

PR #1446, opened 2022-02-16
V0.26 tenant tokens

guimachiavelli:


guimachiavelli:
## Instructions to reviewers
For this round, I'm leaving code samples directly on the guide itself.

This is the first draft of the tenant tokens guide. All feedback is welcome, but I'm particularly interested in knowing your thoughts on:

1. What do you think about the way the guide is structured? Is the gap between the section explaining token generation with an SDK and the specifics of search rule creation too wide?
2. What do you think about the new title for the "Security and permissions" guide (now "Master key and API keys")?

### For integrations team reviewers
I also have a secondary set of questions specifically for you:

1. On the "Using a tenant token with an SDK" example, I believe we should only have one code sample using vanilla JS, correct?
2. The code samples towards the end of the guide ("Tenant token settings") are abbreviated versions of the examples in "Using a tenant token with an SDK" highlighting specific parts of token generation. What do you think about this? Could it be a problem because of e.g. different naming conventions?

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,303 @@
...
+
+The encryption algorithm indicated by `alg` can be one of: `HS256`, `HS384`, or `HS512`. The choice of algorithm has no impact in tenant token behavior.
+
+Meilisearch only supports JWT tokens and `typ` must always be `JWT`.
+
+### Preparing the token payload
+
+Once the header is ready, you must create a payload. In this case, payload means the data containing the token's API key prefix, search rules, and optional expiry date.
+
+First, you must extract the first eight characters from a valid API key with access to the search endpoint:
```
maryamsulemani97, line 116: Idk if this is super important, but do we need to mention that the `apiKeyPrefix` can't be generated from a master key?

maryamsulemani97 (changes requested):
1. Could we have the settings first and then the token generation part?
2. How about "Key Management" / "Managing Keys"?

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,303 @@
+# Tenant tokens
+
+In this guide you'll first learn [what multitenancy is](#what-is-multitenancy) and [how tenant tokens help managing complex applications and sensitive data](#what-are-tenant-tokens-and-how-do-they-solve-multitenancy-in-meilisearch). Then, you'll see [how to generate and use tokens](#generating-tenant-tokens-with-an-sdk), whether with an official SDK or [otherwise](#generating-tenant-tokens-without-a-meilisearch-sdk). The guide ends with a quick explanation of [the main tenant token settings](#tenant-token-settings).
```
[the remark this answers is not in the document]
guimachiavelli: Well, the idea was born out of having an intro paragraph that also offers minor navigation capabilities to each section in the guide. Not sure it works, though.

Why do you think that too many links is not a good idea?

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,303 @@
...
+
+## What is multitenancy?
+
+In software development, multitenancy means that multiple users with different interests—also called tenants—share the same computing resources. Despite existing in the same environment, tenants are not aware of one another and do not have access to each other's data. Proper multitenancy is crucial in cloud computing services such as [Digital Ocean's Droplets](https://www.digitalocean.com/products/droplets) and [Amazon's AWS](https://aws.amazon.com/).
+
+Depending on the type of application you are developing with Meilisearch, it is possible you will face similar challenges. For example, if your application handles sensitive data like medical records, it is very important to make sure users can only search through their own documents.
+
+### What are tenant tokens and how do they solve multitenancy in Meilisearch?
```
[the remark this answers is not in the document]
guimachiavelli: Hmmm…interesting idea. I hadn't considered it, but I guess you're right some users might assume tokens are somehow stored by Meilisearch.

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,303 @@
...
+
+```json
+{
+  "patient_medical_records": {
+    "sort": "date:desc"
+  }
+}
+```
+
+Search rules can combine any number of search parameters. This rule requires all results to be sorted chronologically and excludes documents not belonging to belonging to a specific user:
```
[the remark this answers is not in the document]
guimachiavelli: I thought that the search rule errors were the same as regular search parameter errors. E.g., a search rule with an invalid value for the `filter` parameter would return an invalid filter error and so forth.

Is that not the case?

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,303 @@
...
+const search = medicalRecordsIndex.search('blood test');
+```
+
+Applications may use tenant tokens and API keys interchangeably when searching. For example, the same application might use a default search API key for queries on public indexes and a tenant token for logged-in users searching on private data.
+
+## Generating tenant tokens without a Meilisearch SDK
+
+Though [Meilisearch recommends using an official SDK](#generating-tenant-tokens-with-an-sdk), this is not a requirement. Since tenant tokens follow the JWT standard, you can use a number of [compatible third-party helper libraries](https://jwt.io/libraries). You may also skip all assistance and generate a token from scratch.
+
+The full process requires you to [create a token header](#creating-a-token-header), [prepare a data payload](#preparing-the-token-payload) with at least one set of search rules, and then [sign the token](#signing-a-token) with an API key.
```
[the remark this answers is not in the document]
guimachiavelli: The structure I chose was dictated by two main ideas:

- show users some code as soon as possible;
- be as usage-agnostic and beginner-friendly as possible.

These two choices mean that the hardcore data part of the guide (the one containing exhaustive descriptions of the payload) gets pushed to the bottom of the page, under "Tenant token settings." 

Do you think more and clearer links to the last section of this document, `## Tenant token settings` could help?

guimachiavelli:
Thank you @brunoocasali and @maryamsulemani97 for this initial round of feedback.

After reading your comments, I think the chosen structure is indeed a bit problematic, especially when it comes to the placement of info regarding the token payload (i.e. search rules, expiry date, api key.) 

After briefly consulting @dichotommy, I am going to add a minor summary explaining the payload to each one of the token generation sections. This means some content duplication, but is probably the best way of giving users the info they need exactly where they need it. The exhaustive reference will remain at the bottom of the page so advanced users with edge cases and etc can find all the information we have.

learn/security/tenant_tokens.md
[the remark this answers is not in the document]
brunoocasali: Unfortunately, if the user sets a bad filter like `"filter": "is_admin_user == true"` (should have only one `=` sign) the error will only be found when the user tries to use this token with the search method.

What I mean is, this error is not felt during the token generation but only in the usage of the generated token.

learn/security/tenant_tokens.md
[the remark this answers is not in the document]
brunoocasali: Yes, I think this could help! My point is more related to the idea of "explaining how to build a JWT is really necessary?". No problem if you think otherwise, just raising the hand about that! :D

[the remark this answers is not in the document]
guimachiavelli: Ah, sorry, I misunderstood the first part!

In the @meilisearch/docs-team we tend to prioritise helping less experienced programmers, so there's always the possibility we become too didactic. One of my references for this guide was [Google's IoT token guide](https://cloud.google.com/iot/docs/how-tos/credentials/jwts#iot-core-jwt-nodejs), which does go over token creation, though.

I'll try to get a few more opinions before making a decision. Thanks for raising this point!

learn/security/tenant_tokens.md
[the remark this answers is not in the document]
guimachiavelli: Ok, got it. Thanks!

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,299 @@
...
+{
+ "[index_name]": {
+   "[search_parameter]": "[value]"
+ }
+}
+```
+
+The object key must be an index name. You may also use `*` instead of a specific index name—in this case, search rules will be applied to all indexes.
+
+The object value must consist of `search_parameter:value` pairs. Currently, **tenant tokens only supports the `filter` [search parameter](/reference/api/search.md#filter)**.
```
gmourier, line 191: Love this whole section. ✨

gmourier (commented):
I don't know if it's because I have the feature in my head and I'm therefore biased but I found your explanations extremely clear and precise @guimachiavelli, this is top-notch.

The identified limitations are cited without being long or complex.

From memory, you don't mention that the API Key that signs the token must have the `search` action defined to be usable by an end-user. I think this might be important to clarify.

Thanks a lot for this PR.

learn/security/tenant_tokens.md
maryamsulemani97, line 191:
```suggestion
The object value must consist of `search_parameter:value` pairs. Currently, **tenant tokens only support the `filter` [search parameter](/reference/api/search.md#filter)**.
```

guimachiavelli: I was cringing very hard when I woke up today and immediately noticed this. I should stop pushing commits at the end of the day.

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,299 @@
...
+
+[You can read more about API keys in our dedicated guide.](/learn/security/master_api_keys.md)
+
+### Expiry date
+
+It is possible to define an expiry date when generating a token. This is good security practice and Meilisearch recommends setting relatively short token expiry dates whenever possible.
+
+The expiry date must be a UNIX timestamp or `null`. Additionally, a token's expiration date cannot exceed its parent API key's expiration date. For example, if an API key is set to expire on 2022-10-15, a token generated with that API key cannot be set to expire on 2022-10-16.
+
+Setting a token expiry date is optional, but recommended. Tokens without an expiry date are never expire and can be used indefinitely as long as its parent API key remains valid.
```
maryamsulemani97, line 282:
```suggestion
Setting a token expiry date is optional, but recommended. Tokens without an expiry date never expire and can be used indefinitely as long as the parent API key remains valid.
```

learn/security/tenant_tokens.md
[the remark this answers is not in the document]
maryamsulemani97: Dk if it's just me but when I see a bunch of links in a sentence like this, I just have to open each one individually and skim through it 🤡. But if I see the same content listed as bullets, I just move on as I will eventually see it later on the page.

learn/security/tenant_tokens.md
curquiza: I agree with guigui ze pmeuh 💪

curquiza (commented):
For me the guide is clear, thank you very much for making this non-easy feature that clear! ❤️

.code-samples.meilisearch.yaml
```diff
@@ -657,3 +657,12 @@ document_guide_add_document_primary_key: |-
         "overview": "A boy is given the ability to become an adult superhero in times of need with a single magic word.",
         "release_date": "2019-03-23"
      }]' 
+tenant_token_guide_search_no_sdk_1: |-
+  curl \
+    -X POST 'http://127.0.0.1:7700/indexes/patient_medical_records/search' \
+    -H 'Authorization: Bearer [yourTenantToken]'
```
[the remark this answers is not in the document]
guimachiavelli: You're absolutely right, thanks for spotting it!

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,299 @@
...
+   "[search_parameter]": "[value]"
+ }
+}
+```
+
+The object key must be an index name. You may also use `*` instead of a specific index name—in this case, search rules will be applied to all indexes.
+
+The object value must consist of `search_parameter:value` pairs. Currently, **tenant tokens only supports the `filter` [search parameter](/reference/api/search.md#filter)**.
+
+The following search rule forces every query across all accessible indexes to only return results whose `user_id` equals `1`:
```
dichotommy, line 193: This sentence though 👀 🧠 💢

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,299 @@
...
+
+```json
+{
+  "*": {
+    "filter": "user_id = 1"
+  }
+}
+```
+
+The next rule goes a bit further. When searching on the `patient_medical_records` index, a user can only see records that belong to them and have been marked as published:
```
dichotommy, line 203: I want very badly to change `further` to `farther` here but have no idea why, forcing me to confront the inescapable subjectivity of all of my suggestions 🙆🏻 so, thanks for that

learn/security/tenant_tokens.md
dichotommy, line 282:
```suggestion
Setting a token expiry date is optional, but recommended. A token without an expiry date never expires and can be used indefinitely as long as its parent API key remains valid.
```

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,299 @@
...
+The only way to revoke a token without an expiry date is to [delete](/reference/api/keys.md#delete-a-key) its parent API key.
+
+Changing an instance's master key forces Meilisearch to regenerate all API keys and will also render all existing tenant tokens invalid.
+:::
+
+When using an official Meilisearch SDK, you may indicate the expiry date when generating a token. Consult the documentation of the SDK you are using for more specific instructions.
+
+In [the previous SDK example](#generating-tenant-tokens-with-an-sdk), the expiry date is stored in `expiryDate`:
+
+```js
```
dichotommy, line 294: Not sure where to put this comment, but I am wondering if these three "code sample fragments" are really helpful. I think I understand the value of having a visual/code aid in this reference section, but aren't the payload elements/method arguments being discussed pretty SDK-agnostic? I imagine that it would be enough to show an example value for each payload element in a simple triple-backtick code block.

Then again, maybe there's some value I'm not considering, such as making it easy for users of typed languages to see the recommended data type for each argument. But even then, it's nothing that wouldn't be included in "tenant_token_guide_generate_sdk_1" above. So I'm inclined to replace these with inline code blocks rather than trouble the integrations team unnecessarily. WDYT?

.code-samples.meilisearch.yaml
```diff
@@ -657,3 +657,12 @@ document_guide_add_document_primary_key: |-
...
+tenant_token_guide_generate_sdk_1: |-
+tenant_token_guide_search_sdk_1: |-
```
brunoocasali, line 665: We need to use which one? `tenant_token_guide_search_sdk_1: |-` or `tenant_token_guide_sdk_search_1: |-` (they are different, here and in their usage)

brunoocasali:
@guimachiavelli @dichotommy I opened a PR in the Ruby SDK https://github.com/meilisearch/meilisearch-ruby/pull/301 with the addition of the samples. 
I think this could help us to understand better the entire doc (related to https://github.com/meilisearch/documentation/pull/1446#discussion_r818823122) :)

.code-samples.meilisearch.yaml
```diff
@@ -657,3 +657,12 @@ document_guide_add_document_primary_key: |-
...
+tenant_token_guide_fragment_rules_1: |-
+tenant_token_guide_fragment_key_1: |-
+tenant_token_guide_fragment_expiry_1: |-
```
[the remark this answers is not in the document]
guimachiavelli: hmmm…good point. I thought these were necessary to kickstart the sample fetching process.

@bidoubiwa, we're not going to have `curl` examples for most of these samples. Do we need to list them on `.code-samples.meilisearch.yaml` or is it enough they're on `sample-templates.yaml`?

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,299 @@
+# Multitenancy and tenant tokens
+
+In this guide you'll first learn [what multitenancy is](#what-is-multitenancy) and [how tenant tokens help managing complex applications and sensitive data](#what-are-tenant-tokens-and-how-do-they-solve-multitenancy-in-meilisearch). Then, you'll see [how to generate and use tokens](#generating-tenant-tokens-with-an-sdk), whether with an official SDK or [otherwise](#generating-tenant-tokens-without-a-meilisearch-sdk). The guide ends with a quick explanation of [the main tenant token settings](#tenant-token-payload-reference).
```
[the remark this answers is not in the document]
guimachiavelli: @maryamsulemani97 had similar thoughts on the links here. Will remove them.

re: warning—sure, makes sense.

learn/security/tenant_tokens.md
guimachiavelli: I think they're interchangeable in this case, no? I'm going to leave it as is and see how long before you make a PR changing it.

.code-samples.meilisearch.yaml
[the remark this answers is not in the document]
bidoubiwa: Yes! sample-template is enough :) Thanks !

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,299 @@
...
+};
+
+const expiryDate = 1645461882;
+
+const token = client.generateTenantToken(searchRules, expiryDate, apiKey);
+```
+
+There are three things you should pay attention to when using an SDK to generate a tenant token. In the above example, they are stored in `searchRules`, `apiKey`, and `expiryDate`.
+
+`searchRules` must be an object containing at least one search rule. A search rule must specify either an index name or a `*` wildcard. Search rules must also declare at least one filter. In this example, the filter makes it so a user only has access to their own records.
```
bidoubiwa, line 46: I'm not sure about this, unless I did not understand the sentence but these are valid and non valid formats of searchRules: 

✅
```
searchRules: ["*"]
```

❌
```
searchRules: "*"
```
✅
```
searchRules: {}
```

✅
```
searchRules: {
    movies: {}
}
```

✅
```
searchRules: {
    movies: null
}
```

✅
```
searchRules: {
    movies: {
       filters: "id = 1"
    }
}
```

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,299 @@
...
+
+`apiKey` must be a valid Meilisearch API key. A tenant token will have access to the same indexes as the API key you use when generating it. If no API key is provided, the SDK might be able to infer it automatically.
+
+`expiryDate` is optional when using an SDK. If used, this parameter must be a UNIX timestamp. Tokens becomes invalid after their `expiryDate`.
```
bidoubiwa, line 50: in the SDK`s we use `expiresAt` and it is an [`Date` instance](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date)

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,299 @@
...
+
+The full process requires you to [create a token header](#step-1-creating-a-token-header), [prepare a data payload](#step-2-preparing-the-token-payload) with at least one set of search rules, and then [sign the token](#step-3-signing-a-token) with an API key.
+
+If you are already familiar with the creation of JWTs and only want to know about the specific requirements of a tenant token payload, take a look at this guide's [token payload reference](#tenant-token-payload-reference).
+
+::: note
+The following examples use JavaScript for the sake of simplicity, but Meilisearch should be compatible with the majority of modern stacks and languages.
+:::
+
+### Step 1: Creating a token header
```
bidoubiwa, line 83: I know this might be late into the party but I would very much suggest you make an example using the [jsonwebtoken](https://www.npmjs.com/package/jsonwebtoken) library. 
Apart from the SDK's it's very very rare to create a JWT from scratch. I could barely find documentation online. 
Most of guides, [google's guide as well](https://cloud.google.com/iot/docs/how-tos/credentials/jwts), makes the example using that library!

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,299 @@
...
+
+The first step when generating a tenant token is to create a header stating token type and the chosen encryption algorithm. Once that is done you must encode the header object into `base64`:
+
+```js
+let header = {
+  "alg": "HS256",
+  "typ": "JWT"
+}
+
+const base64Header = base64Encode(header);
```
bidoubiwa, line 93: This function does not exists, the user has to use the `Buffer` API. 

```js
function encode64(str: any) {
  return Buffer.from(JSON.stringify(str)).toString('base64')
}
```

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,299 @@
...
+}
+```
+
+You can also optionally set an expiry date for your token. Tokens are invalid once past their expiry date.
+
+Like the header, it is necessary to encode the payload:
+
+```js
+const base64Payload = base64Encode(payload);
+```
```
bidoubiwa, line 141: Same as above:
```
function encode64(str: any) {
  return Buffer.from(JSON.stringify(str)).toString('base64')
}
```

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,299 @@
...
+
+You can read more about each of element of the tenant token payload in [this guide's final section](#tenant-token-payload-reference).
+
+### Step 3: Signing a token
+
+Tokens following the JWT standard must be signed. The signature is created by concatenating the encoded header and payload using a `.` as the separator, then encrypting the resulting string. **The encryption key must be the same you used when generating the payload's API key prefix.**
+
+```js
+const signature = HS256(base64Header + '.' + base64Payload, meiliSearchApiKey);
+```
```
bidoubiwa, line 151: This function does not exists, user will have to import the crypto standard library in Node 
```js
import crypto from 'crypto'
```

and then create the hash the following way: 
```js
return crypto
    .createHmac('sha256', apiKey)
    .update(`${encodedHeader}.${encodedPayload}`)
    .digest('base64')
    .replace(/\+/g, '-')
    .replace(/\//g, '_')
    .replace(/=/g, '')
```

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,299 @@
...
+
+```js
+const yourTenantToken = base64Header + '.' + base64Payload + '.' + signature;
+```
+
+### Using a tenant token without an SDK
+
+Your token is ready. Tenant tokens can seamlessly replace any requests made to the search endpoint:
+
+<CodeSamples id="tenant_token_guide_search_no_sdk_1" />
```
bidoubiwa, line 165: Is this a cURL code sample?

bidoubiwa (changes requested):
This is the review of the JWT from scratch creation

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,299 @@
...
+
+const searchRules = {
+  'patient_medical_records': {
+    'filter': `user_id = ${currentUserID}`
+  }
+};
+
+const expiryDate = 1645461882;
+
+const token = client.generateTenantToken(searchRules, expiryDate, apiKey);
```
bidoubiwa, line 41: The code sample will look like this: 

```js
const client = new Meilisearch({ host: 'meilisearch' });

const apiKey = '[an_api_key]';
const currentUserID = '[placeholder_user_id]';
const expiresAt = new Date('2030-04-01');
const searchRules = {
  'patient_medical_records': {
    'filter': `user_id = ${currentUserID}`
   }
};


const token = client.generateTenantToken(searchRules, {
   expiresAt: expiresAt,
   apiKey: apiKey,
});
```

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,299 @@
...
+
+```js
+// placeholder code sample. Final version will be supplied by the integrations team
+// <CodeSamples id="tenant_token_guide_sdk_search_1">
+const frontEndClient = new Meilisearch({
+  apiKey: token
+});
+
+const medicalRecordsIndex = frontEndClient.index('patient_medical_records');
+const search = medicalRecordsIndex.search('blood test');
```
bidoubiwa, line 66:
```js
const frontEndClient = new Meilisearch({
  host: "http://localhost:7700", // or mywebsite.com ?
  apiKey: token
});

const search = await frontEndClient
    .index('patient_medical_records')
    .search('blood test');
```

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,299 @@
...
+
+When using an official Meilisearch SDK, you may indicate which API key you wish to use when generating a token. Consult the documentation of the SDK you are using for more specific instructions.
+
+In [the previous SDK example](#generating-tenant-tokens-with-an-sdk), the API key is stored in `apiKey`:
+
+```js
+// placeholder code sample. Final version will be supplied by the integrations team
+// <CodeSamples id="tenant_token_guide_fragment_key_1">
+const apiKey = '[an_api_key]';
+const token = client.generateTenantToken(searchRules, expiryDate, apiKey);
```
bidoubiwa, line 267:
```suggestion
const token = client.generateTenantToken(searchRules, { apiKey });
```

ApiKey is actually optional because the apiKey provided by the client `client` can also be used. 
This works: 

```js
const client = new MeiliSearch({ host: "...", apiKey: "1234567" })

const token = client.generateToken(["*])
```
In this case it will use the `apiKey` of the `client`

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,299 @@
...
+
+When using an official Meilisearch SDK, you may indicate the expiry date when generating a token. Consult the documentation of the SDK you are using for more specific instructions.
+
+In [the previous SDK example](#generating-tenant-tokens-with-an-sdk), the expiry date is stored in `expiryDate`:
+
+```js
+// placeholder code sample. Final version will be supplied by the integrations team
+// <CodeSamples id="tenant_token_guide_fragment_expiry_1">
+const expiryDate = 1645461882;
+const token = client.generateTenantToken(searchRules, expiryDate, apiKey);
```
bidoubiwa, line 298:
```js
const expiresAt = new Date('2030-04-01');

const token = client.generateTenantToken(searchRules, { expiresAt: expiresAt );
```

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,299 @@
...
+
+```js
+// placeholder code sample. Final version will be supplied by the integrations team
+// <CodeSamples id="tenant_token_guide_fragment_rules_1">
+const searchRules = {
+  'patient_medical_records': {
+    'filter': `user_id = ${currentUserID}`
+  }
+};
+const token = client.generateTenantToken(searchRules, expiryDate, apiKey);
```
bidoubiwa, line 242:
```js
const searchRules = {
  'patient_medical_records': {
    'filter': `user_id = ${currentUserID}`
  }
};
const token = client.generateTenantToken(searchRules);
```

This works as the `apiKey` is optional. See the comment below!

bidoubiwa (commented):
Code-samples as they will be in the javascript SDK

learn/security/tenant_tokens.md
gmourier: Thanks, @bidoubiwa! I think this is the pseudo-code used in the specification that was presented here. I will update the specification when the js method is validated by someone from the @meilisearch/integration-team. 💪

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,299 @@
+# Multitenancy and tenant tokens
+
+In this guide you'll first learn [what multitenancy is](#what-is-multitenancy) and [how tenant tokens help managing complex applications and sensitive data](#what-are-tenant-tokens-and-how-do-they-solve-multitenancy-in-meilisearch). Then, you'll see [how to generate and use tokens](#generating-tenant-tokens-with-an-sdk), whether with an official SDK or [otherwise](#generating-tenant-tokens-without-a-meilisearch-sdk). The guide ends with a quick explanation of [the main tenant token settings](#tenant-token-payload-reference).
+
+## What is multitenancy?
+
+In software development, multitenancy means that multiple users with different interests—also called tenants—share the same computing resources. Proper multitenancy is crucial in cloud computing services such as [Digital Ocean's Droplets](https://www.digitalocean.com/products/droplets) and [Amazon's AWS](https://aws.amazon.com/).
```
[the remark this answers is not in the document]
gmourier: This explanation could be focused on the concept of multi-tenancy within Meilisearch or a database.

Something that might help: 

Single-tenant model: A Meilisearch index stores data for a single tenant.
Multi-tenant model: A Meilisearch index stores data for multiple tenants.

learn/security/tenant_tokens.md
bidoubiwa: Here you go: https://github.com/meilisearch/meilisearch-js/blob/e7b9c83ad6feb6044c76c8d71bf51372b2dbf34f/src/lib/token.ts

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,299 @@
...
+
+If your Meilisearch application stores sensitive data like medical records in the same index, it is very important to make sure users can only search through their own documents.
```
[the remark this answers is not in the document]
dichotommy:
```suggestion
If your Meilisearch application stores sensitive data belonging to multiple users in the same index, it is very important to make sure users can only search through their own documents. This can be accomplished with **tenant tokens**.
```

learn/security/tenant_tokens.md
guimachiavelli: Yes. L#660 on `.code-samples.meilisearch.yaml`

dichotommy (changes requested):
I think the only big change we need to make to this guide at this point is to give more information up front about the three parameters you can use to generate tenant tokens--as well as which are mandatory and in which contexts.

Instead of copy+pasting that information redundantly between the two tenant token generation sections, it would be better to put that information in a dedicated section before we even start talking about token generation.

This could take the form of moving the "tenant token payload reference" section to before "generating tenant tokens...", or if we think that reference is too overwhelming in its attempt to be complete, a shorter smaller section with three bullet points based on the content currently in the "generating tenant tokens with an SDK" section (line 44 currently).

Either way this is starting to look quite nice 🥳

learn/security/tenant_tokens.md
guimachiavelli: I understand what you mean, but it felt more user-friendly to have the SDK options here too. Especially in the case a user has already chosen an SDK (and thus explicitly named the language in which they want to see code samples), so it can be annoying/make understanding the sample a tiny bit harder if I'm e.g. a Rust dev and suddenly see some JS.

Also, and this might be more important, different SDKs have slightly different signatures (I believe JS accepts one search rules object and another for `exp` and `apiKey`, whilst Ruby accepts three separate arguments for the same values).

learn/security/tenant_tokens.md
bidoubiwa: the name of my variable is horrible, it should be 
```js
function encore64(data) {
  return Buffer.from(JSON.stringify(data)).toString('base64')
}
```

Then this function `JSON.stringify(data)` hidden in the middle, is transforming any data into a string (if it is a valid JSON but there is no reason why the user would use a non valid one, in which case it will throw an error).

learn/security/tenant_tokens.md
guimachiavelli: I believe the confusion should at least be partly cleared with @dichotommy's suggestion. 

Do let me know otherwise.

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,299 @@
...
+
+```js
+const yourSearchRules = {
+  "patient_medical_records": {
+    "filter": "user_id = 1"
+  }
+}
+```
+
+`searchRules` must be an object containing at least one search rule. A search rule must specify either an index name or a `*` wildcard. Search rules must also declare at least one filter. In this example, the filter makes it so a user only has access to their own records.
```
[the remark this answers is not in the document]
guimachiavelli: Same as above—we're on the no SDK section of the guide and I don't think we should take as a given that people would read both sections. 

I dislike content duplication as well, though. Only alternative I see to this is to start the guide with the full payload reference.

learn/security/tenant_tokens.md
guimachiavelli: Looks great. Only one comment: we have been moving away from marking placeholder values with `[]` (e.g. `[an_api_key]`). @dichotommy, @maryamsulemani97 do you think we should also enforce this in SDK examples?

learn/security/tenant_tokens.md
guimachiavelli: Thanks! I'm removing direct references to variables to reduce coupling between SDK code samples and the text. I'm also removing the bit stating what this parameter expects. Seems safer since we can't assume every language will have a `Date`-equivalent with that exact name.

learn/security/tenant_tokens.md
guimachiavelli: Can you expand a bit on this? I believe you mean it's too complex? I think I agree with that, but want to be sure I understood what you meant.

learn/security/tenant_tokens.md
bidoubiwa: I think removing `[]` is a good idea as well :) It is a bit confusing. If so should I replace it with `an_api_key` ? (same for user_id)

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,299 @@
...
+    'filter': `user_id = ${currentUserID}`
+  }
+};
+const token = client.generateTenantToken(searchRules, expiryDate, apiKey);
+```
+
+::: danger
+Because tenant tokens are generated in your application, Meilisearch cannot check if search rule filters are valid. Invalid search rules will only throw errors when they are used in a query.
+
+Consult the search API reference [more information on Meilisearch filter syntax](/reference/api/search.md#filter).
```
maryamsulemani97, line 248:
```suggestion
Consult the search API reference for [more information on Meilisearch filter syntax](/reference/api/search.md#filter).
```

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,299 @@
...
+
+```json
+{
+ "[index_name]": {
+   "[search_parameter]": "[value]"
+ }
+}
+```
+
+The object key must be an index name. You may also use `*` instead of a specific index name—in this case, search rules will be applied to all indexes.
```
maryamsulemani97, line 189: Maybe mention that * is a wildcard here because you mention overwriting wildcard rules in line 213? It maybe very obvious to some people, so Idk if its necessary

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,299 @@
...
+
+```js
+// placeholder code sample. Final version will be supplied by the integrations team
+// <CodeSamples id="tenant_token_guide_fragment_key_1">
+const apiKey = '[an_api_key]';
+const token = client.generateTenantToken(searchRules, expiryDate, apiKey);
+```
+
+::: note
+If an API key expires, any token tenants created with it will become invalid. The same applies if the API key is deleted or regenerated due to a changed master key.
```
maryamsulemani97, line 271: Did you use "token tenants" on purpose here?
```suggestion
If an API key expires, any tenant tokens created with it will become invalid. The same applies if the API key is deleted or regenerated due to a changed master key.
```

learn/security/tenant_tokens.md
guimachiavelli: Ok, cool, so let's go with that, then: placeholder values without `[]`. You probably don't need to update it here, as I'm going to replace the hardcoded sample with a `<CodeSample>` component soon.

Thanks!

learn/security/tenant_tokens.md
curquiza, line 7:
```suggestion
In software development, multitenancy means that multiple users with different interests—also called tenants—share the same computing resources. Proper multitenancy is crucial in cloud computing services such as [DigitalOcean's Droplets](https://www.digitalocean.com/products/droplets) and [Amazon's AWS](https://aws.amazon.com/).
```

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,235 @@
...
+
+```json
+{
+  "exp": 1646756934,
+  "apiKeyPrefix": "12345678",
+  "searchRules": {
+  "patient_medical_records": {
+    "filter": "user_id = 1"
+  }
+}
```
maryamsulemani97, line 125:
```suggestion
   }
}
```

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,235 @@
...
+
+After signing the token, you can use it to make search queries in the same way you would use an API key.
+
+<CodeSamples id="tenant_token_guide_search_no_sdk_1" />
+
+::: note
+The `curl` example presented here is only for illustration purposes. In production environments, you would likely send the token to the front-end of your application and query indexes from there.
+:::
+
+### Generating a tenant token with from scratch
```
maryamsulemani97, line 98:
```suggestion
### Generating a tenant token from scratch
```

guimachiavelli:
This is, hopefully, the final draft of the tenant tokens guide.

Based on feedback from @brunoocasali and @bidoubiwa, we have further de-emphasised generating tenant tokens without any external libraries and added a small section on creating tokens with a third-party dependency. This new section is very short, as we don't really want to explain how to use someone else's code. The "no SDK" section in general is perhaps not super useful, but I think its function is to make it very clear users are not **forced** to use our SDKs when they're using tenant tokens.

Regarding @dichotommy's feedback on the placement/redundancy of the token payload references, I have reduced the amount of copypasted text. There's still some repetition, but this is by design: 

- I don't want to have the tenant payload reference right after the intro. I believe this would be hostile towards inexperienced users;
- I dislike repeating information, but I don't think we can assume all users will read the guide from start to finish, so crucial bits of data must be repeated.

All feedback is welcome, but do keep in mind v0.26 is very close to release. This means we have limited time to actually do stuff, so I would prefer you kept further comments in this PR focussed on critical things we absolutely need to update. For bigger issues and non-essential/time-consuming polish, it might make more sense to open an issue so we can continue working on it after v0.26's release.

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,240 @@
...
+
+## Generating tenant tokens with an SDK
+
+Imagine that you are developing an application that allows patients and doctors to search through medical records. In your application, it is crucial that each patient can see only their own records and not those of another patient.
+
+The code in this example imports the SDK, creates a filter based on the current user's ID, and feeds that data into the SDK's `generateTenantToken` function. Once the token is generated, it is stored in the `token` variable:
+
+<CodeSamples id="tenant_token_guide_generate_sdk_1" />
+
+There are three important parameters to keep in mind when using an SDK to generate a tenant token: **search rules**, **API key**, and **expiration date**. Together,  they make the token's payload.
```
bidoubiwa, line 29:
```suggestion
There are three important parameters to keep in mind when using an SDK to generate a tenant token: **search rules**, **API key**, and **expiration date**. Together, they make the token's payload.
```

I removed a white space in the last part of the last sentence (its hard to see)

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,240 @@
...
+
+After creating a token, you can sent it back to the front-end. There, you can use it to make queries that will only return results whose `user_id` attribute equals the current user's ID:
+
+<CodeSamples id="tenant_token_guide_search_sdk_1" />
+
+Applications may use tenant tokens and API keys interchangeably when searching. For example, the same application might use a default search API key for queries on public indexes and a tenant token for logged-in users searching on private data.
+
+## Generating tenant tokens without a Meilisearch SDK
+
+Though Meilisearch recommends [using an official SDK to generate tenant tokens](#generating-tenant-tokens-with-an-sdk), this is not a requirement. Since tenant tokens follow the [JWT standard](https://jwt.io), you can use a number of [compatible third-party libraries](https://jwt.io/libraries). You may also skip all assistance and generate a token from scratch, though this is probably unnecessary in most production environments.
```
bidoubiwa, line 49: 🔥

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,240 @@
...
+  searchRules: {
+    'patient_medical_records': {
+      'filter': `user_id = ${currentUserID}`
+     }
+  },
+  apiKeyPrefix: apiKey.substring(0, 8),
+  exp: parseInt(Date.now() / 1000) + 20 * 60 // 20 minutes
+};
+
+const token = jwt.sign(tokenPayload, apiKey, {algorithm: 'HS256'});
```
bidoubiwa, line 73: Very cool example 👌. If you want to shorten it, by default `jsonwebtoken` uses `HS256` as default algorithm :) But it is oke to leave it for clarity if you prefer

bidoubiwa (approved):
LGTM 🔥🔥🔥🔥🔥🔥🔥🔥🔥

learn/security/tenant_tokens.md
guimachiavelli: Yeah, I was about to do that, based on the samples you sent over slack, but I realised it was useful to keep it. That option gives me a good hook to list the encryption algorithms Meilisearch supports.

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,240 @@
...
+
+As its name indicates, **API key** must be a valid Meilisearch API key. A tenant token will have access to the same indexes as the API key used when generating it. If no API key is provided, the SDK might be able to infer it automatically.
+
+**Expiration date** is optional when using an SDK. Tokens become invalid after their expiration date. Tokens without an expiration date will expire when their parent API key does.
+
+You can read more about each element of a tenant token payload in [this guide's final section](#tenant-token-payload-reference).
+
+### Using a tenant token with an SDK
+
+After creating a token, you can sent it back to the front-end. There, you can use it to make queries that will only return results whose `user_id` attribute equals the current user's ID:
```
maryamsulemani97, line 41:
```suggestion
After creating a token, you can send it back to the front-end. There, you can use it to make queries that will only return results whose `user_id` attribute equals the current user's ID:
```

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,240 @@
...
+The token header must specify a `JWT` type and an encryption algorithm. Supported tenant token encryption algorithms are `HS256`, `HS384`, and `HS512`.
+
+```json
+{
+  "alg": "HS256",
+  "typ": "JWT"
+} 
+```
+
+The token payload contains most of the relevant token data. It must be an object containing a set of search rules and the first 8 characters of Meilisearch API key. You may optionally set an expiration date for your token. Consult the [token payload reference](#tenant-token-payload-reference) for more information on how the requirements for each payload field.
```
maryamsulemani97, line 115:
```suggestion
The token payload contains most of the relevant token data. It must be an object containing a set of search rules and the first 8 characters of a Meilisearch API key. You may optionally set an expiration date for your token. Consult the [token payload reference](#tenant-token-payload-reference) for more information on how the requirements for each payload field.
```

maryamsulemani97 (approved):
🐼

learn/security/tenant_tokens.md
bidoubiwa: I agree !

brunoocasali (approved):
🥇 Congratulations @guimachiavelli and @meilisearch/docs-team awesome work! 🎉

dichotommy (approved):

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,240 @@
...
+The token header must specify a `JWT` type and an encryption algorithm. Supported tenant token encryption algorithms are `HS256`, `HS384`, and `HS512`.
+
+```json
+{
+  "alg": "HS256",
+  "typ": "JWT"
+} 
+```
+
+The token payload contains most of the relevant token data. It must be an object containing a set of search rules and the first 8 characters of a Meilisearch API key. You may optionally set an expiration date for your token. Consult the [token payload reference](#tenant-token-payload-reference) for more information on how the requirements for each payload field.
```
dichotommy, line 115:
```suggestion
The token payload contains most of the relevant token data. It must be an object containing a set of search rules and the first 8 characters of a Meilisearch API key. You may optionally set an expiration date for your token. Consult the [token payload reference](#tenant-token-payload-reference) for more information on the requirements for each payload field.
```

learn/security/tenant_tokens.md
```diff
@@ -0,0 +1,240 @@
...
+}
+```
+
+::: danger
+Because tenant tokens are generated in your application, Meilisearch cannot check if search rule filters are valid. Invalid search rules will only throw errors when they are used in a query.
+
+Consult the search API reference for [more information on Meilisearch filter syntax](/reference/api/search.md#filter).
+:::
+
+### API Keys
```
dichotommy, line 210:
```suggestion
### API Key
```

dichotommy (approved):
LGTM! Great work 🌊

PR #1464, opened 2022-02-21
v0.26 - update mentions of ISO 8601

### LINKED ISSUE #1458: v0.26: update mentions of ISO 8601
## Summary

There has been a confusion when implementing time formats in Meilisearch: though the specifications said ISO 8601, we have actually implemented RFC 3339 (which seems to be very similar but still has some minor formatting differences.)

We will continue using RFC 3339 everywhere in Meilisearch, so the docs need to replace mentions of ISO 8601 with RFC 3339.  

**Note:** the only exception to this update is the `duration` field included on `task` objects will still follow ISO 8601. This is because RFC 3339 does not define a way to express a process's duration.

## Tasks

- [x] Find mentions of ISO 8601
- [x] Replace mentions of ISO 8601 with RFC 3339
- [x] If automating the above tasks with a find-and-replace tool, double-check the Task API remains unchanged.

## References
Issue on core: https://github.com/meilisearch/meilisearch/issues/2172#issuecomment-1042798896
Spec: https://github.com/meilisearch/specifications/pull/114

maryamsulemani97:
closes #1458

### DIFF
--- a/learn/advanced/asynchronous_operations.md
+++ b/learn/advanced/asynchronous_operations.md
@@ -39,9 +39,9 @@ The response from the [task API](/reference/api/tasks.md) will always include th
 | `details`    | object  | Detailed information on the task payload                                                               |
 | `error`      | object  | Error details and context. Only present when a task has the `failed` status                                                |
 | `duration`   | string  | The total elapsed time the task spent in the `processing` state, in ISO 8601 format     |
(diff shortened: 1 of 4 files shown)


dichotommy (approved):
It all looks good to me! 🧐 🥂

guimachiavelli:
bors merge

Issue #1471, opened 2022-02-22
Warn users that docker on mac sucks


MarinPostma:
sup.

I've been using meilisearch on docker for a few days, and the recommended usage makes meilisearch suck on macOS (and probably windows too).

The issue is that all mounted volumes from osX to Linux need are on two different filesystems, so docker has to copy files back and forth, which is insanely slow.

I would be a good idea to recommend using the native application instead of people simply wanting to try meilisearch out.

dichotommy:
Fixed by #1472 and #1476

PR #1472, opened 2022-02-22
add warning about docker on macOS

MarinPostma:
following on #1471, I added a small warning for the user to avoid using meilisearch on docker on macOS and windows.

### DIFF
--- a/learn/getting_started/quick_start.md
+++ b/learn/getting_started/quick_start.md
@@ -62,6 +62,8 @@ docker run -it --rm \
 
 Data written to a **Docker container is not persistent** and is wiped every time the container is stopped. We recommend using a shared Docker volume between containers and host machines to provide persistent storage.
 
+This method is not recommended for macOS or Windows, because I/O between the host and the container can be very slow. For these operating systems, we recommend using the native application or a [cloud-hosted option](#cloud-deploy).
+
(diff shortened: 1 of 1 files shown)


learn/getting_started/quick_start.md
```diff
@@ -62,6 +62,8 @@ docker run -it --rm \
 
 Data written to a **Docker container is not persistent** and is wiped every time the container is stopped. We recommend using a shared Docker volume between containers and host machines to provide persistent storage.
 
+This method is not recommended for macOS and Windows, because I/O between the host and the container terribly slow.
```
dichotommy, line 65:
```suggestion
This method is not recommended for macOS or Windows, because I/O between the host and the container can be very slow. For these operating systems, we recommend using the native application or a [cloud-hosted option](#cloud-deploy).
```

dichotommy (changes requested):
Hey @MarinPostma thanks for the PR! 😄 I don't have the expertise to review it from a technical standpoint, but I trust your judgment.

If we want to discourage macOS and Windows users from testing Meilisearch w/ Docker, we should also delete the first line of the Windows tab in the same section; could you make that change in this PR please?

dichotommy (approved):
Looks beautiful, thanks again @MarinPostma ! 🌊 ⛵ 😊

dichotommy:
bors merge

PR #1477, opened 2022-02-24
Fix the link to the permitted list of actions

brunoocasali:
image

I think this link should point to:

image

But currently, it is pointing to the "actions" above:

image

I think we should create a new anchor for this case because what I understood is the docs created an anchor automatically since an anchor with the same name was already defined previously (but I'm not sure how to do that!).

### DIFF
--- a/reference/api/keys.md
+++ b/reference/api/keys.md
@@ -92,7 +92,7 @@ An array of API actions permitted for the key. `["*"]` for all actions.
 
 An array of indexes the key is authorized to act on. `["*"]` for all indexes.
 
-Only the key's [permitted actions](#actions) can be used on these indexes.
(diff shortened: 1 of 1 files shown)


dichotommy (commented):
Hey @brunoocasali , thanks for the PR! ❤️ 

The link is working properly on the deploy preview, but unfortunately it's not passing our link checker (darn thing isn't being maintained anymore 😭 ). 

@bidoubiwa I believe we're using your fork of `check-md` for Yarn package resolutions right now. Would you happen to know how to resolve this? 🤔

dichotommy (approved):
LGTM! 🌮

dichotommy:
bors merge

PR #1491, opened 2022-03-02
Update the geosearch examples

irevoire:
Currently, the geosearch examples return nothing because there are no points in a 2km range around `45.4628328, 9.1076931`.

image

Points 1 and 2 are the two points we already have in Italy.
And I made the query from the point 3 (less than 2km from 1&2).


guimachiavelli:
Thanks for this, @irevoire! 💛

guimachiavelli (approved):
bors merge

PR #1497, opened 2022-03-02
Fix note in updating.md guide

curquiza:
If I'm not wrong, according to this new addition in the [spec](https://github.com/meilisearch/specifications/pull/122/files) (sorry I should have done it before), the version number is not the right one in the note.


guimachiavelli (approved):
I think you're right. Thanks for the fix! 🥇

dichotommy (changes requested):

dichotommy (approved):
LGTM! Thank you very much @curquiza 🌻

PR #1503, opened 2022-03-07
Improve errors

### LINKED ISSUE #1299: Improve errors page
Our current errors page (`errors.yaml`) is lacking in a couple ways.

**Problems**:

- The build-time content fetching from a `.yaml` file does not add any value
  - These error descriptions are not repeated or used on any other pages besides the generated https://docs.meilisearch.com/errors/
- Using a `.yaml` file makes it difficult to integrate w/ VuePress
  - no sidebar
  - no place in the sitemap
  - treated as an external page—`.md` links do not work
- Using a `.yaml` file for this one page is inconsistent with our other site practices, and makes site maintenance less clear for external contributors and new team members
- It is not immediately clear to users or maintainers what is the content of this page, or what are its goals
  - Current content is a mix of a) error descriptions copy + pasted directly as they appear from the MeiliSearch error handler, and b) summarized error descriptions w/ tips for resolving them and links to places in the docs that can be helpful

**Solutions**:

- Make this a regular MarkDown page
- Clarify purpose and expectations of this page
  - Add an introduction stating up front what content is available on this page
  - Attempt to enforce a consistent format for each error, e.g. "Error message", "Why does this error happen", "How to resolve it", "Example"

This is not our highest priority, but it should be done within the next six months—ideally before or during the migration to Docusaurus (#932 ).

maryamsulemani97:
closes #1299

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -279,7 +279,18 @@ module.exports = {
             '/reference/api/health',
             '/reference/api/version',
             '/reference/api/dump',
-            '/reference/api/error_codes',
+          ],
+        },
+        {
+          title: 'Errors',
+          path: '/reference/errors/',
+          collapsable: false,
+          children: [
+            {
+              title: 'Overview',
+              path: '/reference/errors/overview',
+            },
+            '/reference/errors/error_codes',
           ],
         },
       ],
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -282,7 +282,7 @@ module.exports = {
           ],
         },
         {
-          title: 'Errors',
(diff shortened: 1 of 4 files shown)


guimachiavelli:
Something to keep in mind for this PR is that the specs often contain anchor links to errors. In an ideal world, we would probably create redirects for anchor links, not only the errors page, but:

1. I'm not sure our redirects work with anchor links (I suspect netlify might simply ignore anything coming after the `#`;
2. Even if redirects work with anchors, it might be a lot of work to continuously update these links every time we change anything in the errors page

More info in this discussion: https://github.com/meilisearch/specifications/pull/125/files#r822849402

Issue #1351, opened 2021-12-28
API reference reformatting


maryamsulemani97:
Based on [this](https://github.com/meilisearch/documentation/discussions/1347) discussion, we want to standardize and reformat the API references. 

## Possible template

1. A brief introduction + description 
2. Link to guide
3. List endpoints: (possible order)
    - List all
    - List one
    - Create
    - Update
    - Delete all
    - Delete one
4. For each endpoint (the following do not apply to all endpoints):
    1. What it does
    2. `<RouteHighlighter>` element
        1. Its [convention](https://idratherbewriting.com/learnapidoc/docapis_resource_endpoints.html#represent-path-parameters-with-curly-braces) to put path variables in {} in endpoints. But I see a lot of documentation without it as well
    3. Path variables
        1. Use a table with the name, type, description columns
    4.  Query parameters 
        1. Use a table with the name, type, description, default value (optional) columns
    5. Fields returned
        1. Use a table with the name, type, description columns
        2. What would nested objects look like?
    6. Request body
        1. Use a table with the name, type, description columns
        2. JSON sample
    7. Example 
        1. Code sample
        2. Response 
            1. Add a short description if needed
            2. Can we do [this](https://developers.onelogin.com/api-docs/1/users/create-user#) for each response? I think this will be a lot more helpful compared to the current errors page that we have. We can’t cover everything but maybe the major ones?
            3. Possible problems: May get too long

## Questions:
1. Should path variables be in bold or back ticks? **index_uid** or `index_uid`? 
2. Are there any parameters that have a max/min value?

maryamsulemani97:
Hello, 

I have created a [PR](https://github.com/meilisearch/documentation/pull/1357) for this. It's not a template, it's more of a visual reference for what we want our template to include. I also think we need colors. I love [Spotify's API reference](https://developer.spotify.com/documentation/web-api/reference/#/operations/get-current-users-profile), it doesn’t have too many colors. 
I like what I have for the pages called: POST, PUT, PATCH, and DELETE. If anyone has any suggestions, please lmk! 
The pages  called: GET visual reference 1 and GET visual reference 2 have different formats.
##  GET visual reference 1
There's 2 headings with the same format, I used two to see what the page would look like for an object with fewer fields vs. more fields. 
I used a table for explaining the fields returned. 
 ### Pros
 Its a table. Tables look super organized, the user will be able to view all the information without having to scroll too much.
### Cons
[...]

PR #1518, opened 2022-03-14
Update hash to point to the correct header w/ list of actions

dichotommy:
Update hash to point to correct page location (see #1477 )


maryamsulemani97 (approved):
🐼

guimachiavelli (commented):
Is there any way of avoiding two identical headings? I find links like `actions-2` very user unfriendly.

dichotommy:
@guimachiavelli I agree. The work we started last week on the API reference format (see #1351 ) should help with this eventually, but currently, I think it makes sense to repeat the key object explanation for both the get all keys endpoint and the create a key endpoint.

PR #1521, opened 2022-03-16
update known_limitations to add size of integer fields

maryamsulemani97:
Update `learn/advanced/known_limitations.md` to a add size of numeric fields as mentioned [here](https://github.com/meilisearch/meilisearch/issues/2205#issuecomment-1060817510)

### DIFF
--- a/learn/advanced/known_limitations.md
+++ b/learn/advanced/known_limitations.md
@@ -82,3 +82,9 @@ user = 1 OR user = 2 […] OR user = 1500 OR user = 1501 […] OR user = 2000 OR
   ["user = 1", "user = 2", […], "user = 1500", "user = 1501", […], "user = 2000", "user = 2001"]
 ]
 ```  
+
+## Size of numeric fields
+
+**Limitation:** Meilisearch can only exactly represent integers between -2⁵³ and 2⁵³.
+
+**Explanation:** Meilisearch stores numbers as double-precision floating-point numbers. Hence, it can only exactly represent integers between -2⁵³ and 2⁵³. This allows for greater precision and increases the range of magnitudes that can be represented.
--- a/learn/advanced/known_limitations.md
+++ b/learn/advanced/known_limitations.md
@@ -87,4 +87,4 @@ user = 1 OR user = 2 […] OR user = 1500 OR user = 1501 […] OR user = 2000 OR
 
 **Limitation:** Meilisearch can only exactly represent integers between -2⁵³ and 2⁵³.
 
-**Explanation:** Meilisearch stores numbers as double-precision floating-point numbers. This means it can only exactly represent integers between -2⁵³ and 2⁵³. Double-precision floating-point numbers allow for greater precision and increase the range of magnitudes that can be represented.
+**Explanation:** Meilisearch stores numeric values as double-precision floating-point numbers. This allows for greater precision and increases the range of magnitudes that can be represented.
--- a/learn/advanced/known_limitations.md
+++ b/learn/advanced/known_limitations.md
@@ -83,8 +83,8 @@ user = 1 OR user = 2 […] OR user = 1500 OR user = 1501 […] OR user = 2000 OR
 ]
 ```  
 
-## Size of numeric fields
+## Size of integer fields
 
 **Limitation:** Meilisearch can only exactly represent integers between -2⁵³ and 2⁵³.
 
-**Explanation:** Meilisearch stores numbers as double-precision floating-point numbers. Hence, it can only exactly represent integers between -2⁵³ and 2⁵³. This allows for greater precision and increases the range of magnitudes that can be represented.
+**Explanation:** Meilisearch stores numbers as double-precision floating-point numbers. This means it can only exactly represent integers between -2⁵³ and 2⁵³. Double-precision floating-point numbers allow for greater precision and increase the range of magnitudes that can be represented.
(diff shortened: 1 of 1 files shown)


learn/advanced/known_limitations.md
```diff
@@ -82,3 +82,9 @@ user = 1 OR user = 2 […] OR user = 1500 OR user = 1501 […] OR user = 2000 OR
   ["user = 1", "user = 2", […], "user = 1500", "user = 1501", […], "user = 2000", "user = 2001"]
 ]
 ```  
+
+## Size of numeric fields
+
+**Limitation:** Meilisearch can only exactly represent integers between -2⁵³ and 2⁵³.
```
guimachiavelli, line 88: I think we could drop the adverb, no? If we want to keep it, then I would suggest using `accurately` instead.

```suggestion
**Limitation:** Meilisearch can only represent integers between -2⁵³ and 2⁵³.
```

dichotommy, line 88: I'm worried about the use of the word integer here. Since I learned to program with C, integer to me means 32-bit integer, so between -2147483648 and 2147483647. That's a much smaller range (-2<sup>15</sup> to 2<sup>15</sup>). I'm probably not alone in this.

```suggestion
**Limitation:** Meilisearch can only exactly represent numeric values between -2⁵³ and 2⁵³.
```

dichotommy (changes requested):
🧗🏻

learn/advanced/known_limitations.md
maryamsulemani97: Maybe we should rename this to `Size of integer fields`. From what I understand, it can represent decimal values, but this double-precision floating-point format allows you to exactly represent integers between -2⁵³ and 2⁵³.
Does that make sense?

guimachiavelli: I thought that numbers outside that range would actually cause Meilisearch to crash. We definitely need an adverb there.

I always feel that "exactly" is a bit wonky, as if it operates more on a level of emphasis than of "quality of being exact/accurate." 

That said, there's precedent for that usage in a fairly reliable source like Wikipedia, so I'm fine with either. "Accurately" might read better, but "exactly" is probably more technically …accurate?

maryamsulemani97: I did think of "accurately", but since this is "technical" information, maybe we should go for "technically accurate" terms?

dichotommy: The mathematical + computing definition of integer AFAIK does not include decimals (AKA the infinite range of values between two whole numbers). 

Meilisearch does use the double-precision floating-point type under the hood, but in terms of what's being exposed to our users, it's just a numeric field with no specific type (and can include decimals, so using the term "integer" would be actively misleading here).

learn/advanced/known_limitations.md
```diff
@@ -82,3 +82,9 @@ user = 1 OR user = 2 […] OR user = 1500 OR user = 1501 […] OR user = 2000 OR
...
+
+**Explanation:** Meilisearch stores numbers as double-precision floating-point numbers. Hence, it can only exactly represent integers between -2⁵³ and 2⁵³. This allows for greater precision and increases the range of magnitudes that can be represented.
```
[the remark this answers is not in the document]
dichotommy: I see, thanks for the explanation! Will make a new comment w/ my suggestion.

learn/advanced/known_limitations.md
guimachiavelli: Ok, quite honestly, a lot of the subtleties of this discussion go over my head because floating-point operations are a Weird™ subject. All I can say for sure is that the linked issue explicitly names integers and I don't think a core dev would use that term lightly.

But let's take the guesswork out of this. @maryamsulemani97, please address and resolve all remaining open conversations in this PR. Then, request kéro's review. If you want to get to the bottom of this, you might want to get explicit confirmation the limitation we are documenting here concerns integers only.

learn/advanced/known_limitations.md
```diff
@@ -82,3 +82,9 @@ user = 1 OR user = 2 […] OR user = 1500 OR user = 1501 […] OR user = 2000 OR
   ["user = 1", "user = 2", […], "user = 1500", "user = 1501", […], "user = 2000", "user = 2001"]
 ]
 ```  
+
+## Size of integer fields
+
+**Limitation:** Meilisearch can only exactly represent integers between -2⁵³ and 2⁵³.
+
+**Explanation:** Meilisearch stores numeric values as double-precision floating-point numbers. This allows for greater precision and increases the range of magnitudes that can be represented.
```
Kerollmops, line 90: But integers that are only be represented with 64bits will be alternated. (the Wikipedia page explains in which way).

guimachiavelli: Would this be a more complete explanation, @Kerollmops?

```suggestion
**Explanation:** Meilisearch stores numeric values as double-precision floating-point numbers. This allows for greater precision and increases the range of magnitudes that Meilisearch can represent, but leads to inaccuracies in [values beyond certain thresholds](https://en.wikipedia.org/wiki/Double-precision_floating-point_format#Precision_limitations_on_integer_values).
```

Kerollmops: It is very good indeed, this is better!

guimachiavelli: Great! We'll commit this change and merge the PR. Thanks!

guimachiavelli (approved):
👾 
bors merge

guimachiavelli:
bors merge

PR #1534, opened 2022-03-21
Sample api reference 1

maryamsulemani97:
Update on #1357 

**Note:** If you’re looking at the deploy preview, **go to the file called ` Keys Sample`** under API references 

This format follows [Stripe's](https://stripe.com/docs/api/charges) format. It starts with explaining the object and then the endpoints. There is no "Returned fields" heading here. We say something like "Returns the keys object" with each response

Screen Shot 2022-03-21 at 16 54 38

### Pros:
- Looks a lot more organized 
- There is no repetition 

### Cons:
- Goes against our convention of H2=endpoints. I don’t think it's that big of a problem if we're consistent with it 
- Doesn’t use tables


maryamsulemani97:
Close once https://github.com/meilisearch/documentation/pull/1549 is approved

PR #1540, opened 2022-03-22
Create ecosystem dropdown menu in navbar

### LINKED ISSUE #1495: Site title cut off
Adding additional links to the top navbar cuts off the site title

Screen Shot 2022-03-02 at 16 35 20

dichotommy:
Economize horizontal screen space by removing GitHub, Slack, and Blog links from top navbar and placing them in a dropdown called ecosystem. Also 'API Reference' -> 'Reference'

The main questions:

- Is "Ecosystem" a clear enough name for this dropdown? (Idea taken from [Strapi docs](https://docs.strapi.io/developer-docs/latest/getting-started/introduction.html))
- Should the ['landing page'](https://www.meilisearch.com/) be included in this dropdown? If so, how should we refer to it?
  - Will get the opinion of a higher-up on this. May not make sense to direct traffic from the docs back to this product-focused page.

Closes #1495

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -4,7 +4,6 @@ module.exports = {
   description: 'Open source Instant Search Engine',
   theme: 'default-prefers-color-scheme',
   themeConfig: {
-    repo: 'meilisearch/Meilisearch',
     docsRepo: 'meilisearch/documentation',
     editLinks: true,
     lastUpdated: 'Last Updated',
@@ -14,12 +13,11 @@ module.exports = {
 
     nav: [
       { text: 'Learn', link: '/learn/getting_started/quick_start' },
-      { text: 'API Reference', link: '/reference/api/overview' },
+      { text: 'Reference', link: '/reference/api/overview' },
       { text: 'FAQ', link: '/faq' },
       {
         text: 'Integrations',
         items: [
-
           {
             text: 'SDKs',
             items: [
@@ -67,8 +65,14 @@ module.exports = {
           },
         ],
       },
-      { text: 'Slack', link: 'https://slack.meilisearch.com' },
-      { text: 'Blog', link: 'https://blog.meilisearch.com/' },
+      {
+        text: 'Ecosystem',
+        items: [
+          { text: 'GitHub', link: 'https://github.com/meilisearch/meilisearch' },
+          { text: 'Slack', link: 'https://slack.meilisearch.com' },
+          { text: 'Blog', link: 'https://blog.meilisearch.com/' },
+        ]
+      },
     ],
     sidebar: {
       '/learn/': [


maryamsulemani97 (approved):
>Should the ['landing page'](https://www.meilisearch.com/) be included in this dropdown? If so, how should we refer to it?

I don’t think we should. But if we do want to, we could do that on clicking the Meilisearch logo
like [Gatsby](https://www.gatsbyjs.com/docs/)

guimachiavelli:
Like the idea, dislike the choice of title. Ecosystem to me makes sense only when I think about it for a bit and/or click on the link to see what's inside the dropdown. Point being, it is not an immediate thing, and I'd wager most people's reactions would be "Ecosystem? That's not what I need, ain't nobody got time for that."

Not sure what to suggest…"Social"? "Resources"? "Etc"?

---

Regarding the landing page, I think a direct link is good. Not necessary, but in my head it establishes that the docs is a website separate but connected to meilisearch.com, so I'm for it. Not sure how to call a link like that, though. Definitely not "Landing page". Perhaps just "meilisearch.com"?

I'm also against making the logo link to meilisearch.com. I believe a docs user has very little interest in/very little to from visiting a more marketing-oriented page like meilisearch.com.

dichotommy:
> Should the ['landing page'](https://www.meilisearch.com/) be included in this dropdown? If so, how should we refer to it?

> I don’t think we should. But if we do want to, we could do that on clicking the Meilisearch logo
like [Gatsby](https://www.gatsbyjs.com/docs/)

> I'm against making the logo link to meilisearch.com. I believe a docs user has very little interest in/very little to from visiting a more marketing-oriented page like meilisearch.com.

I generally agree with you @guimachiavelli . The exception would be if we get rid of the docs.meilisearch.com homepage (`index.md`) and shift to a single-sidebar site. In that case I would somewhat expect clicking the logo to take me to the product site (since I would have access to all pages on docs.meilisearch.com through the sidebar).

guimachiavelli (commented):
I was about to approve this, not because I like "More Meili", but because I couldn't come up with anything better.

However, and I'm profoundly sorry about bringing this up, are we allowed to say Meili in official stuff?

guimachiavelli (approved):

.vuepress/config.js
```diff
@@ -67,8 +65,15 @@ module.exports = {
           },
         ],
       },
-      { text: 'Slack', link: 'https://slack.meilisearch.com' },
-      { text: 'Blog', link: 'https://blog.meilisearch.com/' },
+      {
+        text: 'More Meili',
```
maryamsulemani97, line 69:
```suggestion
        text: 'More Meilisearch',
```

guimachiavelli (approved):
LGTM! 📡 

bors merge

PR #1545, opened 2022-03-23
remove `/reference/api` folder

### LINKED ISSUE #1451: Remove /reference/api/ folder
Since there are no longer any other folders in `/reference/`, this folder is not necessary and makes URLs longer.

- [ ] Remove `/reference/api/` folder
  - [ ] Move all files from `/reference/api/` into `/reference/`
  - [ ] Delete `/reference/api/`
  - [ ] Update `config.js`
  - [ ] Update all links
  - [ ] Add redirects

maryamsulemani97:
closes #1451


guimachiavelli:
So, two things:

1. About what the PR is doing: unless I'm mistaken, we need to add a metric ton of redirects here. E.g. `reference/api/indexes.html` to `reference/indexes.html`, etc. Otherwise I believe users accessing old addresses would see 404 pages, which is not great;
2. Are we sure we want to do this? Isn't it possible we'll have content other than API stuff sitting under `reference/`? I would hate to move everything a folder up now, then return everything to a new `api` folder in a couple of months. An example is a reference for instance options—the article currently under `learn/` is really more of a reference, right?

Issue #1558, opened 2022-04-05
Update docs to add the Firebase/firestore integration 


maryamsulemani97:
- Update the [Official SDKs / libraries page](https://docs.meilisearch.com/learn/what_is_meilisearch/sdks.html)
- Update Comparison to alternatives [table](https://docs.meilisearch.com/learn/what_is_meilisearch/comparison_to_alternatives.html#features)
- Update the Integrations dropdown

[Link](https://github.com/meilisearch/firestore-meilisearch) to repository

PR #1559, opened 2022-04-06
add the Firebase/firestore integration

maryamsulemani97:
closes https://github.com/meilisearch/documentation/issues/1558

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -32,6 +32,7 @@ module.exports = {
               { text: 'Ruby', link: 'https://github.com/meilisearch/meilisearch-ruby' },
               { text: 'Rust', link: 'https://github.com/meilisearch/meilisearch-rust' },
               { text: 'Swift', link: 'https://github.com/meilisearch/meilisearch-swift' },
+              { text: 'Firebase', link: 'https://github.com/meilisearch/firestore-meilisearch' },
             ],
           },
           {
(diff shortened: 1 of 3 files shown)


maryamsulemani97:
bors merge

dichotommy:
From my perspective, "SDKs" is not quite the right spot for Firebase (but please ignore me if you were acting on the advice of the integrations team).

This has to do with the difference between SDKs and some of the other tools provided by the integrations team. I would argue that Firebase is not an SDK, it is closer to a platform plugin.

Here is an explanation using the terms from [this page](https://docs.meilisearch.com/learn/what_is_meilisearch/sdks.html):

- SDKs: basically libraries that provide methods making it easy to use Meilisearch in a given coding language. An SDK is linked to a specific coding language and must provide code samples.
- Framework integrations: similar to both SDKs and front-end tools, but with some key differences. Integrating w/ a specific web framework rather than a coding language (e.g. we have the Ruby SDK and the Ruby on Rails framework integration; the PHP SDK and the Laravel Scout framework integration; etc etc.). These could more correctly be seen as "back-end framework integrations"
- Platform plugins: extensions developed for a particular platform with a plugin ecosystem (e.g. VuePress, Gatsby, Strapi). My impression is that while these may sell themselves as frameworks, they are for the most part site/app builders of some kind. They also often have strict requirements for developing a plugin.
- Front-End tools: there's a bit of confusing overlap with **framework** integrations since these are also mostly web **frameworks** (e.g. Angular, React, Vue). However, since these mostly deal with front-end the content of the integrations is very different. These could be more correctly seen as "front-end framework integrations"
- DevOps tools: pretty self-explanatory, basically cloud infrastructure providers.

What do you think of this breakdown @meilisearch/integration-team , and where would you place Firebase within it?

maryamsulemani97:
>From my perspective, "SDKs" is not quite the right spot for Firebase (but please ignore me if you were acting on the advice of the integrations team).

I put it under SDKs as it's mentioned as an SDK in the [Comparison to alternatives](https://docs.meilisearch.com/learn/what_is_meilisearch/comparison_to_alternatives.html#features) table 
We might need to make changes here as well

brunoocasali:
I share an opinion with you @dichotommy that the Firebase/Firestore is integration, not an SDK.

For me "Platform plugins" seems appropriate!

guimachiavelli:
Sounds like a good idea to move it, then, as well as correcting the table mentioned by Maryam. 

In any case, since this PR has been merged, perhaps it's more constructive to move the discussion to an issue or another PR?

PR #1556, opened 2022-04-05
Remove path variables section from endpoints

maryamsulemani97:
closes #1527

PR #1569, opened 2022-04-11
Replace {uid} with {index_uid} or {task_uid} in API reference

dichotommy:
As we are planning to remove the path variable tables from all API endpoints, and as there are now multiple objects with the `uid` field in Meilisearch, it makes sense to be more specific about which "type" of uid an endpoint is expecting.

### DIFF
--- a/reference/api/indexes.md
+++ b/reference/api/indexes.md
@@ -52,7 +52,7 @@ Get information about an [index](/learn/core_concepts/indexes.md).
 
 | Variable      | Description   |
 | ------------- | ------------- |
-| **index_uid** | The index UID |
+| **uid** | The index UID |
 
 ### Example
 
@@ -134,7 +134,7 @@ This is an asynchronous task. [You can read more about asynchronous operations i
 
 | Variable      | Description   |
 | ------------- | ------------- |
-| **index_uid** | The index UID |
+| **uid** | The index UID |
 
 ### Body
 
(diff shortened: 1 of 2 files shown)


reference/api/indexes.md
```diff
@@ -44,15 +44,15 @@ List all [indexes](/learn/core_concepts/indexes.md).
 
 ## Get one index
 
-<RouteHighlighter method="GET" route="/indexes/{uid}"/>
+<RouteHighlighter method="GET" route="/indexes/{index_uid}"/>
```
maryamsulemani97, line 47: Agree with the different `uid`s being confusing, but `index_uid` doesn't technically exist, so is it okay to use it here?

maryamsulemani97:
@dichotommy, I'd recommend not updating the tables as part of this as I removed the tables in https://github.com/meilisearch/documentation/pull/1556

reference/api/indexes.md
maryamsulemani97: Maybe we should go with `index_uid` for consistency. We have used this format for [dumps](https://docs.meilisearch.com/reference/api/dump.html#get-dump-status) and indexes at different places:
- [Search](https://docs.meilisearch.com/reference/api/search.html#search-in-an-index-with-get-route) 
- [Documents](https://docs.meilisearch.com/reference/api/documents.html#get-documents)
- [Settings](https://docs.meilisearch.com/reference/api/displayed_attributes.html#get-displayed-attributes)

maryamsulemani97 (approved):
🧃

maryamsulemani97:
bors merge

PR #1573, opened 2022-04-12
fix typo in documents api ref

dichotommy:


reference/api/documents.md
```diff
@@ -232,7 +232,7 @@ You can use this `uid` to get more details on [the status of the task](/referenc
 
 <RouteHighlighter method="POST" route="/indexes/{index_uid}/documents/delete-batch"/>
 
-Delete a selection of documents based on array of document id's. The index [`uid`](/learn/core_concepts/indexes.md#index-uid) is required.
+Delete a selection of documents based on an array of document IDs. The index [`uid`](/learn/core_concepts/indexes.md#index-uid) is required.
```
maryamsulemani97, line 235: Do we want to use `document ID` instead `document id` everywhere?

dichotommy: Good question. I prefer it since I think `id` by itself can be visually confusing (case in point, if you're talking about multiple `ids`, or capitalizing it as part of a title `Id`, it kind of just looks like a typo).

dichotommy: Technically id is an abbreviation of "identifier" here—not an acronym of "identity document" like it is when you travel—and so it shouldn't be capitalized. However I think it might be more visually clear and simple if we just used "ID".

WDYT?

maryamsulemani97: >Technically id is an abbreviation of "identifier" here—not an acronym of "identity document" like it is when you travel—and so it shouldn't be capitalized. However I think it might be more visually clear and simple if we just used "ID"

Agreed

dichotommy: Okay, I'm closing this PR then to not introduce inconsistency in the site and allow the changes to be made together. Here's an issue: https://github.com/meilisearch/documentation/issues/1591

dichotommy: Thanks for noticing that this change would be inconsistent w/ what is already present on the site 😄

PR #1615, opened 2022-04-20
v0.27: Nested fields

### LINKED ISSUE #1582: v0.27: Add support for nested document fields
v0.27 allows indexation and searching of nested fields:

```json
{
  "id": 1,
  "title": "Red Wolf, Black Leopard",
  "author": {
    "firstName": "Marlon",
    "surname": "James",
  }
}
```

It's good to keep in mind that one of the main reasons this feature has been added is because many users assume it already exists. This indicates that our changes regarding this new feature should be relatively minor.

## Tasks
- [x] update learn/advanced/datatypes.md with info on nested fields;
- [x] update Index settings API reference of settings that accept attributes with a note/tip/etc block informing nested fields are supported;
- [x] update a few code samples so they include nested fields? Without any explanation, just so users see the behaviour naturally integrated across different use-cases

## References
- https://github.com/meilisearch/meilisearch/issues/2211
- https://github.com/meilisearch/specifications/pull/121
- SME: @irevoire

guimachiavelli:
This PR edits `learn/advanced/datatypes.md` to add a more thorough explanation of how document fields containing objects are internally parsed by Meilisearch, with a focus on nested objects. I have also updated `datatypes.md`'s intro and made minor styling changes to the content.

It also updates a few index setting pages so it is clear they accept values from nested fields using dot notation.

Resolves #1582

## Code sample updates

All code sample updates restricted to JSON responses. No further action required by other Meilisearch teams.

### DIFF
--- a/learn/advanced/datatypes.md
+++ b/learn/advanced/datatypes.md
@@ -63,9 +63,43 @@ You can also create [filters](/learn/advanced/filtering_and_faceted_search.md). 
 
 A Boolean value, which is either `true` or `false`, is received and converted to a lowercase human-readable text (i.e. `true` and `false`). Booleans can be searched as they are converted to strings.
 
+## `null`
+
+The `null` type can be pushed into Meilisearch but it **won't be taken into account for indexing**.
+
+## Array
+
+An array is an ordered list of values. These values can be of any type: numbers, strings, booleans, objects, or even other arrays.
+
+Meilisearch flattens arrays and concatenates them into strings. Non-string values are converted as described in this article's previous sections.
+
+### Example
+
+The following input:
+
+```json
+[
+  [
+    "Bruce Willis",
+    "Vin Diesel"
+  ],
+  "Kung Fu Panda"
+]
+```
+
+Will be processed as if all elements were arranged at the same level:
+
+```json
+"Bruce Willis. Vin Diesel. Kung Fu Panda."
+```
+
+Once the above array has been flattened, it will be parsed exactly as explained in the [string example](/learn/advanced/datatypes.md#examples).
+
 ### Objects
 
 When a document field contains an object, Meilisearch flattens it and brings the object's keys and values to the root level of the document itself.
+
+Keep in mind that the flattened objects represented here are an intermediary snapshot of internal processes. When searching, the returned document will keep its original structure.
 
 In the example below, the `patient_name` key contains an object:
 
@@ -147,39 +181,7 @@ After flattening, it would look like this:
 }
 ```
 
-This flattened object is an intermediary representation used during indexation. When searching, the returned document will keep its original structure.
-
-## `null`
-
-The `null` type can be pushed into Meilisearch but it **won't be taken into account for indexing**.
-
-## Array
-
-An array is an ordered list of values. These values can be of any type: numbers, strings, booleans, objects, or even other arrays.
-
-Meilisearch flattens arrays and concatenates them into strings. Non-string values are converted as described in this article's previous sections.
-
-### Example
-
-The following input:
-
-```json
-[
-  [
-    "Bruce Willis",
-    "Vin Diesel"
-  ],
-  "Kung Fu Panda"
-]
-```
-
-Will be processed as if all elements were arranged at the same level:
-
-```json
-"Bruce Willis. Vin Diesel. Kung Fu Panda."
-```
-
-Once the above array has been flattened, it will be parsed exactly as explained in the [string example](/learn/advanced/datatypes.md#examples).
+Once all objects inside a document have been flattened, Meilisearch will continue processing it as described in the previous sections. For example, arrays will be flattened, and numeric and boolean values will be turned into strings.
 
 ## Possible tokenization issues
 
--- a/learn/core_concepts/documents.md
+++ b/learn/core_concepts/documents.md
@@ -105,6 +105,8 @@ An attribute functions a bit like a variable in most programming languages, i.e.
 
 Every field has a [data type](/learn/advanced/datatypes.md) dictated by its value. Every value must be a valid [JSON data type](https://www.w3schools.com/js/js_json_datatypes.asp).
 
+If a field contains an object, you can refer directly to its internal properties using dot notation: `attributeA.objectKeyA`. Dot notation also works with nested objects: `attributeA.objectKeyA.objectKeyB`. This syntax is supported across Meilisearch, including index settings and search parameters.
+
 Take note that, in the case of strings, a value **[can contain at most 65535 positions](/learn/advanced/known_limitations.md#maximum-number-of-words-per-attribute). Words exceeding the 65535 position limit will be ignored.**
 
 You can also apply [ranking rules](/learn/core_concepts/relevancy.md#ranking-rules) to some fields. For example, you may decide recent movies should be more relevant than older ones.
--- a/reference/api/searchable_attributes.md
+++ b/reference/api/searchable_attributes.md
@@ -41,7 +41,6 @@ List the settings.
 
 Update the [searchable attributes](/learn/configuration/displayed_searchable_attributes.md#searchable-fields) of an index. The index [`uid`](/learn/core_concepts/indexes.md#index-uid) is required.
 
-
 If an attribute contains an object, you can use dot notation to set one or more of its keys as a value for this setting: `"searchableAttributes": ["release_date.year"]`.
 
 ::: warning


learn/advanced/datatypes.md
```diff
@@ -70,61 +65,127 @@ You can also create [filters](/learn/advanced/filtering_and_faceted_search.md).
 
 A Boolean value, which is either `true` or `false`, is received and converted to a lowercase human-readable text (i.e. `true` and `false`). Booleans can be searched as they are converted to strings.
 
-## Array
+## Object
```
irevoire, line 68: All this part describe the old way of handling objects.
In v0.27.0 we're always doing like you explained in the `nested object` part.

learn/advanced/datatypes.md
```diff
@@ -70,61 +65,127 @@ You can also create [filters](/learn/advanced/filtering_and_faceted_search.md).
...
+```json
+{
+  "id": 0,
+  "patient_name.forename": "Imogen",
+  "patient_name.surname": "Temult",
+}
+```
 
-JSON objects are written in key/value pairs and surrounded by curly braces. An object is broken into separate string tokens, which means separate words.
+Once that is done, the object can be flattened into a string:
```
irevoire, line 115: Same, we don't do that.

irevoire (changes requested):
Hey, that's pretty cool!

As I said in the comments, with the introduction of the nested fields we're not doing anything tricky with objects now. We flatten everything so we always handle simple values.

Also, I don't know if it's worth mentioning but if you have for example this object;
```json
{
  "nested": { "id": 1 },
  "doggos": [
    {
      "name": "bill",
      "race": "patou",
      "age": 8
    },
    {
      "name": "jean",
      "race": "bernese mountain",
      "age": 2
    }
}
```
In all settings, if you decide to set `doggos` as searchable for example.
Then `doggos.name`, `doggos.race` and `doggos.age` will get searchable.
Same for filterable, sortable, ranking rules, etc...

Another thing we may want to explain is that now we can use a nested primary key.
In this case, it would be `nested.id` BUT, the primary key inference won't happen if the key was nested.
So in this case meilisearch can't infer the primary key.

learn/advanced/datatypes.md
guimachiavelli: Ah, brilliant, thanks for pointing that out. 

Just to confirm, this would mean [this section of the spec](https://github.com/meilisearch/specifications/pull/121/files#diff-9c85f6e9818fa52f93114dd8c378870a6822b5cabc13a14993e5d718357f5106R144-R157) (lines 144-157) is out of date with v0.27 behaviour, right?

learn/advanced/datatypes.md
```diff
@@ -1,11 +1,8 @@
...
-This guide describes the different data types supported for the fields in a document and how Meilisearch handles them.
+This article explains how Meilisearch handles the different types of data in your dataset.
 
-No matter the type, the value of a field will remain unchanged in the returned documents upon search.
-For example, if you have a complex field with nested objects, this field will be returned with the same complexity upon search.
-
-Based on their type, however, the fields will be handled and used in different ways by Meilisearch. **The type affects how a field is used for search results**.
+**The behavior described here concerns only Meilisearch's internal processes** and can be helpful in understanding how the tokenizer works. The value of a field, meanwhile, remains unchanged for most practical purposes not related to Meilisearch's inner workings.
 
 [[toc]]
```
maryamsulemani97, line 7: I think we can get rid of this. We can see the headings in the sidebar

reference/api/displayed_attributes.md
```diff
@@ -42,6 +42,8 @@ List the settings.
 
 Update the [displayed attributes](/learn/configuration/settings.md#displayed-attributes) of an index. The index [`uid`](/learn/core_concepts/indexes.md#index-uid) is required.
 
+If your dataset contains nested structures, you can use dot notation to set a nested field as a displayed attribute.
```
maryamsulemani97, line 45: I think you should link the article here and the rest of the API reference 
```suggestion
If your dataset contains [nested structures](/learn/advanced/datatypes.md#nested-objects), you can use dot notation to set a nested field as a displayed attribute.
```

maryamsulemani97 (changes requested):

reference/api/displayed_attributes.md
guimachiavelli: A couple of things.

First of all, I clarified some aspects of the spec with @gmourier yesterday. To keep things short, the behaviour described there by default concerns all objects in a document field in the sense that all document fields containing objects are flattened to the root level of the document. I have simplified the text in this paragraph to reflect that.

Second, dot notation is a fairly simple topic for most programmers (`keyA.keyB[.keyX.etc]`). The behaviour described in `datatypes.md` is super specific to Meilisearch's core engine, not always easy to understand, and unlikely to be relevant to the majority of our users. I would like to avoid connecting them because I'm afraid a user might think e.g. they need to account for the stuff we do to arrays of objects.

gmourier (changes requested):
Thanks @guimachiavelli! This is great 👍 The review contains minor changes suggestions

learn/advanced/datatypes.md
```diff
@@ -70,61 +63,120 @@ You can also create [filters](/learn/advanced/filtering_and_faceted_search.md).
...
+  "patient_name": "Imogen Temult",
+  "appointments.date": [
+     "2022-01-01",
+      "2019-01-01"
+   ],
+  "appointments.doctor":[
+      "Jester Lavorre",
+      "Dorian Storm"
+   ],
+  "appointments.ward": ["psychiatry"]
```
maryamsulemani97, line 142:
```suggestion
  "appointments.ward": [
     "psychiatry"
]
```

learn/advanced/datatypes.md
```diff
@@ -70,61 +63,114 @@ You can also create [filters](/learn/advanced/filtering_and_faceted_search.md).
...
+  "id": 0,
+  "patient_name": {
+    "forename": "Imogen",
+    "surname": "Temult"
+  }
+}
 ```
 
-Will be processed as if all elements were arranged at the same level:
+During indexation, Meilisearch's tokenizer uses dot notation to eliminate nesting:
```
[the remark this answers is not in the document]
guimachiavelli: Unless @irevoire has very strong feelings against it, I'm fine with your edit and will apply it later today.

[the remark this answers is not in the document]
irevoire: Yep, the tokenizer isn't involved when we talk about field names. @gmourier sentence looks good to me :+1:

dichotommy (commented):
Happy about the improvements to `datatypes.md`! 🥳
However, I'm not fond of how the nested field examples were handled (granted, it was a tricky task). I'll explain why.

This line `If an attribute contains an object, you can use dot notation[... etc]` that has been copy-pasted onto several API reference pages describes general Meili behavior; however, a user reading this would not know if this piece of information applies only to this specific endpoint, or is generally applicable within Meilisearch. **That creates ambiguity,** which is the opposite of what you want from documentation/an example.

Once the reader realizes that indeed, it is possible to use dot-notation in any API parameter that accepts an attribute name, they may feel betrayed that this wasn't communicated clearly to them in a relevant location. They may also recognize the same sentence on a different page and scroll past the redundant info.

I'm most likely overstating things, but my concern is that this kind of choice can reduce user confidence in our docs and make it more difficult to parse and maintain API information. I think that **we should strive to fill API endpoint descriptions with content only related to that specific API endpoint**, within reason.

A comparatively elegant solution IMO would be changing a couple of the API reference POST endpoint code samples (by which I mean those in `.code-samples.meiliseach.yaml`) to use nested attributes. I think it would accomplish the same purpose—letting users know that this is possible—in a less distracting and more intuitive way.

I don't think that this should be changed right now; we wouldn't want to create workload for the integrations team by changing code samples at this late point in the release process. But if you agree with what I've said, we may want to slowly update some code samples to use dot-notation and gradually remove the copy-pasted sentences added in this PR in the future.

What do you think about adding some mention of nested fields / dot-notation in the Documents Core Concepts guide? In [this section](https://docs.meilisearch.com/learn/core_concepts/documents.html#fields) for example.

guimachiavelli:
I was on the fence about updating the API reference code samples, but  in the end opted not to do it for this release because of time constraints (release date coincides with the majority of the company being out-of-office). Absolutely fine with updating code samples after release to reflect this.

Though I agree the repeated text is not the most elegant solution, I'm not sure about it creating ambiguity or decreasing trust in the docs. Many users are likely to have a patchy and incomplete understanding of Meilisearch and we cannot assume they will e.g. read `/reference/api/settings.md` before `/reference/api/searchable_attributes.md`. In my head, explicitly stating dot notation is available on each page serves two scenarios: inexperienced users who appreciate the confirmation, and more experienced users who might only read the one or two settings page they need. Purely subjective, but as a developer I really wouldn't mind seeing the repeated text and, as you said, would quickly learn to scroll past it. I don't think they necessarily need to stay in their current form, though.

Regarding adding something to `/learn/core_concepts/documents.md`: great idea.

learn/advanced/datatypes.md
```diff
@@ -70,61 +63,122 @@ You can also create [filters](/learn/advanced/filtering_and_faceted_search.md).
...
 
 ## `null`
 
 The `null` type can be pushed into Meilisearch but it **won't be taken into account for indexing**.
 
+## Array
+
+An array is an ordered list of values. These values can be of any type: numbers, strings, booleans, objects, or even other arrays.
+
+Meilisearch flattens arrays and concatenates them into strings. Non-string values are converted as described in this article's previous sections.
```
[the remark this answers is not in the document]
guimachiavelli: As discussed over Slack, there is no exception to this behaviour. The "Objects" section on arrays of objects describes an intermediary processing step—meaning, those arrays will be eventually flattened as described here.

dichotommy (approved):
LGTM 🥸

guimachiavelli:
bors merge

PR #1635, opened 2022-05-16
Run CIs / fix scraper 404

dichotommy:
This PR re-orders the cloud deploy options in the quick start, but its actual purpose is to re-trigger our build-deploy process and scraper.

The docs Meilisearch instance is currently returning a result belonging to a deleted page. This results in a 404 when the user clicks on it. To resolve this I first tried re-running the scraper, with no success. However this could be due to a problem w/ that specific branch, so I'm retrying with a fresh deploy.


guimachiavelli (approved):
🌚

dichotommy:
bors merge

PR #1638, opened 2022-05-16
Update docs scraper to v0.12.2

dichotommy:
Should fix scraping bug, but may cause problems since this version targets v0.27, and the docs instance is v0.26. Time will tell but it's worth a try!


maryamsulemani97 (approved):

dichotommy:
bors merge

PR #1640, opened 2022-05-18
Apply the direct tag instead of latest alias

brunoocasali:
I found one good practice about [docker here #6](https://developers.redhat.com/blog/2016/02/24/10-things-to-avoid-in-docker-containers) that mentions the usage of the latest tag. They don’t recommend it because it could be misleading and may not represent the actual state of the image throughout the days.

Since this could avoid some misleading problems for the users, the idea is to update the docs with the correct tag every time.

### DIFF
--- a/learn/getting_started/quick_start.md
+++ b/learn/getting_started/quick_start.md
@@ -51,13 +51,13 @@ These commands launch the **latest stable release** of Meilisearch.
 
 ```bash
 # Fetch the latest version of Meilisearch image from DockerHub
-docker pull getmeili/meilisearch:latest
+docker pull getmeili/meilisearch:v0.27.1
 
 # Launch Meilisearch
 docker run -it --rm \
     -p 7700:7700 \
     -v $(pwd)/meili_data:/meili_data \
-    getmeili/meilisearch:latest
+    getmeili/meilisearch:v0.27.1
 ```
 
 Data written to a **Docker container is not persistent** and is wiped every time the container is stopped. We recommend using a shared Docker volume between containers and host machines to provide persistent storage.


learn/getting_started/quick_start.md
```diff
@@ -51,13 +51,13 @@ These commands launch the **latest stable release** of Meilisearch.
...
 # Fetch the latest version of Meilisearch image from DockerHub
-docker pull getmeili/meilisearch:latest
+docker pull getmeili/meilisearch:v0.27.1
 
 # Launch Meilisearch
 docker run -it --rm \
     -p 7700:7700 \
     -v $(pwd)/meili_data:/meili_data \
-    getmeili/meilisearch:latest
+    getmeili/meilisearch:v0.27.1
```
maryamsulemani97, line 60: How about something like:
```suggestion
# Replace `0.X.X` with the version you would like to install
    getmeili/meilisearch:v0.X.X
```
This way we won't need to make changes every time

dichotommy: That's a good idea, we could mention that the latest version can be found in the top left of the docs site (may not be true on mobile though?)

My only concern would be if the docker image is ready a few days or even a week after the release of a new version, in that time a user could try to run docker for the latest version and be confused when it fails.

dichotommy: The other downside of this approach is that it creates some friction for the user who has to edit the code sample. That's why I would generally opt for updating it ourselves, although it will be a pain.

@guimachiavelli Do you have any thoughts on this?

maryamsulemani97: >That's not a bad idea, we could mention that the latest version can be found in the top left of the docs site (may not be true on mobile though?)

And it won't show any patches 🌚

maryamsulemani97 (approved):
bors merge

PR #1645, opened 2022-05-19
update headings to sentence case

maryamsulemani97:
Updates any non-sentence case headings to sentence case

I think it looks wrong, especially
Screen Shot 2022-05-19 at 12 26 02


dichotommy:
I agree, it looks strange—particularly the site title and homepage H1. We should revert these.

I also think the quick start subheadings look a bit strange without the capitalization. Could we try dropping the "Step X" and see how that looks?

guimachiavelli (approved):
LGTM ⭐

maryamsulemani97:
bors merge

PR #1650, opened 2022-05-24
Add warning about authorization header change

### LINKED ISSUE #1649: Add warning about authorization header change
Since v0.25 the `X-MEILI-API-KEY: <apiKey> `header was replaced by the `Authorization: Bearer <apiKey>` header. Some users of older Meilisearch versions are confused when receiving the "missing authorization header" error because they try the new header with older versions. We could add a warning in the headers section of the docs or in the errors section?

maryamsulemani97:
closes #1649


learn/advanced/updating.md
```diff
@@ -18,7 +18,12 @@ Before we begin, you need to **verify the version of Meilisearch that's compatib
 
 If Meilisearch launches successfully, use the [get version endpoint](/reference/api/version.md), note your `pkgVersion`, and [proceed to the next step](#proceed-according-to-your-database-version).
 
-<CodeSamples id="updating_guide_check_version" />
+<CodeSamples id="updating_guide_check_version_new_authorization_header" />
+
+::: note
+ If you're using v0.24 or below, use the `X-MEILI-API-KEY: apiKey` authorization header:
```
maryamsulemani97, line 24: Not sure if having a code sample in a note is a good idea

guimachiavelli: I'm not sure I like the idea, but I'm also not sure I have any reason for not liking it? It doesn't look as bad as I expected on the preview.

Let's go with that. We can always change it if we don't like it once it's online or if we find a better solution.

guimachiavelli (approved):
LGTM 🦑

guimachiavelli (approved):
🧷

.code-samples.meilisearch.yaml
```diff
@@ -857,10 +857,14 @@ settings_guide_typo_tolerance_1: |-
...
         ]
       }'
-updating_guide_check_version: |-
+updating_guide_check_version_new_authorization_header: |-
   curl \
-    -X GET 'http://127.0.0.1:7700/version' \
-    -H 'Authorization: Bearer apiKey'
+      -X GET 'http://127.0.0.1:7700/version' \
+      -H 'Authorization: Bearer apiKey'
+updating_guide_check_version_old_authorization_header: |-
```
brunoocasali, line 864: How do you expect to us to translate that into code samples?

Because we do not support the old version anymore, and the user has no idea what is behind the usage, in other words: 

When they call a `client.index('movies')`, they do not choose what kind of header will contain `apiKey` as the value.
So I'm not sure if this is a valid code sample for the integration team to work on...

dichotommy: Hi Bruno, you are correct, the integration team does not need to work on this code sample, or any other that contains the old authorization method.

Actually, we have decided that the current version of the updating guide, which describes different processes for different versions of the authorization header and dumps, is too complex. We will be [simplifying this guide](https://github.com/meilisearch/documentation/issues/1723) in the future, including removing most of the code samples marked "old".

PR #1655, opened 2022-05-25
Add version of meilisearch-js in algolia migration guide

bidoubiwa:
The version of `meilisearch-js` is missing and impacts the code present in the guide a lot. 

To ensure the user install the correct version of the script I have added that specific version to the npm install command.

I also added some new lines where required (sorry for the out of subject action)


maryamsulemani97 (approved):
bors merge

PR #1658, opened 2022-05-25
Fix duplicate scrapping of some nested pages

bidoubiwa:
## Issue

For some reason, since, we think, the introduction of this [PR](https://github.com/meilisearch/documentation/commit/09dedf36f14a4a019bacfce3029b577f4c637526) some pages were scraped two times by docs-scraper.

These pages have a pattern, these are always the parent of a sub menu with some child's on which it points: 

Example: 
https://github.com/meilisearch/documentation/blob/6687a450c5c3c46cf357bbf502854cd8c329423f/.vuepress/config.js#L143-L157

In these case, `/learn/configuration/settings.html` and `/learn/configuration/settings/` were considered as two different url's by the scraper and thus scraped two times (but not on every run..).

Resulting in situation like this one: 
Screenshot 2022-05-25 at 18 07 07

## Problem solving

**docs-scraper**
Because this behaviour is **flaky** we thought we might find a solution inside [docs-scraper](https://github.com/meilisearch/docs-scraper/) which is the tool used to scrap the documentation.
Unfortunately, this is harder than expected as we are not able to reproduce the problem locally and when trying remote it is flaky so we need to wait a long time for every test.

**Updating the routing**

The working solution is to add the `html` extension to the path of a child:

Example: 
```js
              title: 'Index settings',
              collapsable: false,
              path: '/learn/configuration/settings.html', // Added the extension
              children: [
                {
                  title: 'Overview',
                  path: '/learn/configuration/settings',
                },
            //....
```

By updating the parent path to the exact link (with HTML extension) to its child, we avoid using a link like this one: `/learn/configuration/settings/` that is considered different than the one with the extension by the scrapper.
This may not be the best solution but it is one. 

In the past I created a plugin called `config-path-checker` that checked for trailing slashes in parents paths. 
This checker has been updated to check for trailing `html` extension in parents paths.

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -360,6 +360,7 @@ module.exports = {
     // Change are done in .vuepress/styles/palette.styl
     'vuepress-plugin-element-tabs',
     ['vuepress-plugin-container', { type: 'note' }],
+    [require('./config-path-checker')],
     [require('./custom-markdown-rules')],
     [require('./code-samples')],
     [
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -143,7 +143,7 @@ module.exports = {
             {
               title: 'Index settings',
               collapsable: false,
-              path: '/learn/configuration/settings/',
+              path: '/learn/configuration/settings',
               children: [
                 {
                   title: 'Overview',
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -80,7 +80,7 @@ module.exports = {
       '/learn/': [
         {
           title: '🔎 What is Meilisearch?',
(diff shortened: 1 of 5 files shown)


.vuepress/config.js
```diff
@@ -177,7 +177,7 @@ module.exports = {
             '/learn/advanced/sorting',
             {
               title: 'Updating Meilisearch',
-              path: '/learn/advanced/updating.md',
+              path: '/learn/advanced/updating',
```
maryamsulemani97, line 180:
```suggestion
              path: '/learn/advanced/updating.html',
```

maryamsulemani97 (approved):
bors merge

PR #1646, opened 2022-05-19
Add vale prose linter

maryamsulemani97:
This PR adds the [Vale prose linter](https://docs.errata.ai/) to the docs. This PR will add some basic rules, we will add to them as we go. You can find the list in [Notion](https://www.notion.so/meilisearch/Vale-84ca4f880e8f40a1b4946d494afb9291)

### For reviewers
- Is the message for each rule clear?
- Is the correct level (suggestion, warning, error) used?
- Are the links to further explanation good enough?
- ~~The `HeadingPunctuation.txt` file contains two rules that need to be updated~~
  - Updated to add `HeadingPunctuation.yml` and `Headings.yml`. Since these don't work for a lot of current headings, we can ignore these rules by adding:
 
```
style_guide.Headings = NO
style_guide.HeadingPunctuation = NO
```

to `vale.ini`

#### Current issues
- The `ignore` parameter only works for `spelling` extension point
  - How will we define file exceptions for certain rules?
- ~~What do we do about words that are meant to be typos? We should not add them to `accept.txt` and we can’t ignore the whole file.   Should we have these words in ``?~~
- Should we add names to `accept.txt`?
  - `learn/cookbooks/azure.md`
   - Maneu
  - `learn/what_is_meilisearch/comparison_to_alternatives.md`
   - Yonik
   - Bleve
   - Seeley
   - Hoxie
- ~~We use a lot of “condescending” words. See https://github.com/meilisearch/documentation/pull/1662~~
- ~~The Oxford comma rule isn't very reliable. It suggested adding a comma after "enable" for this sentence: `Finally, enable and start the Nginx service again to make sure it is still available.`~~ -> Now a suggestion
- The heading rules don't work for headings like "1. Click this"
- Currently, all typos in alt text are inside ''. Vale throws an error for them. They don't appear in the deploy preview if they are in backticks
 
  Example: 
  >  [Demonstrating the typo ranking rule by searching for `vogli`]

 **Deploy preview:** 
Screen Shot 2022-06-07 at 15 33 49

#### Current error count:

We have 0 errors, 23 warnings, and 125 suggestions
The warnings are mostly for words that are meant to be typos but can't be put in bacticks as they are used in alt text or for names
The suggestions are for:
- The FirstPerson rule which should ideally ignore files like resources `/faq.md` and `learn/what_is_meilisearch/telemetry.md`. Unfortunately, the `ignore` parameter doesn't work for this rule
[...]

PR #1662, opened 2022-05-26
sound less condescending

maryamsulemani97:
Update docs to sound less condescending. We might need to consider if [Condescending.yml](https://github.com/meilisearch/documentation/pull/1646/files#diff-de6e498212d2c7f02ba048aa9312f113b1f5a6f12244375115db5c685f67c5dc) is a good idea at this point. We use a lot of "easy"s, "simple"s, and "simply"s, and we can't get rid of all of them.

Part of https://github.com/meilisearch/documentation/pull/1646

### DIFF
--- a/learn/advanced/updating.md
+++ b/learn/advanced/updating.md
@@ -106,7 +106,7 @@ Now that you know which Meilisearch version your database is compatible with, pr
 
 ## Updating from v0.15.0 or above
 
-Because Meilisearch v0.15.0 and above include the [dumps feature](/learn/advanced/dumps.md), updating is relatively simple.
+Because Meilisearch v0.15.0 and above include the [dumps feature](/learn/advanced/dumps.md), updating is relatively simpler.
 
 In this guide, we will:
 
--- a/learn/configuration/displayed_searchable_attributes.md
+++ b/learn/configuration/displayed_searchable_attributes.md
@@ -36,11 +36,11 @@ There are two possible modes for the `searchableAttributes` list.
 
 This default behavior is indicated by a `searchableAttributes` value of `["*"]`. To verify the current value of your `searchableAttributes` list, use the [get searchable attributes endpoint](/reference/api/searchable_attributes.md#get-searchable-attributes).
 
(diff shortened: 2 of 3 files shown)


maryamsulemani97:
Around 37 if we exclude `/learn/contributing/contributing_to_docs.md` and `/resources/faq.md` as these files use a different tone compared to the rest of the docs. The current "condescending terms" include:

    - obvious
    - obviously
    - simple
    - simply
    - easy
    - easily
    - clearly
    - everyone knows

- If you look at the errors, many of the  "easy"s and "simple"s make sense. In that case, do we even need to consider them condescending? Should we define a new list?
- Our style guide does say we should avoid terms like these. Should we get rid of each of these terms or define a lot of exceptions?

dichotommy:
> If you look at the errors, many of the "easy"s and "simple"s make sense. In that case, do we even need to consider them condescending?

I would say yes. There may be exceptions, but for the most part saying that something is "easy" or "simple" is not helpful. Even in the case where you want your reader to know that something can be accomplished quickly, or without learning much background information, it's still more helpful to say exactly that than to use one of these words.

Because "easy", "simple" etc. are quite vague and subjective, they generally do not contribute a lot to a sentence. Sentences where the only idea is "X is easy to do" should probably be deleted and replaced with something more direct, such as "X can be installed in five minutes and requires no additional dependencies".

dichotommy:
I think we should give fixing all 37 errors a try. It could be fun and educational to work on it together, so let me know if you'd like a second eye.

learn/advanced/updating.md
```diff
@@ -106,7 +106,7 @@ Now that you know which Meilisearch version your database is compatible with, pr
 
 ## Updating from v0.15.0 or above
 
-Because Meilisearch v0.15.0 and above include the [dumps feature](/learn/advanced/dumps.md), updating is relatively simple.
+Because Meilisearch v0.15.0 and above include the [dumps feature](/learn/advanced/dumps.md), updating is relatively simpler.
```
dichotommy, line 109: Let's revert this change for now since we will be altering this guide soon and the other changes in this PR are ready to be merged.

```suggestion
Because Meilisearch v0.15.0 and above include the [dumps feature](/learn/advanced/dumps.md), updating is relatively simple.
```

dichotommy (approved):
LGTM 🕺🏻

PR #1665, opened 2022-05-30
Standardize curl code samples format

### LINKED ISSUE #1311: Standardize curl code samples format
We should standardize the format of our `curl` samples.

**All samples should:**

- Start with `curl \` on a single line before adding the main endpoint call and the header on subsequent lines
- Indent all lines after the first one time (two spaces)
- `\` character should be preceded by a single space character

**An additional concern:**

Currently, headers are always given after methods, e.g.

```
curl \
  -X POST 'http://127.0.0.1:7700/indexes/movies/documents' \
  --data-binary @movies.json
```

This may help users to see the route first, instead of a boring `Content-Type` header, which is certainly valuable. At the same time, it visually separates the method being used from the data payload being sent w/ it. It also seems more logical to me that the header should come before the method, e.g. 

```
curl \
  -X POST 'http://127.0.0.1:7700/indexes/movies/documents' \
  --data-binary @movies.json
```

I imagine that after a bit of time, users would get used to `Content-Type` being the first line on all of our POST endpoints and stop noticing it. But in the meantime, it may be less clear than the current order.

If you have an opinion about this, please let me know!

maryamsulemani97:
closes #1311 

This PR:
- replaces `127.0.0.1` with `localhost`
- fixes curl code indentation so it follows the [JSON format](https://www.notion.so/meilisearch/Style-Guide-61b848f4d4634beb9980492141a9e62e)


.code-samples.meilisearch.yaml
```diff
@@ -773,23 +825,25 @@ primary_field_guide_add_document_primary_key: |-
...
         "reference_number": 287947,
         "title": "Diary of a Wimpy Kid",
         "author": "Jeff Kinney",
-        "genres": ["comedy","humor"],
+        "genres": [
+          "comedy",
+          "humor"
+        ],
         "price": 5.00
      }]'
```
guimachiavelli, line 838: Shouldn't these also be separated into two different lines?

.code-samples.meilisearch.yaml
```diff
@@ -853,60 +901,57 @@ settings_guide_typo_tolerance_1: |-
...
-    -X GET 'http://127.0.0.1:7700/indexes/{index_uid}/settings/displayed-attributes' \
+    -X GET 'http://localhost:7700/indexes/{index_uid}/settings/displayed-attributes' \
     -H 'Authorization: Bearer apiKey'
 updating_guide_get_displayed_attributes_old: |-
   # whenever you see {index_uid}, replace it with your index's unique id
   curl \
-    -X GET 'http://127.0.0.1:7700/indexes/{index_uid}/settings/displayed-attributes' \
-    -H 'X-Meili-API-Key: apiKey'
+    -X GET 'http://localhost:7700/indexes/{index_uid}/settings/displayed-attributes' \
+    -H 'X-Meili-API-Key: apiKey'
```
guimachiavelli, line 919: Are the spaces after `apiKey'` required?

guimachiavelli (changes requested):
Two comments I think we should address before approving. Both of them minor, though

.code-samples.meilisearch.yaml
maryamsulemani97: Do you mean:

```
"price":
   5.00
```

.code-samples.meilisearch.yaml
maryamsulemani97: Nope

dichotommy (commented):
I'm noticing some inconsistency in indentation size. I think this comes from the fact that most of our JSON snippets use 4-space indentation, but our curl samples use 2-space. It would be good to make a decision on this ASAP. Ideally it would be consistent across curl code samples and JSON snippets.

From what I've read, 4-space indents are the default: they are used by most IDEs and generally provide a better reading experience. The main drawback is that they take up a lot of horizontal space, so nested code can quickly go off the edge of the screen requiring scrolling. This is a big deal for us, since our code blocks are pretty small in width.

I think we should try 2-space indentation for all code samples and json snippets, starting with the ones in this file. Could you [format the document](https://codesource.io/how-to-fix-vscode-indentation/) with 2 space indentation please @maryamsulemani97 ?

guimachiavelli (approved):
LGTM 🦆 
FYI, my experience with 4 vs 2 spaces for indentation is that the absolute majority of programmers uses 2. I also have the impression that the more inexperienced the developer is, the more likely they are to use 4 spaces. Perhaps something related to starting with the default and slowly realising that they can fit more code in the screen by using smaller indentation increments?

PR #1669, opened 2022-05-30
update movies.json

### LINKED ISSUE #1664: update movies.json
https://docs.meilisearch.com/movies.json seems to be an older version than what is on the demo at meilisearch.com

maryamsulemani97:
closes #1664

### DIFF
--- a/learn/advanced/filtering_and_faceted_search.md
+++ b/learn/advanced/filtering_and_faceted_search.md
@@ -27,7 +27,7 @@ Suppose you have a collection of movies containing the following fields:
 ```json
 [
   {
-    "id": "458723",
+    "id": 458723,
     "title": "Us",
     "director": "Jordan Peele",
     "genres": [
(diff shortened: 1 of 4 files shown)


dichotommy (commented):
LGTM. The only significant change I notice at first glance is that the `id` field is now numeric and no longer a string. Are there any code samples we would want to update to reflect that?

(Making our code samples consistent w/ the real movies.json is not a high priority, but if it's easy to do then it might be worth it just to avoid a user opening an issue down the line 😂 )

maryamsulemani97:
>The only significant change I notice at first glance is that the id field is now numeric and no longer a string. Are there any code samples we would want to update to reflect that?

Good catch.
- We'll need to update most of our JSON responses to `id` not being a string
- We'll need to update examples where we set random strings as the `id`. (Example: `learn/core_concepts/documents.md` line 37)

Or we could update `id` to be a string for this dataset. A single find and replace should work, but then this won't be consistent with the dataset in the [dataset](https://github.com/meilisearch/datasets) repository 😕

dichotommy:
> We'll need to update most of our JSON responses to id not being a string
> We'll need to update examples where we set random strings as the id. (Example: learn/core_concepts/documents.md line 37)

Agreed. Could you make these changes within this PR?

> Or we could update id to be a string for this [dataset](https://github.com/meilisearch/datasets). A single find and replace should work, but then this won't be consistent with the dataset in the dataset repository 😕

It would indeed be easier to do it that way, but I think it's important to maintain some consistency with the dataset repo.

maryamsulemani97:
bors merge

PR #1677, opened 2022-06-01
Add `old` to the code sample name for old header

alallema:
Some code samples use the old `X-Meili-API-Key` header, to avoid confusion between an error and the fact that this header is used voluntarily it would be nice to add the suffix `old` at the end of the sample code name as it is already the case for some.
If you have a better idea I'm interested!
See with @dichotommy


dichotommy (commented):
Thank you @alallema 😊 

Could you also update the file `.vuepress/public/sample-template.yaml`?

maryamsulemani97:
And `/learn/advanced/updating.md` 🙏

maryamsulemani97 (approved):
LGTM, thanks @alallema! 🧚🏻‍♀️

bors merge

PR #1679, opened 2022-06-02
Move authorization after front-end in quick start

dichotommy:
Users following the quick start sequentially will have problems with the current order, since the front-end integration code samples are not configured to work with a protected instance.

This is a temporary solution just for the weekend until we can add instructions on retrieving the search API key and incorporate that into the front-end code samples.


dichotommy:
bors merge

PR #1682, opened 2022-06-06
update quick start to add note on master key

maryamsulemani97:
Adds note to [Securing Meilisearch](https://docs.meilisearch.com/learn/getting_started/quick_start.html#securing-meilisearch) that the master key should only be used for managing API keys


maryamsulemani97:
bors merge

PR #1715, opened 2022-06-13
remove `getting_started_communicating_with_a_protected_instance`

maryamsulemani97:


dichotommy:
Tell me if you think I'm being crazy... but it seems like a useful code sample to me. It shows how to search on movies with an API key for authorization. I think we can find a place for it outside of the 101.

What do you think about adding this sample at the end of the security section of the getting started? We can turn the warning that's currently present there into regular text and use it to introduce the code sample: "The master key should only be used for retrieving and managing API keys. For regular API calls, such as search, use an API key:"

dichotommy (approved):
💬 🛡️ 🔑 👌🏻

maryamsulemani97:
bors merge

PR #1718, opened 2022-06-13
Update azure.md

dichotommy:
Put "Meilisearch_apikey" in backticks in lieu of adding a strange term to our vale dictionary.


maryamsulemani97 (commented):
Shouldn't we put all these fields in backticks? For consistency?

dichotommy:
@maryamsulemani97 We could. My logic was that `Meilisearch_apikey` is the only field formatted in a code-like way.

I believe that for our style guide, we agreed that button text, field names, and other things like this should be enclosed in double quotes. So we could enforce that rule here.

Test: does "`this`" look strange?

maryamsulemani97:
> Test: does "`this`" look strange?

It does, but maybe we'll get used to it. Let's keep it in backticks for now and add the double quotes part later?

dichotommy:
No need to wait, just give me 30 seconds to add the double quotes.

maryamsulemani97 (approved):
🦇

bors merge

PR #1735, opened 2022-06-16
v0.28: accepted `_geo` `lat`, `lng` types

### LINKED ISSUE #1694: v0.28: `_geo` accepts string values
Starting with v0.28, `_geo` fields (both `lat` and `lng`) should accept numbers in both numeric and string formats:

```js
"_geo": {
  "lat": 0.0, // floating point number, valid
  "lng": "0.0" // a string that can be adequately converted into a float, also valid
}
```

We need to update the Geosearch guide so this is stated clearly.

## References
Spec: https://github.com/meilisearch/specifications/pull/83
Issue on core: https://github.com/meilisearch/meilisearch/issues/2354
SME: @irevoire

guimachiavelli:
Closes #1694

### DIFF
--- a/learn/advanced/geosearch.md
+++ b/learn/advanced/geosearch.md
@@ -8,14 +8,14 @@ In order to start filtering and sorting documents based on their geographic loca
 
 `_geo` is a reserved field. If you include it in your documents, Meilisearch expects its value to conform to a specific format.
 
-When using JSON and NDJSON, `_geo` must contain an object with two keys: `lat` and `lng`. Both must contain a floating point number indicating, respectively, latitude and longitude:
+When using JSON and NDJSON, `_geo` must contain an object with two keys: `lat` and `lng`. Both fields must contain either a floating point number or a string indicating, respectively, latitude and longitude:
 
 ```json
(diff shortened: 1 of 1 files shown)


maryamsulemani97 (changes requested):
I can't comment on lines 132 and 219, but we need to update them as well

guimachiavelli:
> I can't comment on lines 132 and 219, but we need to update them as well

I initially thought the same thing, but after looking at it a bit more closely I changed my mind. Both `_geoPoint` and `_geoRadius` must be used as a value for a search parameter (`sort` and `filter`, respectively). This means both are always sent as strings, no? E.g. `{ "filter": "_geoRadius(45.472735, 9.184019, 2000)" }`.

We could argue that we shouldn't refer to these as floating point numbers, but I _think_ the intent is clear enough. If you think it's better to change it anyway, I guess we could replace "two floating  point numbers indicating" with "two values indicating".

maryamsulemani97 (approved):
🥐

maryamsulemani97:
bors merge

PR #1736, opened 2022-06-16
v0.28: Changes to `/documents`

### LINKED ISSUE #1700: v0.28: Document API changes
v0.28 introduces three main changes to `/documents`.

## Resource pagination
As the name implies, getting a list of documents now returns paginated results using the new parameters `limit` and `offset`. Response bodies include a `results` array, together with `limit`, `offset`, and `total` fields.

## Rename `attributesToRetrieve` to `fields`
The API endpoint for getting documents `attributesToRetrieve` is now called `fields`. This is to reduce confusion with the similarly named search parameter `attributesToRetrieve`.

## `displayedAttributes`
There is no longer any relationship between `displayedAttributes`, the index setting, and the `fields` returned by the documents endpoint. Previously, `attributesToRetrieve` (now `fields`) would use the `displayedAttributes` list  by default.

### References
Specs:
  - general: https://github.com/meilisearch/specifications/pull/143/files
  - pagination: https://github.com/meilisearch/specifications/pull/138/files
Issue on core: https://github.com/meilisearch/meilisearch/issues/2372
SME: @curquiza, @irevoire

guimachiavelli:
Closes #1700

### DIFF
--- a/reference/api/documents.md
+++ b/reference/api/documents.md
@@ -79,7 +79,7 @@ Documents are ordered by Meilisearch depending on the hash of their id.
   ],
   "offset": 0,
   "limit": 2,
-  "total": 50134
+  "total": 500134
 }
 ```
 
--- a/reference/api/documents.md
+++ b/reference/api/documents.md
@@ -53,23 +53,32 @@ Documents are ordered by Meilisearch depending on the hash of their id.
 #### Response: `200 Ok`
 
 ```json
-[
-  {
-    "id": 25684,
-    "release_date": "1993-01-01",
-    "poster": "https://image.tmdb.org/t/p/w1280/iuAQVI4mvjI83wnirpD8GVNRVuY.jpg",
-    "title": "American Ninja 5",
-    "overview": "When a scientists daughter is kidnapped, American Ninja, attempts to find her, but this time he teams up with a youngster he has trained in the ways of the ninja."
-  },
-  {
-    "id": 468219,
-    "title": "Dead in a Week (Or Your Money Back)",
-    "release_date": "2018-09-12",
-    "poster": "https://image.tmdb.org/t/p/w1280/f4ANVEuEaGy2oP5M0Y2P1dwxUNn.jpg",
-    "overview": "William has failed to kill himself so many times that he outsources his suicide to aging assassin Leslie. But with the contract signed and death assured within a week (or his money back), William suddenly discovers reasons to live... However Leslie is under pressure from his boss to make sure the contract is completed."
-  }
-]
+{
+  "results": [
+    {
+      "id": 25684,
+      "release_date": "1993-01-01",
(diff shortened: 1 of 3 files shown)


curquiza (changes requested):
According to [the spec](https://github.com/meilisearch/specifications/pull/143/files), the `fields` parameter is now also available for the `GET /indexes/:uid/documents/:document_uid` route.

maryamsulemani97 (commented):
Looks good! 🦋 
A few suggestions:

- Update `learn/getting_started/algolia_migration.md` line 193: `displayedAttributes` setting does not impact the displayed fields returned
- I think we should update the `get_documents_1` code sample to show how to use the `fields` parameter 
- Add to note that `fields` values are case-sensitive.

guimachiavelli:
@maryamsulemani97: thanks for the suggestions. I agree with the second and third, but unless I'm mistaken the `attributesToRetrieve` mentioned in the line you indicated refers to the search parameter `attributesToRetrieve`, not `/documents`'s `fields`.

curquiza (approved):
🤘

maryamsulemani97:
>unless I'm mistaken the attributesToRetrieve mentioned in the line you indicated refers to the search parameter attributesToRetrieve, not /documents's fields.

Two rows use `attributesToRetrieve`, lines 182 and 193. I'm assuming 182 refers to the search parameter

reference/api/documents.md
```diff
@@ -53,24 +60,33 @@ Documents are ordered by Meilisearch depending on the hash of their id.
...
+      "id": 468219,
+      "title": "Dead in a Week (Or Your Money Back)",
+      "release_date": "2018-09-12",
+      "poster": "https://image.tmdb.org/t/p/w1280/f4ANVEuEaGy2oP5M0Y2P1dwxUNn.jpg",
+      "overview": "William has failed to kill himself so many times that he outsources his suicide to aging assassin Leslie. But with the contract signed and death assured within a week (or his money back), William suddenly discovers reasons to live... However Leslie is under pressure from his boss to make sure the contract is completed."
+    }
+  ],
+  "offset": 0,
+  "limit": 2,
+  "total": 50134
```
guimachiavelli, line 82:
```suggestion
  "total": 500134
```

guimachiavelli:
@maryamsulemani97: hmmmm…I'm not sure why it's there two times—could be an error, or that Algolia does not distinguish between things we do (and thus need two separate names). In any case, the line introducing the table reads (emphasis mine):

> The below table compares Algolia's API parameters with the equivalent **Meilisearch setting or search parameter**.

It explicitly states that this concerns settings and search parameters while `fields` is only concerned with the documents route. The main goal of this update seems to be to further distance `/documents` from `/search`— to the point that `fields` is now completely independent from `displayedAttributes`.

So I still think that, since the table is about index settings and search parameters, it should not be affected by changes to the /documents route.

maryamsulemani97 (approved):
🥷

PR #1739, opened 2022-06-16
Add link about opening an issue

curquiza:
It will be easier for the users to be redirected to the right place


maryamsulemani97 (approved):
Good idea, thanks @curquiza! 🙏 

bors merge

PR #1759, opened 2022-06-27
v0.28: Pagination settings

### LINKED ISSUE #1722: v0.28: New pagination setting
With v0.27, Meilisearch introduced a hardcoded limit restricting the amount of search results a query can return to 1000.

In v0.28, this harcoded limit will be a configurable index setting, `pagination`. `pagination` accepts a settings object with one field, ~`limitedTo`~ `maxTotalHits`, which accepts an integer and has a default value of `1000`.

Users may update the full settings object or use the direct API route `indexes/:index/settings/pagination` with `GET`, `PATCH`, and `DELETE` endpoints.

When implementing this change in the docs, we should review the hardcoded limits from the "Known limitations" page. When doing so, we should take into account that the current text can be confusing and [might not properly explain how the search results limit interacts with the `limit` and `offset` search parameters](https://github.com/meilisearch/meilisearch/issues/2547#issuecomment-1165847096).

### Reference
- spec: https://github.com/meilisearch/specifications/pull/157
- issue on core: https://github.com/meilisearch/meilisearch/issues/2495
- SME: @Kerollmops

guimachiavelli:
Closes #1722

### DIFF
--- a/reference/api/pagination.md
+++ b/reference/api/pagination.md
@@ -6,7 +6,6 @@ This route allows you to configure the pagination settings for an index.
 
 Pagination settings can also be updated directly through the [global settings route](/reference/api/settings.md#update-settings) along with the other settings.
 
-
 To learn more about paginating search results with Meilisearch, refer to our [dedicated guide](/learn/advanced/pagination.md).
 
 ::: warning
@@ -47,9 +46,12 @@ Partially update the pagination settings for an index. The index [`uid`](/learn/
 
 #### `maxTotalHits`
 
-An integer indicating the maximum number of search results Meilisearch can return.
+**Type:** integer
+**Default value:** `1000`
 
-`maxTotalHits` takes priority over search parameters such as `limit` and `offset`. For example, if you set `maxTotalHits` to 100, you will not be able to access search results beyond 100 no matter the value configured for `offset`.
+An integer indicating the maximum number of search results Meilisearch can return. `maxTotalHits` takes priority over search parameters such as `limit` and `offset`.
+
+For example, if you set `maxTotalHits` to 100, you will not be able to access search results beyond 100 no matter the value configured for `offset`.
 
 #### Example
 
--- a/reference/api/pagination.md
+++ b/reference/api/pagination.md
@@ -1,0 +1,90 @@
+# Pagination
+
+_Child route of the [settings route](/reference/api/settings.md)._
+
+This route allows you to configure the pagination settings for an index.
+
+Pagination settings can also be updated directly through the [global settings route](/reference/api/settings.md#update-settings) along with the other settings.
+
+To learn more about paginating search results with Meilisearch,, refer to our [dedicated guide](/learn/advanced/pagination.md).
+
+::: warning
+Updating the settings means overwriting the default settings of Meilisearch. You can reset to default values using the `DELETE` routes.
+:::
+
+## Get pagination settings
+
+<RouteHighlighter method="GET" route="/indexes/{index_uid}/settings/pagination"/>
+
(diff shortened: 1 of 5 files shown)


reference/api/pagination.md
```diff
@@ -0,0 +1,94 @@
...
+
+#### Example
+
+<CodeSamples id="update_pagination_settings_1" />
+
+#### Response: `200 OK`
+
+```json
+{
+    "uid": 1,
```
maryamsulemani97, line 64:
```suggestion
    "taskUid": 1,
```

reference/api/pagination.md
```diff
@@ -0,0 +1,94 @@
...
+
+#### Example
+
+<CodeSamples id="reset_pagination_settings_1" />
+
+#### Response: `200 OK`
+
+```json
+{
+    "uid": 1,
```
maryamsulemani97, line 86:
```suggestion
    "taskUid": 1,
```

reference/api/pagination.md
```diff
@@ -0,0 +1,94 @@
...
+<RouteHighlighter method="PATCH" route="/indexes/{index_uid}/settings/pagination"/>
+
+Partially update the pagination settings for an index. The index [`uid`](/learn/core_concepts/indexes.md#index-uid) is required.
+
+### Body
+
+#### `maxTotalHits`
+
+**Type:** integer
+**Default value:** `1000`
```
maryamsulemani97, line 50: Do we need to mention that 1000 is the recommended value? [Spec (lines 31-33)](https://github.com/meilisearch/specifications/pull/157/files#diff-ec4edaf515880ccb83c711f72245fb90498bea850b4256a25d0d215c8550fe31):
>The value of 1000 ensures good performance and prevents malicious users from scraping documents from a Meilisearch instance.
>Increasing this value can degrade performance as well as expose the data of an instance to scrapping.

maryamsulemani97 (changes requested):

reference/api/pagination.md
guimachiavelli: Good question. I think I would rather us not say that it is recommended (there are many legitimate reasons for changing this value), but mentioning it may have an effect on performance sounds reasonable.

reference/api/pagination.md
```diff
@@ -0,0 +1,98 @@
...
+{
+    "taskUid": 1,
+    "indexUid": "books",
+    "status": "enqueued",
+    "type": "settingsUpdate",
+    "enqueuedAt": "2022-04-14T20:56:44.991039Z"
+}
+```
+
+You can use the returned `uid` to get more details on [the status of the task](/reference/api/tasks.md#get-task).
```
maryamsulemani97, line 76:
```suggestion
You can use the returned `taskUid` to get more details on [the status of the task](/reference/api/tasks.md#get-task).
```

reference/api/pagination.md
```diff
@@ -0,0 +1,98 @@
...
+{
+    "taskUid": 1,
+    "indexUid": "books",
+    "status": "enqueued",
+    "type": "settingsUpdate",
+    "enqueuedAt": "2022-04-14T20:53:32.863107Z"
+}
+```
+
+You can use the returned `uid` to get more details on [the status of the task](/reference/api/tasks.md#get-task).
```
maryamsulemani97, line 98:
```suggestion
You can use the returned `taskUid` to get more details on [the status of the task](/reference/api/tasks.md#get-task).
```

maryamsulemani97 (approved):
Just a minor change but all good 👾

guimachiavelli:
PR approved by @meilisearch/docs-team, let's see wait to see what core has to say before merging.

Kerollmops (approved):
LGTM, thank you!

PR #1734, opened 2022-06-16
v0.28: /search name changes

maryamsulemani97:
closes https://github.com/meilisearch/documentation/issues/1692

PR #1761, opened 2022-06-27
v0.28: Faceting settings

### LINKED ISSUE #1704: v0.28: New faceting index setting
The new `facet` search parameter (previously `facetsDistribution`) will now return a maximum of 100 facets per query by default. The main reasons for this are usability—a query returning a huge number of facets (e.g. 1000) can be difficult to handle properly—and performance—a query returning thousands of facets is likely to be much heavier than it otherwise should.

v0.28 will have a new index setting, `faceting`. `faceting` will contain a settings object with one field, `maxValuesPerFacet`, which accepts an integer and defaults to `100`.

`faceting` will be availabe in the main settings object and its own API route with `GET`, `PATCH` and `DELETE` endpoints.

### References
spec: https://github.com/meilisearch/specifications/pull/157
issue on core: https://github.com/meilisearch/meilisearch/issues/2368
SME: @Kerollmops

guimachiavelli:
Closes #1704


reference/api/settings.md
```diff
@@ -30,17 +30,18 @@ Get the settings of an index. The index [`uid`](/learn/core_concepts/indexes.md#
...
-| **sortableAttributes**    | [Strings] | Attributes to use when [sorting](/learn/advanced/sorting.md) search results  | `[]`                                                                         |
-| **stopWords**             | [Strings] | List of words ignored by Meilisearch when present in search queries              | `[]`                                                                                              |
-| **synonyms**              | Object    | List of associated words treated similarly                                       | `{}`                                                                                              |
-| **typoTolerance**         | Object    | Typo tolerance settings                                                          | `{}`                                                                                              |
+| Variable                 | Type      | Description                                                                                | Default value                                                                                |
```
maryamsulemani97, line 33: I can't comment on line 9, but you should add:
`- [Faceting](/reference/api/faceting.md)`

reference/api/faceting.md
```diff
@@ -0,0 +1,94 @@
...
+<RouteHighlighter method="PATCH" route="/indexes/{index_uid}/settings/faceting"/>
+
+Partially update the faceting settings for an index. The index [`uid`](/learn/core_concepts/indexes.md#index-uid) is required.
+
+### Body
+
+#### `maxValuesPerFacet`
+
+**Type:** integer
+**Default value:** `100`
```
maryamsulemani97, line 50: Should we mention that 100 is the recommended value? From the [spec (lines 31-33)](https://github.com/meilisearch/specifications/blob/fb8b067925afda0ec51855df87a165a50222da5b/text/157-faceting-setting-api.md#311-maxvaluesperfacet)

>The value of 100 ensures good performance and prevents malicious users from scraping data from a Meilisearch instance.
>Increasing this value can degrade performance as well as expose the data of an instance to scrapping.

maryamsulemani97 (changes requested):

guimachiavelli:
Not sure why github actions is complaining. The problematic link is legit. It should be safe to ignore the error.

reference/api/faceting.md
```diff
@@ -0,0 +1,98 @@
...
+
+Configure the maximum number of facet values returned for each facet. Values are alphabetically sorted.
+
+For example, suppose a query's search results contain a total of three values for a `colors` facet: `blue`, `green`, and `red`. If you set `maxValuesPerFacet` to `2`, Meilisearch will only return `blue` and `green` in the response body's `facetDistribution` object.
+
+::: note
+Settings `maxValuesPerFacet` to a high value might negatively impact performance.
```
maryamsulemani97, line 57:
```suggestion
Setting `maxValuesPerFacet` to a high value might negatively impact performance.
```

reference/api/faceting.md
```diff
@@ -0,0 +1,98 @@
...
+{
+  "taskUid": 1,
+  "indexUid": "books",
+  "status": "enqueued",
+  "type": "settingsUpdate",
+  "enqueuedAt": "2022-04-14T20:56:44.991039Z"
+}
+```
+
+You can use the returned `uid` to get more details on [the status of the task](/reference/api/tasks.md#get-task).
```
maryamsulemani97, line 76:
```suggestion
You can use the returned `taskUid` to get more details on [the status of the task](/reference/api/tasks.md#get-task).
```

reference/api/faceting.md
```diff
@@ -0,0 +1,98 @@
...
+{
+  "taskUid": 1,
+  "indexUid": "books",
+  "status": "enqueued",
+  "type": "settingsUpdate",
+  "enqueuedAt": "2022-04-14T20:53:32.863107Z"
+}
+```
+
+You can use the returned `uid` to get more details on [the status of the task](/reference/api/tasks.md#get-task).
```
maryamsulemani97, line 98:
```suggestion
You can use the returned `taskUid` to get more details on [the status of the task](/reference/api/tasks.md#get-task).
```

maryamsulemani97 (commented):
We don't mention that the faceting settings are configurable in `learn/advanced/filtering_and_faceted_search.md` or anywhere. Maybe we could add something under  "Using facets" or work on this after v0.28?

maryamsulemani97:
>Not sure why github actions is complaining. The problematic link is legit. It should be safe to ignore the error.

That's because `/reference/api/tasks.md#get-task` was replaced with `/reference/api/tasks.md#get-one-task` in https://github.com/meilisearch/documentation/pull/1746

guimachiavelli:
But that's on the base v0.28 branch, not on this one, right? Or do our actions check against the base PR branch?

learn/advanced/filtering_and_faceted_search.md
```diff
@@ -316,6 +316,10 @@ Using the `facetsDistribution` search parameter adds two new keys to the returne
 `exhaustiveFacetsCount` is not currently implemented in and will always return `false`.
 :::
 
+::: note
+By default, `facetDistribution` returns a maximum of 100 facet values for each faceted field. You can change this value using the `maxValuesPerFacet` property of the [`faceting` index settings](/reference/api/faceting.md).
```
maryamsulemani97, line 320:
```suggestion
By default, `facets` returns a maximum of 100 facet values for each faceted field. You can change this value using the `maxValuesPerFacet` property of the [`faceting` index settings](/reference/api/faceting.md).
```
`facetsDistribution` is now called [`facets`](https://github.com/meilisearch/documentation/pull/1734)

maryamsulemani97 (approved):
Looks good, just one minor change 🧛‍♀️

learn/advanced/filtering_and_faceted_search.md
guimachiavelli: hmmmm…that's interesting. In my head I was referring to the response value, which is now `facetDistribution`, but I think you're right, it makes more sense to refer to the request parameter.

reference/api/faceting.md
```diff
@@ -0,0 +1,98 @@
...
+Partially update the faceting settings for an index. The index [`uid`](/learn/core_concepts/indexes.md#index-uid) is required.
+
+### Body
+
+#### `maxValuesPerFacet`
+
+**Type:** integer
+**Default value:** `100`
+
+Configure the maximum number of facet values returned for each facet. Values are alphabetically sorted.
```
Kerollmops, line 52: I am not sure that we should specify that the results are returned in this order, as it is not a specified behavior, that's just how the engine works right now and not how the specs want it.

Kerollmops (changes requested):

reference/api/faceting.md
guimachiavelli: I think the spec might have been recently updated: https://github.com/meilisearch/specifications/pull/157/files#diff-6738a194cb97b82d2f0051b59a87cb03a343c5bad665ee11c2da0faf00a273c8R35

Unless I'm misunderstanding what we mean by lexicographical order on L#35?

Kerollmops: We mean that if there were capital letters, for example, the capital one would be before the lowercase versions. It really applies to the raw values encoded as UTF-8. The real order is the order in terms of byte representation and not determined by human things like alphabetic or numeric order. For example `2` will be listed before `1000`.

gmourier: Indeed, the actual behavior has been updated in the specification because I had internal questions about it.

I think it's not bad to explain it, it also allows people to know about it and to tell us if there are frustrations about it so we can improve it later. (my 2 cents)

gmourier, line 52:
```suggestion
Configure the maximum number of facet values returned for each facet. Values are sorted in ascending lexicographical order.
```

gmourier (changes requested):
@guimachiavelli If you decide to explain how facets values are sorted. I made a little suggestion to make it clearer.

Other than that everything seems clear to me ✨

reference/api/faceting.md
guimachiavelli: Thanks for the clarification. I was hoping we could have gotten away with alphabetical order since it's a bit more user-friendly, but, indeed, it sacrifices too much technical accuracy. 

I have accepted @gmourier's suggestion changing "alphabetical" to "ascending lexicographical" as I also think some users might appreciate knowing the expected behaviour.

Thanks to both of you for the assistance!

gmourier (approved):
LGTM!

Kerollmops (approved):
LGTM thank you very much!

guimachiavelli:
bors merge

PR #1785, opened 2022-07-21
Hotfix: global settings code samples

guimachiavelli:
This PR fixes two issues with code samples:

1. Code samples updating the global settings object in https://docs.meilisearch.com/learn/configuration/settings.html were using PUT instead of PATCH
2. The code sample updating the global settings object in https://docs.meilisearch.com/reference/api/settings.html#update-settings contained malformed data

Since these are errors specific to cURL, this PR should require no intervention by the integrations team.


maryamsulemani97 (changes requested):
We'll need to update `faceted_search_update_settings_1` line 631 to use PATCH as well

maryamsulemani97 (approved):

guimachiavelli:
bors merge

Issue #1711, opened 2022-06-09
Vale next steps


maryamsulemani97:
https://github.com/meilisearch/documentation/pull/1646 adds basic rules to the docs. We will be updating these rules and adding new ones over time. The full list of current+possible rules can be found in [Notion](https://www.notion.so/meilisearch/Vale-84ca4f880e8f40a1b4946d494afb9291). 

The next step is to fix the following current rules:

- [x] [FirstPerson.yml](https://github.com/meilisearch/documentation/pull/1646/files#diff-1caa38c79d4a695e087c17f61e9a19d38d72ddc9371c34dd112628c54c0028e2)

  - Possible solution: 
       [Update the message](https://github.com/vultr/vultr-mdtk/blob/84e2213b281380f45c176b873cdf01cf24152623/styles/Vultr/FirstPerson.yml) to something like: `Avoid first-person pronouns such as '%s' unless writing FAQs"`
  - Figure out how to ignore this rule for certain pages like [FAQs](https://docs.meilisearch.com/faq.html), [Telemetry](https://docs.meilisearch.com/learn/what_is_meilisearch/telemetry.html), etc.
- [x]  [HeadingPunctuation.yml](https://github.com/meilisearch/documentation/pull/1646/files#diff-a56955066196e52adbd9e9f7d9b163c16bc8890d55c9ebdf537917f6e3d78da1) and [Headings.yml](https://github.com/meilisearch/documentation/pull/1646/files#diff-c67024c30495a15ac4562a42cbd10e53ffc12a5c47204d83faab02ff9c9b424b)
   - Both rules don't work for headings like `1. Click this` and `1.2. Click here`. #1776  

Add new rules:
- [ ]  ~~No capitals after colons in sentences~~
  - Too many exceptions 
- [x]  Decide max sentence length (most rules use 25-30) 
    - Most of our “longer” (more than 25 words) sentences are between 27-30 words. The longest are ~42. We can start with a max sentence length of ~40 words and see how that goes. #1777  
 - [ ]  Sound less condescending. See discussion here: https://github.com/meilisearch/documentation/pull/1662

Start working on:
- [ ]  a list of [words to replace](https://github.com/errata-ai/Microsoft/blob/master/Microsoft/Wordiness.yml) to avoid wordy sentences (e.g., replace `in order to` with `to`)
[...]

PR #1786, opened 2022-07-25
add GitHub Action for Vale

maryamsulemani97:
Part of #1711 

Adds Vale checks to the CI. Copied from the [vale-action](https://github.com/errata-ai/vale-action) repository


.github/workflows/vale.yml
```diff
@@ -0,0 +1,18 @@
+name: reviewdog
+on: push
```
bidoubiwa, line 2:
```suggestion
on:
  pull_request:
  push:
```

I'm not sure without `pull_request` that it trigger a test whenever you create the PR for the first time

bidoubiwa (changes requested):

.github/workflows/vale.yml
maryamsulemani97: You're right, thanks! 🙏

.github/workflows/vale.yml
```diff
@@ -0,0 +1,20 @@
+name: reviewdog
+on:
+  pull_request:
+  push:
+
```
bidoubiwa, line 5:
```suggestion
on:
  pull_request:
  push:
         # trying and staging branches are for BORS config
        branches:
          - main
```

It appears it runs the tests both times since my change. I looked into the other repo's and that is how we do it :)

bidoubiwa (approved):
LGTM 🔥🔥

.github/workflows/vale.yml
```diff
@@ -0,0 +1,22 @@
+name: reviewdog
+on:
+  pull_request:
+  push:
+    # trying and staging branches are for BORS config
```
bidoubiwa, line 5:
```suggestion
```
my bad!

bidoubiwa (approved):
LGTM 🔥

.github/workflows/vale.yml
```diff
@@ -0,0 +1,21 @@
+name: reviewdog
+on:
+  pull_request:
+  push:
+    branches:
+      - main
```
curquiza, line 6: I'm pretty sure you don't have any `main` branch but a `master` one 😄 
However I'm not sure these 3 lines are useful: see my global comment of the review

curquiza (changes requested):
Is it something you want as a requirement before merging to `master`?
Because in this case
- you might want to add this check in the `bors.toml` to avoid bors merging your PR if the vale check is not passed
- the 3 lines I commented on are not useful: if you cannot merge a PR without passing this check, no need to run it again on `main`/`master`
- you already have a GHA manifest running on every PR and working with Bors, you might want to update [this file](https://github.com/meilisearch/documentation/blob/master/.github/workflows/check-links.yml) instead of creating a new one

bors.toml
```diff
@@ -1,6 +1,7 @@
 status = [
   'check-links',
-  'check-style'
+  'check-style',
+  'vale'
```
bidoubiwa, line 4: I know this is not the point of this PR but you could add the builds in the bors toml as well. 

```
check-builds (14)
check-builds (16)
```

maryamsulemani97: Does it have to be 
```
check-builds (14)
check-builds (16)
```

Or will just `check-builds` work too?

bidoubiwa: It has to be as its the names of the jobs



```
check-builds (14)
check-builds (16)
```

bidoubiwa (approved):
LGTM 🔥

dichotommy (approved):
LGTM! Let's give it a shot 💥

maryamsulemani97:
bors merge

maryamsulemani97:
bors merge

PR #1787, opened 2022-07-25
API reference: put all query parameters in ``

maryamsulemani97:
Part of #1351

This PR updates the API reference to put all query parameters in backticks


guimachiavelli (commented):
Do we need both **bold** and `code` formatting? What if we dropped bold? Would query parameters become too subtle?

maryamsulemani97:
>Would query parameters become too subtle?

I think so

Screen Shot 2022-08-02 at 20 21 16

reference/api/search.md
```diff
@@ -27,22 +27,22 @@ This is the preferred route to perform search when an API key is required, as it
...
-| **[attributesToCrop](#attributes-to-crop)**           | Attributes whose values have to be cropped         | `null`        |
-| **[cropLength](#crop-length)**                        | Maximum length of cropped value in words           | `10`          |
-| **[cropMarker](#crop-marker)**                        | String marking crop boundaries                     | `"…"`         |
-| **[attributesToHighlight](#attributes-to-highlight)** | Highlight matching terms contained in an attribute | `null`        |
-| **[highlightPreTag](#highlight-tags)**                | String inserted at the start of a highlighted term | `"<em>"`      |
-| **[highlightPostTag](#highlight-tags)**               | String inserted at the end of a highlighted term   | `"</em>"`     |
-| **[showMatchesPosition](#show-matches-position)**     | Return matching terms location                     | `false`       |
-| **[sort](#sort)**                                     | Sort search results by an attribute's value        | `null`        |
+| Query Parameter                                         | Description                                        | Default Value |
```
dichotommy, line 30: Just wanted to point out that these are not "Query parameters", as they are not part of the query string. They are part of the JSON payload.

Some ideas for how we could rename this column: "Body parameter", "Body property", "Payload parameter", "Payload property"... or drop the prefix and simply name the column "Property" or "Parameter". WDYT @guimachiavelli @maryamsulemani97 ?

maryamsulemani97: I think just "parameter" works. We refer to these everywhere as "the `filter` search parameter" or "the `filter` parameter"

guimachiavelli: Ackchyually…they are query parameters. You can make requests such as

`curl   -X GET 'http://localhost:7700/indexes/movies/search?q=batman&attributesToCrop=overview'`

And receive an adequate response. It's of course unpractical to say the least, especially when you have complex values like arrays and objects, but I think they do classify as query parameters.

Unless I'm missing something?

guimachiavelli:
I tend to prefer using only the minimum level of distinction required to establish hierarchy, so personally I find bold and code too much. That said, I am overruled by the majority ☺️ 

In any case, perhaps something to get a designer's opinion when we migrate to the new site and etc?

reference/api/search.md
dichotommy: Well, we are both right. The code sample you posted works, but only because it is using the GET endpoint.

With the POST endpoint, the so-called search parameters are part of the payload/body. However since GET does not accept a payload, they have to be query parameters when using GET /indexes/{index_uid}/search.

dichotommy: My point was that since this table is describing the POST endpoint, we shouldn't call the column "Query parameters", we should call it "Body", "Parameter", "Field" or something similar.

guimachiavelli: hmmm…although technically correct for `POST`, I'm not convinced differentiating them is helpful for users.

I'm fine with an unqualified "Parameter", though, which would translate to both POST and GET looking the same. Perhaps we could that for all query parameters tables across the reference?

Either way, this discussion is out of scope for this PR, right? I'm going to give my approval so we can merge this and we can continue talking in a dedicated PR.

guimachiavelli (dismissed):
🖇️

guimachiavelli (approved):
🍏

maryamsulemani97:
bors merge

PR #1788, opened 2022-07-25
API reference: add settings objects page

maryamsulemani97:


maryamsulemani97:
closing this as we will be going with #1795

Issue #1273, opened 2021-11-10
Review asynchronous updates page


dichotommy:
This page has been the victim of a lot of content bloat over the last year and needs to be reviewed. Should be made more brief, focused, clear.

Note that this feature is changing in v0.25, so this may happen as part of that release.

maryamsulemani97:
Rename [Asynchronous operations](https://docs.meilisearch.com/learn/advanced/asynchronous_operations.html) to include the word “tasks”, e.g. “Tasks and asynchronous processes”.

maryamsulemani97:
We could move the “[Summarized task object](https://docs.meilisearch.com/learn/advanced/asynchronous_operations.html#summarized-task-objects)” table to the API reference overview page. Under the “[Response body](https://docs.meilisearch.com/reference/api/overview.html#response-body)” section

PR #1789, opened 2022-07-25
API reference: move task filtering and pagination to async guide

maryamsulemani97:
Part of #1351 

- Moved task filtering and pagination to the [async operations guide](https://docs.meilisearch.com/learn/advanced/asynchronous_operations.html)
  - I think compared to the rest of the page, these headings are too technical? But that's something we can fix when we [rewrite the whole page ](https://github.com/meilisearch/documentation/issues/1273)
- Added `sidebarDepth: 2` so all subheadings are visible, especially under `Understanding tasks`

### DIFF
--- a/learn/advanced/asynchronous_operations.md
+++ b/learn/advanced/asynchronous_operations.md
@@ -1,3 +1,8 @@
+---
+
+sidebarDepth: 2
+
+---
 # Asynchronous operations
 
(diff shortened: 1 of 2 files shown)


guimachiavelli (approved):
🐤

maryamsulemani97:
bors merge

PR #1791, opened 2022-07-28
add object to keys and tasks

maryamsulemani97:
Part of #1351 

This PR adds an "Object" section to the tasks and keys API reference

## For reviewers

- `keys.md`
   -  We explain all the fields in the "Keys" object heading and then in the "Create key" endpoint. Is it a good idea to have this in two places?
   - I used an asterisk next to the required fields in the "Create key" endpoint. Is it visible?
 - `tasks.md`
   - the tables in the "Details" section have different widths


maryamsulemani97:
closing this for #1796

PR #1796, opened 2022-08-02
Api refrence update tasks and keys

maryamsulemani97:
Part of #1351
This PR updates the tasks and keys API reference to add an object heading and removes the tasks table from async tasks

## For reviewers

- `keys.md`
   -  We explain all the fields in the "Keys" object heading and then in the "Create key" endpoint. Is it a good idea to have this in two places?
   - I used an asterisk next to the required fields in the "Create key" endpoint. Is it visible?
 - `tasks.md`
   - the tables in the "Details" section have different widths


reference/api/tasks.md
```diff
@@ -6,6 +6,152 @@ The `/tasks` route gives information about the progress of [asynchronous operati
...
+
+**Type**: String
+
+**Description**: The date and time when the task was first `enqueued`, in RFC 3339 format
+
+### `startedAt`
+
+**Type**: String
+
+**Description**: The date and time when the task began processing, in RFC 3339 format
```
dichotommy, line 147:
```suggestion
**Description**: The date and time when the task began `processing`, in RFC 3339 format
```

reference/api/tasks.md
```diff
@@ -6,6 +6,152 @@ The `/tasks` route gives information about the progress of [asynchronous operati
...
+
+### `finishedAt`
+
+**Type**: String
+
+**Description**: The date and time when the task finished processing, whether failed or succeeded, in RFC 3339 format
```
dichotommy, line 153:
```suggestion
**Description**: The date and time when the task finished processing, whether `failed` or `succeeded`, in RFC 3339 format
```

reference/api/tasks.md
```diff
@@ -14,19 +160,19 @@ List all tasks globally, regardless of index. The `task` objects are contained i
 
 Tasks are always returned in descending order of `uid`. This means that by default, **the most recently created `task` objects appear first**.
 
-Task results are paginated and can be filtered. To learn more, refer to our [asynchronous operations](/learn/advanced/asynchronous_operations.md#filtering-tasks) guide.
+Task results are [paginated](#paginating-tasks) and can be [filtered](#filtering-tasks).
```
dichotommy, line 163: Branch needs to be updated w/ the latest from master

reference/api/tasks.md
```diff
@@ -14,19 +160,19 @@ List all tasks globally, regardless of index. The `task` objects are contained i
...
+| Query Parameter | Description                                                          |         Default Value          |
+|-----------------|----------------------------------------------------------------------|:------------------------------:|
+| **limit**       | number of tasks to return                                            |               20               |
+| **from**        | `uid` of the first task returned                                     | `uid` of the last created task |
+| **status**      | [filter tasks](#filtering-tasks) by their `status`                   |          all statuses          |
+| **type**        | [filter tasks](#filtering-tasks) by their `type`                     |           all types            |
+| **indexUid**    | [filter tasks](#filtering-tasks) by their `indexUid`. Case-sensitive |          all indexes           |
 
-### Example
+#### Example
```
dichotommy, line 175: Can we define when example should be an H3 and when it should be an H4?

H4 in the settings API reference, H3 everywhere else?

reference/api/error_codes.md
```diff
@@ -183,8 +183,8 @@ The given `name` is invalid. It should either be a string or `null`.
 
 ### `invalid_task_status`
 
-The requested task status is invalid. Please use one of [these four possible values](/learn/advanced/asynchronous_operations.md#task-status).
+The requested task status is invalid. Please use one of [these four possible values](/reference/api/tasks.md#status).
 
 ### `invalid_task_type`
 
-The requested task type is invalid. Please use one of the values defined in our [asynchronous operations guide](/learn/advanced/asynchronous_operations.md#task-api-response).
+The requested task type is invalid. Please use one of the values defined in our [asynchronous operations guide](/reference/api/tasks.md#type).
```
dichotommy, line 190:
```suggestion
The requested task type is invalid. Please use one of the [possible values](/reference/api/tasks.md#type).
```

reference/api/error_codes.md
```diff
@@ -183,8 +183,8 @@ The given `name` is invalid. It should either be a string or `null`.
 
 ### `invalid_task_status`
 
-The requested task status is invalid. Please use one of [these four possible values](/learn/advanced/asynchronous_operations.md#task-status).
+The requested task status is invalid. Please use one of [these four possible values](/reference/api/tasks.md#status).
```
dichotommy, line 186: Future-proofing: if you don't use an exact number you won't have to change it later when possible values are added or removed.

```suggestion
The requested task status is invalid. Please use one of the [possible values](/reference/api/tasks.md#status).
```

reference/api/keys.md
```diff
@@ -4,22 +4,143 @@ The `/keys` route allows you to create, manage, and delete API keys. To use thes
...
+Since `key` is a hash of the `uid` and master key, `key` values are deterministic between instances sharing the same configuration. This means if the master key changes, all `key` values are automatically changed.
+
+Since the `key` field depends on the master key, it is computed at runtime and therefore not propagated to dumps and snapshots. As a result, even if a malicious user comes into possession of your dumps or snapshots, they will not have access to your instance's API keys.
+:::
+
+### `actions`
+
+**Type**: Array
+
+**Description**: An array of API actions permitted for the key. `["*"]` for all actions.
```
dichotommy, line 71: So, I'm not sure of the best way to put this, but I see we have a repeated warning a few times in the below table:

> "**Non-authorized `indexes` will be omitted from the response**"

I'd like to express this in some form in the description so we could remove it from the table and reduce the overall size/clutter a bit.

```suggestion
**Description**: An array of API actions permitted for the key, represented as strings. API actions are only possible on authorized [`indexes`](#indexes). `["*"]` for all actions.
```

reference/api/keys.md
```diff
@@ -4,22 +4,143 @@ The `/keys` route allows you to create, manage, and delete API keys. To use thes
...
+| `settings.update`  | Provides access to the [update settings](/reference/api/settings.md#update-settings) and [reset settings](/reference/api/settings.md#reset-settings) endpoints and equivalents for all subroutes on authorized indexes |
+| `stats.get`        | Provides access to the [get stats of an index](/reference/api/stats.md#get-stats-of-an-index) endpoint and the [get stats of all indexes](/reference/api/stats.md#get-stats-of-all-indexes) endpoint. For the latter, **non-authorized `indexes` are omitted from the response** |
+| `dumps.create`     | Provides access to the [create dump](/reference/api/dump.md#create-a-dump) endpoint. **Not restricted by `indexes`** |
+| `version`          | Provides access to the [get Meilisearch version](/reference/api/version.md#get-version-of-meilisearch) endpoint |
+| `keys.get`         | Provides access to the [get all keys](#get-all-keys) endpoint |
+| `keys.create`      | Provides access to the [create key](#create-a-key) endpoint |
+| `keys.update`      | Provides access to the [update key](#update-a-key) endpoint |
+| `keys.delete`      | Provides access to the [delete key](#delete-a-key) endpoint |
```
dichotommy, line 92: If my above suggestion is accepted, I think we can get rid of "on authorized indexes" etc.. `dumps.create` is an exception!

reference/api/keys.md
```diff
@@ -4,22 +4,143 @@ The `/keys` route allows you to create, manage, and delete API keys. To use thes
...
+
+**Default value**: None
+
+### `indexes`
+
+**Type**: Array
+
+**Description**: An array of indexes the key is authorized to act on. `["*"]` for all indexes. Only the key's [permitted actions](#actions-2) can be used on these indexes
```
dichotommy, line 100:
```suggestion
**Description**: An array of indexes the key is authorized to act on. `["*"]` for all indexes. Only the key's [permitted actions](#actions) can be used on these indexes
```

reference/api/keys.md
```diff
@@ -4,22 +4,143 @@ The `/keys` route allows you to create, manage, and delete API keys. To use thes
...
+
+**Description**: Date and time when the key will expire, represented in RFC 3339 format. `null` if the key never expires
+
+**Default value**: None
+
+### `createdAt`
+
+**Type**: String
+
+**Description**: Date and time when the key was created, represented in RFC 3339 format.
```
dichotommy, line 116:
```suggestion
**Description**: Date and time when the key was created, represented in RFC 3339 format
```

reference/api/keys.md
```diff
@@ -4,22 +4,143 @@ The `/keys` route allows you to create, manage, and delete API keys. To use thes
...
+**Description**: Date and time when the key was last updated, represented in RFC 3339 format
+
+**Default value**: `null`
+
 ## Get all keys
 
 <RouteHighlighter method="GET" route="/keys"/>
 
-List the 20 most recently created keys. **Expired keys are included in the response**, but deleted keys are not. Results can be paginated by using the `offset` and `limit` query parameters.
+Returns a `results` array that lists the 20 most recently created keys. **Expired keys are included in the response**, but deleted keys are not. `total` shows the number of API keys that can be browsed.
```
dichotommy, line 132:
```suggestion
Returns the 20 most recently created keys in a `results` array. **Expired keys are included in the response**, but deleted keys are not. The field `total` attached to the response indicates the total number of API keys in existence.
```

reference/api/keys.md
```diff
@@ -4,22 +4,143 @@ The `/keys` route allows you to create, manage, and delete API keys. To use thes
...
 
-[See below for an explanation of returned fields.](#description)
+### Query parameters
 
-#### Query parameters
+Results can be paginated by using the `offset` and `limit` query parameters.
```
dichotommy, line 136:
```suggestion
Results can be paginated using the `offset` and `limit` query parameters.
```

reference/api/keys.md
```diff
@@ -182,80 +237,84 @@ Get information on the specified key. Attempting to use this endpoint with a non
...
 
 ## Create a key
 
 <RouteHighlighter method="POST" route="/keys"/>
 
 Create an API key with the provided description, permissions, and expiration date.
 
-Only the `indexes`, `actions`, and `expiresAt` fields are mandatory.
-
 ### Body
```
dichotommy, line 248: I think we should do a table for the body, with links to the extended object definition where necessary. Just seems cleaner, less redundant, and more consistent w/ what we're doing elsewhere. WDYT?

dichotommy (changes requested):
Once some comments are addressed, this looks pretty good to me.

It's not super critical, but at some point I would like us to discuss and clarify the "Default value" field. For example, if the default is a randomly or semi-randomly generated value, should we state that? What about if the default value is an empty array/object? I think in these two cases we're saying `None` or `N/A` right now but I'm not super satisfied with those options, and don't know when I would use one vs. the other.

I think we should try using CSS to force tables to fill the `content-width`, as we discussed in our meeting. I find the tables of inconsistent width to be distracting. Can be done in a separate PR.

reference/api/tasks.md
```diff
@@ -6,6 +6,152 @@ The `/tasks` route gives information about the progress of [asynchronous operati
...
+|--------------|------------------------------------------------------------------------------------------------------------|
+| `primaryKey` | Value of the `primaryKey` field . `null` if no `primaryKey` is specified at the time of the index update   |
+
+#### `indexDeletion`
+
+| Name               | Description                                                                           |
+|--------------------|---------------------------------------------------------------------------------------|
+| `deletedDocuments` | Number of deleted documents. This should equal all the documents in the deleted index |
+
+#### `settingsUpdate`
```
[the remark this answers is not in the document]
maryamsulemani97: > I think it should be clarified that these are the updated values for the settings object, not the current ones. Unless that's obvious?

I think it's obvious, it is called `settingsUpdate`

> Also, do all settings appear here every time, or is it just the settings that will be updated?

Just the updated ones. I'll add a note for this

reference/api/tasks.md
maryamsulemani97: Wouldn't it be better to set it to H4 everywhere?

reference/api/tasks.md
```diff
@@ -2,10 +2,188 @@
...
+}
+```
+
+:::
+
+#### `documentAdditionOrUpdate`
+
+<style>
+table th:first-of-type {
+    width: 10%;
```
maryamsulemani97, line 94:
```suggestion
    width: 25%;
```

maryamsulemani97, line 94: I don't completely understand how this works. When I set it to 10%, it works for every table on the docs. If I change it, it only seems to work on a few tables in `tasks.md`

reference/api/tasks.md
guimachiavelli: That makes sense, but the actual goal should be a bit more comprehensive, imho. It's not necessarily establishing a specific h# for code samples, but creating a document structure in which that the desired the h# is the most sensible choice, if that makes sense.

reference/api/keys.md
```diff
@@ -182,80 +237,84 @@ Get information on the specified key. Attempting to use this endpoint with a non
...
 
-**Type:** string
+**Type:** String
+
 **Default value:** `null`
 
 An optional description for the key.
 
-#### `actions`
+#### `actions` *
```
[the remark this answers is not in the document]
maryamsulemani97: >Do we need to specify that the * means required?

I think most users would know as many API references use it, but it might be better to specify it. We could mention it in the overview, so we don't have to mention it on every page

reference/api/tasks.md
maryamsulemani97: I put it in backticks everywhere, it makes sense either way

reference/api/keys.md
```diff
@@ -4,22 +4,143 @@ The `/keys` route allows you to create, manage, and delete API keys. To use thes
...
+
+**Description**: A human-readable name for the key
+
+**Default value**: `null`
+
+### `description`
+
+**Type**: String
+
+**Description**: A description for the key
```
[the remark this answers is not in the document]
maryamsulemani97: I couldn't find anything for `name` but added a few words for `description`

reference/api/tasks.md
maryamsulemani97: I like when Example and Response are both H4s, they look better
Plus, if we ever use `sidebarDepth`, they won't show up in the sidebar
Screen Shot 2022-08-04 at 19 26 52

reference/api/keys.md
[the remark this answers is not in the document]
dichotommy: Let's assume that users can understand it for now, and if we get a complaint about it we can clarify at that time.

reference/api/tasks.md
dichotommy: I do not understand this either. In any case, since it appears to be affecting every table in the docs—and as far as I can tell, is doing exactly what we wanted: making every table use the full `content-width`—it would be better if it was in a more explicit place rather than this particular file. What happens if you add this to `.vuepress/styles/index.styl`, for example?

reference/api/tasks.md
dichotommy: I agree @maryamsulemani97 , they do look better as H4s. However I'm a bit concerned about parent-child relationships. With the current structure, this would make example and response children of whichever H3 came before—most likely `Query parameters`, `Body`, or similar.

One solution would be to move the example and response earlier, to immediately follow the endpoint description. That way they would be children of the endpoint H2—however we would be jumping from an H2 to an H4. Does that matter to us? 🤔

reference/api/tasks.md
maryamsulemani97: Moved to `.vuepress/styles/index.styl`. I set `width` to `10`. All tables are the same width. 

If I set it to `25`, tables with 3 columns look better, but somehow all tables have different widths. We can work on this in a seperate PR

`width: 10%;`:
Screen Shot 2022-08-09 at 15 27 01

`width: 25%;`:
Screen Shot 2022-08-09 at 15 45 16

reference/api/tasks.md
maryamsulemani97: >However I'm a bit concerned about parent-child relationships. With the current structure, this would make example and response children of whichever H3 came before—most likely Query parameters, Body, or similar.

I did not think of this. We should keep Example as an H3, I guess.

reference/api/tasks.md
guimachiavelli: I think working on a separate PR is a good idea. I think the rest of these changes are good and important, so wouldn't want to delay it because of presentation tweaks.

reference/api/keys.md
```diff
@@ -4,20 +4,132 @@ The `/keys` route allows you to create, manage, and delete API keys. To use thes
...
+
+**Type**: String
+**Default value**: N/A
+**Description**: A [uuid v4](https://www.sohamkamani.com/uuid-versions-explained) to identify the API key. If not specified, it is automatically generated by Meilisearch
+
+### `key`
+
+**Type**: String
+**Default value**: N/A
+**Description**: An alphanumeric key value generated by Meilisearch by hashing the `uid` and the master key on API key creation. Used for authorization when [making calls to a protected Meilisearch instance](/learn/security/master_api_keys.md#communicating-with-a-protected-instance).
```
guimachiavelli, line 49: I'm not sure I like removing the period, but it felt inconsistent with the majority of the other field descriptions. Alternatively we could add periods to all other descriptions, but this solution seems more efficient.

```suggestion
**Description**: An alphanumeric key value generated by Meilisearch by hashing the `uid` and the master key on API key creation. Used for authorization when [making calls to a protected Meilisearch instance](/learn/security/master_api_keys.md#communicating-with-a-protected-instance)
```

reference/api/keys.md
```diff
@@ -4,20 +4,132 @@ The `/keys` route allows you to create, manage, and delete API keys. To use thes
...
+Since `key` is a hash of the `uid` and master key, `key` values are deterministic between instances sharing the same configuration. This means if the master key changes, all `key` values are automatically changed.
+
+Since the `key` field depends on the master key, it is computed at runtime and therefore not propagated to dumps and snapshots. As a result, even if a malicious user comes into possession of your dumps or snapshots, they will not have access to your instance's API keys.
+:::
+
+### `actions`
+
+**Type**: Array
+**Default value**: N/A
+**Description**: An array of API actions permitted for the key, represented as strings. API actions are only possible on authorized [`indexes`](#indexes). `["*"]` for all actions.
```
guimachiavelli, line 63:
```suggestion
**Description**: An array of API actions permitted for the key, represented as strings. API actions are only possible on authorized [`indexes`](#indexes). `["*"]` for all actions
```

guimachiavelli (approved):
A couple minor comments, but I think this is good enough to merge and would suggest any further changes/comments to be left for another PR.

reference/api/tasks.md
```diff
@@ -2,10 +2,172 @@
...
+      }
+   },
+   "duration": "PT0.000682S",
+   "enqueuedAt": "2022-08-02T15:41:57.955977Z",
+   "startedAt": "2022-08-02T15:41:57.957002Z",
+   "finishedAt": "2022-08-02T15:41:57.957684Z"
+}
+```
+
+:::
```
dichotommy, line 83: I don't think this is true. Either way it seems to be more confusing than helpful so I think we should remove it.

```suggestion
```

reference/api/keys.md
```diff
@@ -287,12 +284,19 @@ Date and time when the key will expire, represented in RFC 3339 format. `null` i
 
 <RouteHighlighter method="PATCH" route="/keys/{key_or_uid}"/>
```
dichotommy, line 285: We are missing a body section + table for this endpoint! It should include `name` and `description`.

dichotommy (changes requested):
Final changes!

reference/api/keys.md
maryamsulemani97: I added the links for `name` and `description` in line 287, but a table would be better

dichotommy (approved):
Looks good 😎

maryamsulemani97:
bors merge

PR #1798, opened 2022-08-03
Clarify interactions between `maxTotalHits` and `offset`

### LINKED ISSUE #1784: Clarify distinction between `limit` and `maxTotalHits`
[A user reported being surprised by the effect of `maxTotalHits` on `offset`](https://github.com/meilisearch/meilisearch/issues/2547). This is something that is not clearly described in our documentation as of now.

## Summary

It's my opinion that we need to do a better job illustrating the difference between these two types of search limitation.

`limit` restricts the number of results returned by a single query. `limit` will not stop you from using `offset` to view later results.

`maxTotalHits` on the other hand restricts the total number of results fetched by Meilisearch, _even when using `offset`_. That is an explicit part of its design, since it was essentially created to stop people from scraping the database using `offset`.

As an example, let's say we make a search with `"maxTotalHits": 20`. Meilisearch will return up to 20 documents that match the search query, ordered by relevancy. However, it will _never_ return any documents ordered past this number—even using `offset`.

`maxTotalHits` is not just a limit on the number of documents returned by a search—it's a limit on how exhaustive a search can be. It means that by default, when you make a search query, Meilisearch finds up to `1000` of the most relevant documents and stops there. It does not find the 1001st most relevant document or any beyond that.

Since this seems to be a difficult subject, I would love the confirmation of a developer @meilisearch/core-team 🙏🏻 

## Where in the docs do changes need to happen?

- In [Known Limitations](https://docs.meilisearch.com/learn/advanced/known_limitations.html#maximum-number-of-results-per-search): “By default, Meilisearch returns up to 1000 documents per search”
- In [Settings guide](https://docs.meilisearch.com/learn/configuration/settings.html#pagination): “The maximum number of results Meilisearch can return”

Both of these could be improved by mentioning that an `offset` greater than `maxTotalHits` leads to an empty results array. See the previous section for a detailed explanation of `maxTotalHits`.

In [Settings guide](https://docs.meilisearch.com/learn/configuration/settings.html#pagination): "`maxTotalHits` takes priority over search parameters such as `limit` and `offset`"

This could be misinterpreted, as a query made w/ a `limit` of `50` and a `maxTotalHits` of `100` should still return up to `50` documents. In this sense, the lower limit will always be the one applied to search results. 

`maxTotalHits` _does_ take priority in the sense that it is a hard limit: for search queries, Meilisearch will only ever fetch and rank up to `maxTotalHits` documents. Attempts to retrieve documents past that using `offset` will return empty arrays.

## Wrap up

A few questions to be considered by whoever works on this:

- Are there any other spots in the documentation that should be updated?
- The most accurate description is not always the most helpful for users. In this case, what is necessary and helpful for the user to understand?

guimachiavelli:
Closes #1784


learn/advanced/known_limitations.md
```diff
@@ -99,4 +99,4 @@ user = 1 OR user = 2 […] OR user = 1500 OR user = 1501 […] OR user = 2000 OR
...
 
-**Explanation:** Meilisearch limits the maximum amount of returned search results to protect your database from malicious scraping. You may change this by using the `maxTotalHits` property of the [pagination index settings](/reference/api/pagination.md#maxtotalhits). `maxTotalHits` only applies to the [search route](/reference/api/search.md) and has no effect on the [get documents endpoint](/reference/api/documents.md#get-documents).
+**Explanation:** Meilisearch limits the maximum amount of returned search results to protect your database from malicious scraping. You may change this by using the `maxTotalHits` property of the [pagination index settings](/reference/api/pagination.md#maxtotalhits).  While the `limit` search parameter restricts the number of results returned in a single request, `maxTotalHits` restricts the total number of search results. This means a search query with an `offset` greater than the index's `maxTotalHits` returns an empty array. Note that this index setting only applies to the [search route](/reference/api/search.md) and has no effect on the [get documents endpoint](/reference/api/documents.md#get-documents).
```
dichotommy, line 102:
```suggestion
**Explanation:** Meilisearch limits the total number of returned search results to protect your database from malicious scraping. There is no way to access search results past this limit, but you _can_ change it using the `maxTotalHits` property of the [pagination index settings](/reference/api/pagination.md#maxtotalhits).
```

dichotommy (changes requested):

reference/api/search.md
```diff
@@ -309,7 +313,7 @@ If you want to skip the **first** result in a query, set `offset` to `1`:
 Sets the maximum number of documents returned by a single query.
 
 ::: tip
-This parameter is often used together with `offset` in order to paginate results.
+You can paginate search results by making queries combining both `offset` and `limit`.
 :::
```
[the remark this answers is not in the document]
guimachiavelli:
```suggestion
:::

::: warning
A search query cannot return more results than configured in [`maxTotalHits`](/reference/api/pagination.md#update-pagination-settings), even if the value of `limit` is greater than the value of `maxTotalHits`.
:::
```

reference/api/search.md
```diff
@@ -309,7 +313,11 @@ If you want to skip the **first** result in a query, set `offset` to `1`:
 Sets the maximum number of documents returned by a single query.
 
 ::: tip
-This parameter is often used together with `offset` in order to paginate results.
+You can paginate search results by making queries combining both `offset` and `limit`.
```
dichotommy, line 316:
```suggestion
You can [paginate search results](/learn/advanced/pagination.md) by making queries combining both `offset` and `limit`.
```

maryamsulemani97 (approved):
👾

PR #1800, opened 2022-08-04
The tasks response has missing information

brunoocasali:
I've suggested some data to make this PR, but feel free to change it.

Btw, I missed some information about the response. Since the task routes have a different way of pagination, it may be good to explain a little about it (sorry if this is solved somewhere else).

https://docs.meilisearch.com/reference/api/tasks.html#get-tasks


maryamsulemani97 (approved):
Thanks @brunoocasali!  

bors merge

maryamsulemani97:
>Since the task routes have a different way of pagination, it may be good to explain a little about it (sorry if this is solved somewhere else).

You can find it in the [async guide](https://docs.meilisearch.com/learn/advanced/asynchronous_operations.html#paginating-tasks)

PR #1803, opened 2022-08-08
simplify `updating.md`

### LINKED ISSUE #1723: Simplify updating guide
The [updating guide](https://docs.meilisearch.com/learn/advanced/updating.html#update-to-the-latest-meilisearch-version) currently provides two solutions:

1. For users of v0.15 or above
2. For users of v0.14 and below

This distinction is due to the addition of dumps in v0.15, which greatly simplified the process of updating a Meilisearch instance.

However, the two separate processes have added too much complexity to using and maintaining this guide. This, combined with additional factors such as changes in authorization headers and API methods, has led us to decide to drop the v0.14 and below section, and instead provide a single process that works for all users with some modifications.

Essentially, we will delete the `Updating from v0.14.0 or below` section and all of its code samples. We will also need to add some new disclaimers and information at the top of the guide:

- Guide does not work for versions under v0.15: for more information, contact support
- Authorization header: versions between v0.15 and v0.25 should use `X-Meili-Api-Key`, while versions from v0.25 onwards should use `Authorization: Bearer`
- Dump compatibility: users of v0.20 and below may have to update in two steps. See [the specification](https://specs.meilisearch.com/specifications/text/0105-dumps-api.html#_3-4-dump-version-support) for more information

maryamsulemani97:
closes #1723


learn/advanced/updating.md
```diff
@@ -1,30 +1,33 @@
 # Update to the latest Meilisearch version
 
-Currently, Meilisearch databases can only be opened by the Meilisearch version you used to create them. The following guide will walk you through all the steps to **migrate an existing database** from an older version of Meilisearch to the most recent one.
+:::warning
```
maryamsulemani97, line 3: I don't like how this looks:

Screen Shot 2022-08-08 at 16 40 05

But I'm not sure what would be a better solution

guimachiavelli: I think this should be split in two warnings. 

The first (guide  does not work on <v0.15) is in the right place. Perhaps we could move it to the end of the intro—might be more aesthetically pleasing not to have the warning right next to the page title. You could also use `::: danger` instead of `::: warning` if you want to be extra scary.

The second warning (auth header) could perhaps be leave where it was? Or perhaps just as the last line on the intro, as regular text? If you think that the problem is the code sample taking too much space, I think we could do without it.

learn/advanced/updating.md
```diff
@@ -1,30 +1,33 @@
...
+:::
+
+Currently, Meilisearch databases can only be opened by the Meilisearch version you used to create them. The following guide will walk you through all the steps to migrate an existing database from an older version of Meilisearch to the most recent one.
 
 If you already have a Meilisearch database with some data you don’t want to lose, you are in the right place!
 
 :::note
-If you have already **installed the latest version and manually indexed your data and settings**, you can ignore this guide.
+If you have already installed the latest version and manually indexed your data and settings, you can ignore this guide.
 :::
```
guimachiavelli, line 17: Do we really need this note? Isn't it essentially saying "if you have already updated meilisearch, you don't need to follow this guide on updating meilisearch"?

learn/advanced/updating.md
```diff
@@ -37,7 +40,7 @@ The response should look something like this:
 
 If you get the error `Cannot open database, expected Meilisearch engine version: 0.X.X, current engine version 0.Y.Y`, your database is not compatible with the currently installed Meilisearch version.
 
-In this case, you need to **download the compatible version now** (i.e., `0.X.X` in the above error message) so that you can access and export your database.
+In this case, you need to download the compatible version now (i.e., `0.X.X` in the above error message) to access and export your database.
```
guimachiavelli, line 43: Great edit 👍

learn/advanced/updating.md
```diff
@@ -106,47 +109,29 @@ docker run -it --rm \
...
 
 ::: note
-If your dump was created in Meilisearch v0.21 or above, continue to step 2.
+If your dump was created in Meilisearch v0.21 or above, continue to [step 3](#step-3-create-the-dump).
 :::
 
 When creating dumps, Meilisearch calls the same method as the [get documents endpoint](/reference/api/documents.md#get-documents). This means that all fields must be [displayed](/learn/configuration/displayed_searchable_attributes.md#displayed-fields) in order to be saved in the dump.
 
-Start by using the [get displayed attributes endpoint](/reference/api/displayed_attributes.md#get-displayed-attributes) to verify that **all attributes are displayed**.
+Start by using the get displayed attributes endpoint to verify that all attributes are displayed:
```
guimachiavelli, line 120: If the link is not necessary, I think we can make this sentence sound a bit more natural:

```suggestion
Start by verifying that all attributes are included in the displayed attributes list:
```

Or something to that effect. I find the `the [something] endpoint` construct a bit wooden when we're not using it to link to the reference.

learn/advanced/updating.md
```diff
@@ -106,47 +109,29 @@ docker run -it --rm \
...
 
 <CodeSamples id="updating_guide_get_displayed_attributes_new" />
 
-If the response is `{'displayedAttributes': '["*"]'}`, you can move on to the [next step](#step-2-create-the-dump).
+If the response is `{'displayedAttributes': '["*"]'}`, you can move on to the [next step](#step-3-create-the-dump).
 
-If it's something else, then you need to use the [reset displayed attributes endpoint](/reference/api/displayed_attributes.md#reset-displayed-attributes). Before doing this, make sure you save your list of displayed attributes somewhere so you can restore it afterwards.
+If it's something else, then you need to use the reset displayed attributes endpoint. Before doing this, make sure you save your list of displayed attributes somewhere so you can restore it afterwards.
```
guimachiavelli, line 126:
```suggestion
If it's something else, then you need to reset the list of displayed attributes. Before doing this, make sure you save your list of displayed attributes somewhere so you can restore it afterwards.
```

learn/advanced/updating.md
```diff
@@ -202,7 +187,7 @@ It should return something like this:
 
 ::::
 
-To create a dump, use the [create dump endpoint](/reference/api/dump.md#create-a-dump).
+To create a dump, use the create dump endpoint:
```
guimachiavelli, line 190:
```suggestion
You can then create a dump of your database:
```

guimachiavelli (commented):
Content seems good and love how much we were able to cut from this guide.

I think the spots where we removed the endpoint links could be improved so the text flows more naturally. I've made a few suggestions, feel free to edit them if you don;t like how they sound.

guimachiavelli (approved):
👨‍⚕️

maryamsulemani97:
bors merge

Issue #1593, opened 2022-04-13
Review FAQ content


dichotommy:
Like most FAQs, our FAQ is a big collection of answers that is often added to, but not often returned to by our writers.

The main task of whoever reviews this is to ensure that all content is still accurate + up to date. Sub-tasks would be to improve clarity of answers where they are unclear, move what you see as the most important answers to the top of the file, and/or to create some kind of content organization grouping answers by topic.

guimachiavelli:
FAQs are notoriously disliked by some technical writers. Although I'm not on team "remove all the FAQs"—one of our past surveys indicated that at least some people use it—I still think we should generally strive to move content away from this page whenever possible.

I say that because of two main reasons:
1. I've seen the FAQ sometimes "poisoning" our search results. Meaning, a user searches for e.g. `document id` and the first result is the FAQ instead of our Core concepts > Documents guide;
2. It's harder to keep track of content that's here. I think we do a good job most of the time, but it's always possible core  updates e.g. Error types and we only update the Errors page and not the relevant FAQ question;
3. (bonus reason/I don't know how to count)The FAQ becomes more maintainable because there's less text to review (important since the FAQ naturally requires a lot of context switching).

At a quick glance, I think these sections are prime cutting candidates: 
1. "How can I contact the Meilisearch team" — there are links to github, slack, twitter around the website. If this is a common question, there's something wrong with the design of the docs;
4. "I do not understand the relevancy of my search results." — first of all, not a question (pet peeve). Second, we have a page called Relevancy;
5. "Does Meilisearch deliver an interface to search in my documents"—we already mention the existence the search preview in other pages;
6. "What do the different error types mean"—probably better to mention this in our Errors reference

maryamsulemani97:
Looked up some [best practices](https://larryludwig.com/faq-page-best-practices/), and here are a few things we might want to keep in mind:
1. Can these FAQs be easily addressed in an existing section of our website?
2. Inventing questions for the sake of answering them creates confusion and noise
3. Keep them simple. Make sure the questions are clearly worded and the answers focus on providing a solution, not creating more questions or options.
4. Organize FAQs so users can find answers quickly and easily(if we end up with too many)


Adding to Gui's list:
5. "My document upload failed with the document id is missing error."- this looks like it was copy-pasted from the [document id](https://docs.meilisearch.com/learn/core_concepts/primary_key.html#document-id) section
[...]

PR #1810, opened 2022-08-09
remove 400 - Bad Request FAQ

maryamsulemani97:
part of #1593 

Moves content from the `400 - Bad Request` FAQ to Core concepts/Documents

### DIFF
--- a/resources/faq.md
+++ b/resources/faq.md
@@ -67,7 +67,7 @@ This response indicates that the operation has been taken into account and will 
 
 ## I am trying to add my documents but I keep receiving a `400 - Bad Request` response
 
-The `400 - Bad request` response probably means that your data is not in an expected format. You might have extraneous commas, mismatched brackets, missing quotes,etc.. Meilisearch API accepts JSON, CSV, and NDJSON formats.
+The `400 - Bad request` response probably means that your data is not in an expected format. You might have extraneous commas, mismatched brackets, missing quotes, etc. Meilisearch API accepts JSON, CSV, and NDJSON formats.
 
 In case of a [document addition](/reference/api/documents.md#add-or-replace-documents), only an array of objects is expected.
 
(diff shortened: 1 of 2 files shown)


dichotommy (commented):
I have mixed feelings about moving this content to the documents guide. On the one hand, the guide is already a bit messy so it doesn't make things significantly worse. On the other hand, this info seems mostly useful for troubleshooting; it's not something essential that every user needs to know (and it's not conceptual).

I think it would make more sense in the API reference, under the add documents endpoints, or in the errors page, under the `malformed_payload` code. I also think it could be shortened a lot: instead of providing an example, we could just say make sure that all documents are in an array and that the JSON is correctly formatted. But perhaps I'm overthinking this?

@guimachiavelli WDYT?

guimachiavelli:
I think @dichotommy's suggestion that this is probably longer than it should is relevant. That FAQ answer also focusses too much on commas, when malformed JSON occurs because of many other things (missing quotes around a key, mismatched [] and {}, etc). If we simplify the text and remove the examples and the tip, it could easily fit into either Core Concepts or the Errors page. It could even be adequate to mention it in both pages, actually, if we're ok with minor content duplication.

Out of scope, but it is interesting that you mention troubleshooting, as that is something I was reading about on _Docs for Devs_. To keep things short, the authors recommend having a troubleshooting reference with instructions on how to fix common errors. Not 100% how that would work, especially from an IA perspective, but could be interesting to discuss.

resources/faq.md
```diff
@@ -67,44 +67,9 @@ This response indicates that the operation has been taken into account and will
...
-
-Wrong:
-
-```json
-{
-  "id": 123,
-  "title": "Pride and Prejudice"
-}
-```
+The `400 - Bad request` response probably means that your data is not in an expected format. You might have extraneous commas, mismatched brackets, missing quotes,etc.. Meilisearch API accepts JSON, CSV, and NDJSON formats.
```
maryamsulemani97, line 70:
```suggestion
The `400 - Bad request` response probably means that your data is not in an expected format. You might have extraneous commas, mismatched brackets, missing quotes, etc. Meilisearch API accepts JSON, CSV, and NDJSON formats.
```

guimachiavelli (dismissed):
Two suggestions, one very minor, and we should be good to go

guimachiavelli (approved):
🐤

maryamsulemani97:
bors merge

PR #1814, opened 2022-08-10
query parameters != payload fields

dichotommy:
A small clarification, mostly because we don't want users to think these can be used as query parameters for the POST endpoint, but also for consistency.


maryamsulemani97 (approved):
🧚🏻‍♀️

dichotommy:
bors merge

PR #1826, opened 2022-08-18
API reference: update stats.md

maryamsulemani97:
part of #1351


dichotommy (changes requested):
As I understand it, the full stats object is closer to the response of the global endpoint (`Get stats of all indexes`) than that of the index endpoint (`Get stats of an index`). Basically, the index-specific endpoint is requesting a specific chunk of the larger stats object.

In line with that, I think the stats object section should include the three additional fields and index name attributes that are present when calling the global stats endpoint.

For fields whose names vary, such as the index names, we could represent them in the table as {index_uid} or something like that.

For the 3–4 fields that only appear in the global stats endpoint, we could also note this in their description.

Let me know if any of this is unclear 👍🏻

dichotommy (changes requested):
Just a few small changes, then good to go 👍🏻

dichotommy (approved):
Looks great! ⚽ ▶️ 🥅

maryamsulemani97:
bors merge

PR #1841, opened 2022-08-24
update synonym explanation

### LINKED ISSUE #1454: Update synonym explanation
The current documentation says: 

>In general, a search on a word will return the same results as a search on any of its synonyms.

This is not true, as Meilisearach takes other things like type tolerance, split queries etc. into account when searching.

For more details: https://github.com/meilisearch/meilisearch/issues/2162

maryamsulemani97:
closes #1454


.vuepress/public/sample-template.yaml
```diff
@@ -148,3 +148,4 @@ get_faceting_settings_1: |-
 update_faceting_settings_1: |-
 reset_faceting_settings_1: |-
 settings_guide_faceting_1: |-
+synonyms_guide_1: |-
```
guimachiavelli, line 151:
```suggestion
synonyms_guide_1: |-

```

learn/configuration/settings.md
```diff
@@ -234,7 +234,7 @@ Suppose you would like to search `the mask` in a movie database. Since `the` is
 
 ## Synonyms
 
-A set of words defined for an index. Synonyms are **different words that have the same meaning**, and thus are treated similarly. If either of the associated words is searched, the same results will be displayed.
+A set of words defined for an index. Synonyms are **different words that have the same meaning**, and thus are treated similarly.
```
guimachiavelli, line 237: I wonder if "treated" by itself is clear enough? I find it a bit vague without the deleted sentence, as if treated might mean loads of different and mysterious things 🤔

learn/getting_started/customizing_relevancy.md
```diff
@@ -126,7 +126,7 @@ The `synonyms` array lets you associate certain words in your dataset, telling M
 
 <CodeSamples id= "getting_started_synonyms" />
 
-This will set `winnie` and `piglet` as synonyms; searching for either word will show the same results.
+This will set `winnie` and `piglet` as synonyms. Searching for either won't always return the same results due to factors like allowed typos and splitting the word.
```
guimachiavelli, line 129: I think this addition is great, but requires a bit more work.

1. What do we mean with "allowed typos"?  Perhaps just "typos" is better because users are not lead to wonder about allowed vs disallowed typos?

2. I don't understand what we mean by "splitting the word." Can you clarify it?

learn/core_concepts/indexes.md
```diff
@@ -81,7 +81,7 @@ Designate the distinct attribute using the [update settings endpoint](/reference
 
 ### Synonyms
 
-Your dataset may contain words with similar meanings. For these, you can define a list of synonyms: words that will be treated as the same or similar for search purposes.
+Your dataset may contain words with similar meanings. For these, you can define a list of synonyms: words that will be treated as the same or similar for search purposes. Words set as synonyms won't always return the same results due to factors like allowed typos and splitting the word.
```
guimachiavelli, line 84: same comments as the ones on `customizing_relevancy.md`

.code-samples.meilisearch.yaml
```diff
@@ -965,3 +965,10 @@ settings_guide_faceting_1: |-
     --data-binary '{
       "maxValuesPerFacet": 5
     }'
+synonyms_guide_1: |-
+  curl \
+    -X PUT 'http://localhost:7700/indexes/movies/settings/synonyms' \
+    -H 'Content-Type: application/json' \
+    --data-binary '{
+      "great": ["fantastic"], "fantastic": ["great"]
+      }'
```
guimachiavelli, line 974:
```suggestion
  curl \
    -X PUT 'http://localhost:7700/indexes/movies/settings/synonyms' \
    -H 'Content-Type: application/json' \
    --data-binary '{
      "great": ["fantastic"], "fantastic": ["great"]
    }'
```

learn/configuration/synonyms.md
```diff
@@ -2,8 +2,15 @@
 
 If multiple words have an equivalent meaning in your dataset, you can [create a list of synonyms](/reference/api/synonyms.md#update-synonyms). This will make your search results more relevant.
 
-In general, **a search on a word will return the same results as a search on any of its synonyms**.
-There is one exception to this rule, [detailed below](#multi-word-phrases).
+Words set as synonyms won't always return the same results. With the default settings, the `movies` dataset should return 547 results for `great` and 66 for `fantastic`. Let's set them as synonyms:
+
+<CodeSamples id="synonyms_guide_1" />
+
+With the new settings, searching for `great` returns 595 results and `fantastic` returns 423 results. This is due to various factors like [allowed typos](/learn/configuration/typo_tolerance.md#minwordsizefortypos) and [splitting the word](/learn/advanced/concat.md#split-queries) to find relevant documents. The search for `great` will allow only one typo and take into account all variations of `great`, e.g., `creatures`, `greatest`, etc. along with `fantastic`.
```
guimachiavelli, line 9: A few comments on this:

1. I would remove "allowed" from "allowed typos"—it's not wrong, but I think it adds potential for confusion. Like, as a reader I immediately start wondering about the differences between allowed and disallowed typos;
2. "splitting the word" sounds really vague to me—words are a bit of a funny concept in Meilisearch, especially when taking languages like Japanese into account. I recommend "split queries" instead, which seems clearer and more technically correct;
3. On the last sentence we say "take into account all variations of `great`, e.g., `creatures`, `greatest`, etc"—is `creatures` supposed to be there? Perhaps "greater" was the intended word?

```suggestion
With the new settings, searching for `great` returns 595 results and `fantastic` returns 423 results. This is due to various factors like [allowed typos](/learn/configuration/typo_tolerance.md#minwordsizefortypos) and [splitting the word](/learn/advanced/concat.md#split-queries) to find relevant documents. The search for `great` will allow only one typo and take into account all variations of `great`, e.g., `creatures`, `greatest`, etc. along with `fantastic`.
```

guimachiavelli (changes requested):

learn/configuration/synonyms.md
maryamsulemani97: 3. Yep, it's ` <em>creat</em>ures`. Should I use ` <em>creat</em>ures` instead of `creatures`? Would that make it clearer?

learn/configuration/settings.md
maryamsulemani97: You're right. I got rid of the sentence because `the same results will be displayed` isn't always the case. I updated it to `get some of the same results`.

learn/configuration/synonyms.md
guimachiavelli: Ah, yeah, that makes sense. What confused me is that we say:

a. "The search for `great` will allow only one typo"
b. "and take into account all variations of `great`, e.g., `creatures`, `greatest`, etc. along with `fantastic`."

So we have two cases, a and b, but only one example. I read the example as illustrating only the nearest situation, that is, variations of great.

GH marked this discussion as outdated, so I'll make a new comment with a suggestion where the examples are closer to each situation they're illustrating.

learn/configuration/synonyms.md
```diff
@@ -2,8 +2,15 @@
 
 If multiple words have an equivalent meaning in your dataset, you can [create a list of synonyms](/reference/api/synonyms.md#update-synonyms). This will make your search results more relevant.
 
-In general, **a search on a word will return the same results as a search on any of its synonyms**.
-There is one exception to this rule, [detailed below](#multi-word-phrases).
+Words set as synonyms won't always return the same results. With the default settings, the `movies` dataset should return 547 results for `great` and 66 for `fantastic`. Let's set them as synonyms:
+
+<CodeSamples id="synonyms_guide_1" />
+
+With the new settings, searching for `great` returns 595 results and `fantastic` returns 423 results. This is due to various factors like [typos](/learn/configuration/typo_tolerance.md#minwordsizefortypos) and [splitting the query](/learn/advanced/concat.md#split-queries) to find relevant documents. The search for `great` will allow only one typo and take into account all variations of `great`, e.g., `creatures`, `greatest`, etc. along with `fantastic`.
```
guimachiavelli, line 9: As I mentioned in my previous comment, here's my suggestion: we separate the examples of typos and word variations so it's clear what we're referring to. I also shortened `creatures` to `create` hoping that less letters might make it easier for users to identify the typo.

```suggestion
With the new settings, searching for `great` returns 595 results and `fantastic` returns 423 results. This is due to various factors like [typos](/learn/configuration/typo_tolerance.md#minwordsizefortypos) and [splitting the query](/learn/advanced/concat.md#split-queries) to find relevant documents. The search for `great` will allow only one typo (e.g. `create`) and take into account all variations of `great` (e.g. `greatest`) along with `fantastic`.
```

We could also wrap the matching term with `<em>`, if you think it's necessary.

guimachiavelli (changes requested):
Almost there, just one final suggestion for the `creatures` example.

guimachiavelli (approved):
💎

maryamsulemani97:
bors merge

PR #1842, opened 2022-08-24
Move front-end code-samples to relevant directories

bidoubiwa:
The maintain of the front-end code-samples should be handled by the @meilisearch/integration-team. SInce we are already maintaining the code-samples present on the landing page, it seems not that much of a hastle to also maintain the samples present in the docs.

It also is way easier for us and the doc team to keep these samples up to date as they are often forgotten. 

⚠️ The tab preference is reset! This means users will have to click on their favorite language again in order to store it. I can add a work-around if you think it is important

Needs the following pr's to be merged: 

https://github.com/meilisearch/instant-meilisearch/pull/826
https://github.com/meilisearch/meilisearch-vue/pull/116
https://github.com/meilisearch/meilisearch-react/pull/32


guimachiavelli (dismissed):
A bit hard to test in the preview/locally, but as far as I see from the code and the linked PRs from other repos, LGTM!

Two minor comments that make no difference, but because this is the docs repo we feel a moral obligation to point out typos. I'm approving it anyway because variable names don't need to be grammatically correct, but in case you're interested:

1. `prefered` should have two `r`s: `preferred`
2. as far as my dictionaries go, `cacheable` is preferred over `cachable`

Thanks! 💕

guimachiavelli (approved):
💎

guimachiavelli:
bors merge

PR #1865, opened 2022-09-06
v0.29: add new search parameter `matchingStrategy`

### LINKED ISSUE #1840: v0.29: Add new search parameter `matchingStrategy`
v0.29 introduces a new search parameter: `matchingStrategy` that defines which strategy to use to match query terms within the documents as search results:
- `last`: documents containing all the query words (i.e., in the `q` parameter) are returned first by Meilisearch
- `all`: Only the documents containing all the query words (i.e., in the `q` parameter) are returned by Meilisearch

Update [Search API reference](https://docs.meilisearch.com/reference/api/search.html)

## References:
- Spec: https://github.com/meilisearch/specifications/pull/173
- Issue on core: https://github.com/meilisearch/meilisearch/issues/2578
- SME: @ManyTheFish

maryamsulemani97:
closes #1840


reference/api/search.md
```diff
@@ -735,3 +738,27 @@ Queries using `_geoPoint` will always include a `geoDistance` field containing t
...
 
 [You can read more about location-based sorting in our dedicated guide.](/learn/advanced/geosearch.md#sorting-results-with-geopoint)
+
+### Matching strategy
+
+**Parameter**: `matchingStrategy`
+**Expected value**: `last` or `all`
+**Default value**: `last`
+
+Defines the strategy used to match query terms in documents. `last` returns documents containing all the query words first. If there aren't enough documents to meet the requested `limit`, it will remove one query term at a time, starting from the second last query term.
```
maryamsulemani97, line 748:
```suggestion
Defines the strategy used to match query terms in documents. `last` returns documents containing all the query words first. If there aren't enough documents to meet the requested `limit`, it will remove one query term at a time, starting from the end of the query term.
```

reference/api/search.md
```diff
@@ -735,3 +738,27 @@ Queries using `_geoPoint` will always include a `geoDistance` field containing t
...
 
 [You can read more about location-based sorting in our dedicated guide.](/learn/advanced/geosearch.md#sorting-results-with-geopoint)
+
+### Matching strategy
+
+**Parameter**: `matchingStrategy`
+**Expected value**: `last` or `all`
+**Default value**: `last`
+
+Defines the strategy used to match query terms in documents. `last` returns documents containing all the query words first. If there aren't enough documents to meet the requested `limit`, it will remove one query term at a time, starting from the end of the query term.
```
maryamsulemani97, line 748:
```suggestion
Defines the strategy used to match query terms in documents. `last` returns documents containing all the query words first. If there aren't enough documents to meet the requested `limit`, it will remove one query term at a time, starting from the end of the query.
```

ManyTheFish (approved):

gmourier (commented):
LGTM @maryamsulemani97, thanks!

~~The specification also mentions [telemetry changes](https://github.com/meilisearch/specifications/pull/173/files#diff-dbac052211a8ea4b2c5d068a6264380740b022efead552b4dfc9e4e8d961f0f4). Do you manage everything related to the telemetry changes for `v0.29` in another PR?~~

Edit: Yes, I found it. 🧱

gmourier (approved):

reference/api/search.md
```diff
@@ -735,3 +738,27 @@ Queries using `_geoPoint` will always include a `geoDistance` field containing t
...
+
+Searching for `big fat liar` will only return documents that contain all three words.
+
+<CodeSamples id="search_parameter_guide_matching_strategy_2" />
+
+The above code sample only returns one result compared to 476 when the `matchingStrategy` was set to `last`.
+
+::: warning
+The number of search results may vary depending on changes to the `movies` dataset.
+:::
```
guimachiavelli, line 764: I think we should remove mentions of specific numbers here instead of relying on the warning to dispel any confusion.

For example, we could replace this paragraph and the warning with a short text explaining that `all` will usually return fewer results than `last`.

guimachiavelli (changes requested):

guimachiavelli:
Forgot to add this to my review.

I personally prefer using "terms" instead of "words" when talking about the words in a query. It seems a bit safer because a query term might be a number, a code, an acronym, etc (e.g. `1984`, `1Q84`, `TLDR`) and because what counts as a word can be tricky in some languages.

.code-samples.meilisearch.yaml
```diff
@@ -986,3 +986,18 @@ getting_started_pagination: |-
     --data-binary '{
       "maxTotalHits": 500
     }'
+search_parameter_guide_matching_strategy_1: |
+  curl \
+    -X POST 'http://localhost:7700/indexes/movies/search' \
+    -H 'Content-Type: application/json' \
+    --data-binary '{
+      "q": "big fat liar"
```
dichotommy, line 994: Even if it's the default, I think we should still set the matching strategy explicitly to make it more clear what the example is demonstrating.

```suggestion
      "q": "big fat liar",
      "matchingStrategy": "last"
```

reference/api/search.md
```diff
@@ -735,3 +738,21 @@ Queries using `_geoPoint` will always include a `geoDistance` field containing t
...
+**Expected value**: `last` or `all`
+**Default value**: `last`
+
+Defines the strategy used to match query terms in documents. `last` returns documents containing all the query words first. If there aren't enough documents to meet the requested `limit`, it will remove one query term at a time, starting from the end of the query.
+
+<CodeSamples id="search_parameter_guide_matching_strategy_1" />
+
+With the above code sample, Meilisearch will first return documents that contain all three words. If the results don't meet the requested `limit`, it will return documents with just `big fat` followed by just `big`.
+
+`all`, on the other hand, only returns documents that contain all query terms. Meilisearch will not match any more documents even if there aren't enough to meet the requested `limit`. Searching for `big fat liar` will only return documents that contain all three words.
```
dichotommy, line 754: We want to avoid ambiguous constructions like "on the other hand" in our API reference. We don't know what the reader may have read before this: he or she could be speed-reading, or could have followed a link directly to this paragraph. Either way they would be missing the necessary context to understand what we mean by "on the other hand". (Not to mention that we may not always have only two values for this parameter)

```suggestion
#### `all`

`all` only returns documents that contain all query terms. Meilisearch will not match any more documents even if there aren't enough to meet the requested `limit`. 
```

reference/api/search.md
```diff
@@ -735,3 +738,21 @@ Queries using `_geoPoint` will always include a `geoDistance` field containing t
...
+
+<CodeSamples id="search_parameter_guide_matching_strategy_2" />
+
+`all` usually returns fewer search results compared to `last`.
```
dichotommy, line 758: Makes more sense to give this sentence after the code sample.

```suggestion
The above code sample would only return documents containing all three words.
```

dichotommy (commented):
I recommended a couple changes since things can always be improved, but this looks good to go to me 👍🏻

reference/api/search.md
gmourier: Thanks for the tips Tommy!

guimachiavelli (approved):
🦝

PR #1866, opened 2022-09-06
v0.29: Soft deletion

### LINKED ISSUE #1852: v0.29: Soft deletion warning
With v0.29, Meilisearch will start soft deleting database records. In short, soft deletion means marking documents as deleted instead of deleting them outright. Meilisearch will wipe the data completely, sooner or later, based on DB size and remaining storage space. This is a common optimization technique that seems to giving us significant performance gains.

Soft-deleted documents are hidden from end-users and inaccessible by usual means (e.g. API endpoints), but they still remain in the database for some time. This needs to be communicated clearly so experienced developers are aware that unwanted data might linger around for some time—something particularly relevant for those serving users under GDPR and etc.

## Tasks
- [ ] Create short new section in `/learn/advanced/storage.md`
- [ ] Check for other pages that might benefit from a link to `storage.md`s new section

## References
- issue: https://github.com/meilisearch/meilisearch/issues/2583
- SME: @irevoire

guimachiavelli:
Closes #1852

### DIFF
--- a/learn/advanced/storage.md
+++ b/learn/advanced/storage.md
@@ -74,6 +74,6 @@ Database size is affected by a large number of criteria, including settings, rel
 
 ## Soft deletion
 
-Meilisearch renders deleted documents inaccessible to all users but does not immediately remove them from the database. This is a common optimization technique called "soft deletion". Soft deleted documents are permanently deleted during a later update, depending on your index size and the available disk space. It might be important to check how soft deletion interacts with data retention legislation relevant to your application.
+Meilisearch renders deleted documents inaccessible to all users but does not immediately remove them from the database. This is a common optimization technique called soft deletion. Soft deleted documents are permanently deleted during a later update, depending on your index size and the available disk space. It might be important to check how soft deletion interacts with data retention legislation relevant to your application.
 
 Soft deletion also affects document updates: when you update a document, Meilisearch removes the current record and creates a new document with updated data.
--- a/learn/advanced/storage.md
+++ b/learn/advanced/storage.md
@@ -74,8 +74,6 @@ Database size is affected by a large number of criteria, including settings, rel
 
 ## Soft deletion
 
-Meilisearch renders deleted documents inaccessible to all users but does not immediately remove them from the database. This is a common optimization technique called "soft deletion". Soft deleted documents are permanently deleted at a later date—the precise time depends on index size and available disk space.
+Meilisearch renders deleted documents inaccessible to all users but does not immediately remove them from the database. This is a common optimization technique called "soft deletion". Soft deleted documents are permanently deleted at a later date—the precise time depends on index size and available disk space. It might be important to check how soft deletion interacts with relevant data retention legislation.
 
 When updating a document, Meilisearch deletes the current record and creates a new one containing the new data. This means soft deletion, as well as its associated performance increase, also affect document updates.
-
-If you are developing an application that handles sensitive data, it might be important to check how soft deletion interacts with relevant data retention legislation.
--- a/learn/advanced/storage.md
+++ b/learn/advanced/storage.md
@@ -76,4 +76,4 @@ Database size is affected by a large number of criteria, including settings, rel
 
 Meilisearch renders deleted documents inaccessible to all users but does not immediately remove them from the database. This is a common optimization technique called "soft deletion". Soft deleted documents are permanently deleted at a later date—the precise time depends on index size and available disk space. It might be important to check how soft deletion interacts with relevant data retention legislation.
 
-When updating a document, Meilisearch deletes the current record and creates a new one containing the new data. This means soft deletion, as well as its associated performance increase, also affect document updates.
+Soft deletion also affects document updates: when you update a document, Meilisearch removes the current record and creates a new document with updated data.
--- a/learn/advanced/storage.md
+++ b/learn/advanced/storage.md
@@ -74,6 +74,6 @@ Database size is affected by a large number of criteria, including settings, rel
 
 ## Soft deletion
 
-Meilisearch renders deleted documents inaccessible to all users but does not immediately remove them from the database. This is a common optimization technique called "soft deletion". Soft deleted documents are permanently deleted at a later date—the precise time depends on index size and available disk space. It might be important to check how soft deletion interacts with relevant data retention legislation.
+Meilisearch renders deleted documents inaccessible to all users but does not immediately remove them from the database. This is a common optimization technique called "soft deletion". Soft deleted documents are permanently deleted during a later update, depending on your index size and the available disk space. It might be important to check how soft deletion interacts with data retention legislation relevant to your application.
 
 Soft deletion also affects document updates: when you update a document, Meilisearch removes the current record and creates a new document with updated data.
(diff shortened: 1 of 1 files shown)


learn/advanced/storage.md
```diff
@@ -71,3 +71,11 @@ These metrics are highly dependent on the machine that is running Meilisearch. R
...
 
 Database size is affected by a large number of criteria, including settings, relevancy rules, use of facets, the number of different languages present, and more.
+
+## Soft deletion
+
+Meilisearch renders deleted documents inaccessible to all users but does not immediately remove them from the database. This is a common optimization technique called "soft deletion". Soft deleted documents are permanently deleted at a later date—the precise time depends on index size and available disk space.
+
+When updating a document, Meilisearch deletes the current record and creates a new one containing the new data. This means soft deletion, as well as its associated performance increase, also affect document updates.
+
+If you are developing an application that handles sensitive data, it might be important to check how soft deletion interacts with relevant data retention legislation.
```
curquiza, line 81: I am afraid this sentence is afraid the users because every single user thinks they are working on sensitive data 😂 
But I don't know how we can better saying this...
@irevoire maybe?

guimachiavelli: Yeah, I think it's particularly difficult because it's not like we can say "if this is a problem, you can do [something]". At the same time, I think it's a bit dishonest not to mention it? 

Unless my sources are wrong and soft deletion poses no risk for people who need to follow GDPR and etc, in which case I'll delete this straightaway.

curquiza: I know Elastic is doing soft deletion with what they call "segments", maybe we can check on their documentation

Also, you have to know a LOT of programs are doing it also, like gmail. When you delete a mail, even from the trash, it's not immediately deleted from the machine, for optimization.

What we do is not really soft delete, but a temporary soft delete, it's not easy if it's GDPR compliant or not 🫠

guimachiavelli: > What we do is not really soft delete, but a temporary soft delete, it's not easy to know if it's GDPR compliant or not 

Yeah, makes sense. So far, everything I read about soft deletion and GDPR gives me the impression that no one really knows and/or that it might technically violate GDPR, but because the data is very hard to access it's unlikely to become a problem.

I get your point about every user thinking they're handling sensitive data, though. I'll rework it so the information is still there, but it's a bit less scary.

learn/advanced/storage.md
```diff
@@ -71,3 +71,9 @@ These metrics are highly dependent on the machine that is running Meilisearch. R
 It is important to note that **there is no reliable way to predict the final size of a database**. This is true for just about any search engine on the market—we're just the only ones saying it out loud.
 
 Database size is affected by a large number of criteria, including settings, relevancy rules, use of facets, the number of different languages present, and more.
+
+## Soft deletion
+
+Meilisearch renders deleted documents inaccessible to all users but does not immediately remove them from the database. This is a common optimization technique called "soft deletion". Soft deleted documents are permanently deleted at a later date—the precise time depends on index size and available disk space. It might be important to check how soft deletion interacts with relevant data retention legislation.
+
+When updating a document, Meilisearch deletes the current record and creates a new one containing the new data. This means soft deletion, as well as its associated performance increase, also affect document updates.
```
maryamsulemani97, line 79: I don't get the last sentence.

curquiza (approved):
LGTM 🤘

learn/advanced/storage.md
```diff
@@ -71,3 +71,9 @@ These metrics are highly dependent on the machine that is running Meilisearch. R
 It is important to note that **there is no reliable way to predict the final size of a database**. This is true for just about any search engine on the market—we're just the only ones saying it out loud.
 
 Database size is affected by a large number of criteria, including settings, relevancy rules, use of facets, the number of different languages present, and more.
+
+## Soft deletion
+
+Meilisearch renders deleted documents inaccessible to all users but does not immediately remove them from the database. This is a common optimization technique called "soft deletion". Soft deleted documents are permanently deleted at a later date—the precise time depends on index size and available disk space. It might be important to check how soft deletion interacts with relevant data retention legislation.
```
irevoire, line 77:
```suggestion
Meilisearch renders deleted documents inaccessible to all users but does not immediately remove them from the database. This is a common optimization technique called "soft deletion". Soft deleted documents are permanently deleted during a later update—the precise time depends on index size and available disk space. It might be important to check how soft deletion interacts with relevant data retention legislation.
```

Meilisearch won't delete your documents depending on how long they stayed in the database, the available processing power, or anything else.
The only time it checks if it needs to delete your soft-deleted documents is at the end of an update. So maybe this slight change in wording would avoid us some questions?

irevoire, line 77:
```suggestion
Meilisearch renders deleted documents inaccessible to all users but does not immediately remove them from the database. This is a common optimization technique called "soft deletion". Soft deleted documents are permanently deleted at a later date—the precise time depends on index size and available disk space.
```

I can see why it could interact with data retention legislation, but I find it strange to say it here while we don't talk about it at all when we explain the;
- a snapshot that stores the whole database in a big file
- a dump that stores the whole database on disk
- tasks that are slow to execute. The document deletions are not prioritized and thus may take time to execute
- Or a mix of both; document deletions that weren't executed and then got dumped/snapshotted will still contain all the data obviously.

But I'm not an expert at all in this domain.
As a developer, I think I would prefer seeing this information on another page about « legal information » to something instead of in the explanation of a feature :thinking: 

What do you think?

irevoire: Yeah sure :+1:

guimachiavelli: Ah, wow, I did not know that at all. Your wording is definitely clearer in that respect, thanks—I'm making a couple of minor edits to your suggestion for better text flow and accepting it.

learn/advanced/storage.md
```diff
@@ -71,3 +71,9 @@ These metrics are highly dependent on the machine that is running Meilisearch. R
 It is important to note that **there is no reliable way to predict the final size of a database**. This is true for just about any search engine on the market—we're just the only ones saying it out loud.
 
 Database size is affected by a large number of criteria, including settings, relevancy rules, use of facets, the number of different languages present, and more.
+
+## Soft deletion
+
+Meilisearch renders deleted documents inaccessible to all users but does not immediately remove them from the database. This is a common optimization technique called "soft deletion". Soft deleted documents are permanently deleted during a later update, depending on your index size and the available disk space. It might be important to check how soft deletion interacts with data retention legislation relevant to your application.
```
maryamsulemani97, line 77: Should soft deletion be in quotes?

guimachiavelli: It was intentional, but I'm not 100% convinced. The idea is to bring attention to "soft deletion" as a title instead of an action (like, "Meilisearch soft deletes documents" vs "the name of this technique is 'soft deletion'"). In essence, what grammarly defines as ["using words as words"[(https://www.grammarly.com/blog/quotation-marks/).

What do you think?

maryamsulemani97: We currently use quotes when referring to button text, field names, etc. Might be a longer discussion, but we could use italics instead

guimachiavelli: Hmmm…not sure I like italics. I'll just leave it without for now.

maryamsulemani97 (approved):
🫡

guimachiavelli:
bors merge

PR #1869, opened 2022-09-07
v0.29: `filter` updates

### LINKED ISSUE #1853: v0.29: Updates to the `filter` search parameter
v0.29 will expand functionality of `filter`:

- New `IN` operator;
- New `EXIST` operator;
- Updated `NOT` operator behavior;
- Filter values identical to filter operators must be enclosed in quotation marks (`x = 1 AND y = 0` vs `x = "AND"`).

## Tasks
- [ ] update `/reference/api/search.md`
- [ ] update `/learn/advanced/filtering_and_faceted_search.md`
- [ ] update `learn/getting_started/filtering_and_sorting.md`
- [ ] search for other articles mentioning and update them if necessary

## References
Issue: https://github.com/meilisearch/meilisearch/issues/2580
SMEs: @loiclec, @curquiza

guimachiavelli:
Closes #1853.

### DIFF
--- a/reference/api/search.md
+++ b/reference/api/search.md
@@ -346,7 +346,7 @@ If you want your query to return only **two** documents, set `limit` to `2`:
 
 Uses filter expressions to refine search results. Attributes used as filter criteria must be added to the [`filterableAttributes` list](/reference/api/settings.md#filterable-attributes).
 
-[Read our guide on filtering, faceted search and filter expressions.](/learn/advanced/filtering_and_faceted_search.md)
+For more information on how to use filters and build filter expressions, [read our guide on filtering, faceted search and filter expressions](/learn/advanced/filtering_and_faceted_search.md).
 
 #### Example
 
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -1,6 +1,6 @@
 const ogprefix = 'og: http://ogp.me/ns#'
 module.exports = {
-  title: 'Meilisearch Documentation v0.28',
+  title: 'Meilisearch Documentation v0.29',
   description: 'Open source Instant Search Engine',
   theme: 'default-prefers-color-scheme',
   themeConfig: {


learn/advanced/filtering_and_faceted_search.md
```diff
@@ -102,10 +102,14 @@ The [`GET` route of the search endpoint](/reference/api/search.md#search-in-an-i
 
 String expressions combine conditions using the following filter operators and parentheses:
 
-- `NOT` only returns documents that do not satisfy a condition : `NOT genres = horror`
+- `NOT` returns all documents that do not satisfy a condition, including results missing the specified attribute. The expression `NOT genres = horror` returns all documents whose `genre` is not `horror` and documents missing a `genres` field
```
curquiza, line 105:
```suggestion
- `NOT` returns all documents that do not satisfy a condition, including results missing the specified attribute. The expression `NOT genres = horror` returns all documents whose `genres` is not `horror` and documents missing a `genres` field
```

learn/advanced/filtering_and_faceted_search.md
```diff
@@ -102,10 +102,14 @@ The [`GET` route of the search endpoint](/reference/api/search.md#search-in-an-i
...
 - `AND` operates by connecting two conditions and only returns documents that satisfy both of them: `genres = horror AND director = 'Jordan Peele'`
 - `OR` connects two conditions and returns results that satisfy at least one of them: `genres = horror OR genres = comedy`
 - `TO` is equivalent to `>= AND <=`. The expression `release_date 795484800 TO 972129600` translates to `release_date >= 795484800 AND release_date <= 972129600`
+- `IN [valueA, valueB]` selects all documents whose chosen field contains at least one of the specified values. The expression `genres IN [horror, comedy]` returns all documents whose `genres` includes either `horror`, `comedy`, or both
+- `EXIST` checks for the existence of a field. Fields with empty or null values still count as existing. The expression `release_date NOT EXIST` returns all documents without a `release_date`
```
curquiza, line 110:
```suggestion
- `EXISTS` checks for the existence of a field. Fields with empty or null values still count as existing. The expression `release_date NOT EXISTS` returns all documents without a `release_date`
```

curquiza: I also realized "not exists" is not correct in English. But "XXX exist" is not correct as well

curquiza (changes requested):

curquiza (approved):
👏

learn/advanced/filtering_and_faceted_search.md
```diff
@@ -102,10 +102,14 @@ The [`GET` route of the search endpoint](/reference/api/search.md#search-in-an-i
 
 String expressions combine conditions using the following filter operators and parentheses:
 
-- `NOT` only returns documents that do not satisfy a condition : `NOT genres = horror`
+- `NOT` returns all documents that do not satisfy a condition, including results missing the specified attribute. The expression `NOT genres = horror` returns all documents whose `genres` do not contain `horror` and all documents missing a `genres` field
 - `AND` operates by connecting two conditions and only returns documents that satisfy both of them: `genres = horror AND director = 'Jordan Peele'`
 - `OR` connects two conditions and returns results that satisfy at least one of them: `genres = horror OR genres = comedy`
 - `TO` is equivalent to `>= AND <=`. The expression `release_date 795484800 TO 972129600` translates to `release_date >= 795484800 AND release_date <= 972129600`
+- `IN [valueA, valueB]` selects all documents whose chosen field contains at least one of the specified values. The expression `genres IN [horror, comedy]` returns all documents whose `genres` includes either `horror`, `comedy`, or both
```
loiclec, line 109: Nice and concise :) I would try to make it explicit that IN[..] can contain 0 or more values inside the brackets, and not strictly 2. 

If 0 values are given to it, it selects 0 documents. But I don't know whether that information should be in the documentation, it might unnecessarily confuse some people I guess.

The syntax `value NOT IN [..]` is also not explained. That is fine by me since they can use `NOT value IN [..]` instead, which is equivalent. But I just wanted to remind you of it in case omitting it was not a conscious decision :-)

learn/advanced/filtering_and_faceted_search.md
```diff
@@ -102,10 +102,14 @@ The [`GET` route of the search endpoint](/reference/api/search.md#search-in-an-i
 
 String expressions combine conditions using the following filter operators and parentheses:
 
-- `NOT` only returns documents that do not satisfy a condition : `NOT genres = horror`
+- `NOT` returns all documents that do not satisfy a condition, including results missing the specified attribute. The expression `NOT genres = horror` returns all documents whose `genres` do not contain `horror` and all documents missing a `genres` field
```
loiclec, line 105: This is not strictly correct. To take a convoluted example:
```
NOT x NOT EXISTS
```
does not return "results missing the specified attribute". In fact, it only returns results with the specified attribute.

The example you chose with `NOT genres = horror` is very instructive though and I would keep it.

loiclec (commented):
Thanks :-) I had some minor nitpicks but I think it is well explained!

learn/advanced/filtering_and_faceted_search.md
guimachiavelli: Ah, good catch. I'll add a hint indicating `IN` accepts more than two values.

As for your other two comments, indeed, it was conscious decision. We tend to prioritise information that is useful in the majority of cases over exhaustiveness—being exhaustive for the sake of completeness leads to people not reading the full thing because "ain't nobody got time for that." 

Not documenting both `NOT value IN` and `value NOT IN` additionally follows my personal preference of avoiding giving users options that I cannot also present clear reasons to choose one over the other.

learn/advanced/filtering_and_faceted_search.md
guimachiavelli: hmmmm…yeah, I see what you mean. The example you gave does sound like a _very_ edgy edge case, but the problematic clause is not strictly necessary since the next sentence already describes what we need (that `NOT x = y` will return documents with `x = z` and documents without an `x` field). I'll remove it.

loiclec (approved):
perfect 👍 thanks :-)

reference/api/search.md
```diff
@@ -346,7 +346,7 @@ If you want your query to return only **two** documents, set `limit` to `2`:
 
 Uses filter expressions to refine search results. Attributes used as filter criteria must be added to the [`filterableAttributes` list](/reference/api/settings.md#filterable-attributes).
 
-[Read our guide on filtering, faceted search and filter expressions.](/learn/advanced/filtering_and_faceted_search.md)
+For more information on how to use filters and build filter expressions, [read our guide on filtering, faceted search and filter expressions](/learn/advanced/filtering_and_faceted_search.md).
```
maryamsulemani97, line 349:
```suggestion
For more information on how to use filters and build filter expressions, [read our guide on filtering, faceted search, and filter expressions](/learn/advanced/filtering_and_faceted_search.md).
```

learn/advanced/filtering_and_faceted_search.md
```diff
@@ -102,10 +102,14 @@ The [`GET` route of the search endpoint](/reference/api/search.md#search-in-an-i
...
 
-- `NOT` only returns documents that do not satisfy a condition : `NOT genres = horror`
+- `NOT` returns all documents that do not satisfy a condition. The expression `NOT genres = horror` returns all documents whose `genres` do not contain `horror` and all documents missing a `genres` field
 - `AND` operates by connecting two conditions and only returns documents that satisfy both of them: `genres = horror AND director = 'Jordan Peele'`
 - `OR` connects two conditions and returns results that satisfy at least one of them: `genres = horror OR genres = comedy`
 - `TO` is equivalent to `>= AND <=`. The expression `release_date 795484800 TO 972129600` translates to `release_date >= 795484800 AND release_date <= 972129600`
+- `IN [valueA, valueB, …, valueN]` selects all documents whose chosen field contains at least one of the specified values. The expression `genres IN [horror, comedy]` returns all documents whose `genres` includes either `horror`, `comedy`, or both
+- `EXISTS` checks for the existence of a field. Fields with empty or null values still count as existing. The expression `release_date NOT EXISTS` returns all documents without a `release_date`
```
maryamsulemani97, line 110: We should mention that both 
```
attribute NOT EXISTS
NOT attribute EXISTS
``` 
are valid and equal

learn/advanced/filtering_and_faceted_search.md
```diff
@@ -102,10 +102,14 @@ The [`GET` route of the search endpoint](/reference/api/search.md#search-in-an-i
 
 String expressions combine conditions using the following filter operators and parentheses:
 
-- `NOT` only returns documents that do not satisfy a condition : `NOT genres = horror`
+- `NOT` returns all documents that do not satisfy a condition. The expression `NOT genres = horror` returns all documents whose `genres` do not contain `horror` and all documents missing a `genres` field
 - `AND` operates by connecting two conditions and only returns documents that satisfy both of them: `genres = horror AND director = 'Jordan Peele'`
 - `OR` connects two conditions and returns results that satisfy at least one of them: `genres = horror OR genres = comedy`
 - `TO` is equivalent to `>= AND <=`. The expression `release_date 795484800 TO 972129600` translates to `release_date >= 795484800 AND release_date <= 972129600`
+- `IN [valueA, valueB, …, valueN]` selects all documents whose chosen field contains at least one of the specified values. The expression `genres IN [horror, comedy]` returns all documents whose `genres` includes either `horror`, `comedy`, or both
```
maryamsulemani97, line 109: "The expression `genres IN [horror, comedy]` returns all documents whose `genres` includes either `horror`, `comedy`, or both"
This does cover it, but do you think it would be helpful to add that:
```
attribute IN[value1, value2, value3,]
attribute = value1 OR attribute = value2 OR attribute = value3
```
are the same

learn/advanced/filtering_and_faceted_search.md
```diff
@@ -102,10 +102,14 @@ The [`GET` route of the search endpoint](/reference/api/search.md#search-in-an-i
 
 String expressions combine conditions using the following filter operators and parentheses:
 
-- `NOT` only returns documents that do not satisfy a condition : `NOT genres = horror`
+- `NOT` returns all documents that do not satisfy a condition. The expression `NOT genres = horror` returns all documents whose `genres` do not contain `horror` and all documents missing a `genres` field
```
maryamsulemani97, line 105: We should update the example on line 217. The existing example would return movies that weren't directed by Tim Burton and movies that don't have a `director` field.
Maybe verify the results for the example on line 225 as well

maryamsulemani97 (changes requested):
1. Since we have time, do you think we should restructure the "Filter expressions" section?
- Instead of listing each operator as a bullet, we could convert it to a sub-heading with an example, like Loïc has done in the spec. It will make it a lot easier to explain each filter and it looks a lot more organised
2. We should update the example (`/learn/advanced/filtering_and_faceted_search.html#example-2`) to add the `EXISTS` and `IN` operators

learn/advanced/filtering_and_faceted_search.md
guimachiavelli: As discussed over slack: we'll address this at the imminent rewrite of this guide in the near future. For now we'll keep the descriptions lean.

learn/advanced/filtering_and_faceted_search.md
guimachiavelli: Same as below: we'll address this post v0.29, when rewriting the guide.

guimachiavelli:
Thanks for the feedback, @maryamsulemani97—as we discussed over slack, your suggestions are great but the current structure for the operators doesn't really support adding exhaustive information. We will review the whole filters guide soon (#1902) and keep the current version, warts and all, for the release.

learn/advanced/filtering_and_faceted_search.md
maryamsulemani97: I don't see it 👀

guimachiavelli: That's because I forgot to push 😅 

Should be there now.

learn/advanced/filtering_and_faceted_search.md
```diff
@@ -220,6 +224,12 @@ You can use this filter when searching for `Planet of the Apes`:
 
 <CodeSamples id="filtering_guide_3" />
 
+`NOT director = "Tim Burton"` will include both documents that do not contain `"Tim Burton"` in its `director` field and documents without a `director` field. To return only documents that have a `director` field, expand the filter expression with the `EXISTS` operator:
```
maryamsulemani97, line 227: I think you should move this to 217, as we're saying two different things. But then the code sample won't be accurate. How about we get rid of the code sample for now and explain with just SQL?

maryamsulemani97: That makes sense 👍

maryamsulemani97 (approved):
🛸

PR #1889, opened 2022-09-20
indexes.md add periods at the end of sentences

maryamsulemani97:


learn/core_concepts/indexes.md
```diff
@@ -69,69 +69,69 @@ By default, every document field is searchable and displayed in response to sear
...
+[Learn more about displayed and searchable attributes.](/learn/configuration/displayed_searchable_attributes.md)
 
 ### Distinct attribute
 
 If your dataset contains multiple similar documents, you may want to return only one on search. Suppose you have numerous black jackets in different sizes in your `costumes` index; setting `costume_name` as the distinct attribute will mean Meiliserch will not return more than one black jacket with the same `costume_name`.
 
 Designate the distinct attribute using the [update settings endpoint](/reference/api/settings.md#update-settings) or the [update distinct attribute endpoint](/reference/api/settings.md#update-distinct-attribute). **You can only set one field as the distinct attribute per index.**
 
-[Learn more about distinct attribute](/learn/configuration/distinct.md)
+[Learn more about the distinct attribute.](/learn/configuration/distinct.md)
```
guimachiavelli, line 80: This is really small and unimportant, but I am a bit bothered by the `the` here. I think we could use the plural like the other index settings—not because an index can have multiple distinct attributes, but because we're referring to distinct attributes as a category.

An example to try and illustrate how it works in my head. If we were documenting how to get dressed we could still say "learn more about hats," despite most humans being able to wear only one hat. "Learn more about the hat" is still fine and correct, but feels less natural. Does this make sense? Or is it a false equivalence?

```suggestion
[Learn more about distinct attributes.](/learn/configuration/distinct.md)
```

guimachiavelli (changes requested):
One minor comment about distinct attributes, everything else LGTM ✌️.

learn/core_concepts/indexes.md
maryamsulemani97: Makes sense, I didn't think of it that way

maryamsulemani97, line 80:
```suggestion
[Learn more about distinct attributes.](/learn/configuration/distinct.md)
```

guimachiavelli (approved):
🎩

maryamsulemani97:
bors merge

PR #1891, opened 2022-09-20
quick start: remove duplicate note

maryamsulemani97:


dichotommy (commented):
Thanks for catching this. Would you mind switching it so that you delete the above note and keep the note at this location? Either wording is good so I don't have a preference about which one you use, but I prefer this position since I'd rather not interrupt users right as they start the first section of the quick start (and before the term "document" has even been defined).

maryamsulemani97:
bors merge

PR #1895, opened 2022-09-21
Clarify some ambiguity in updating guide

dichotommy:
It may not have been clear that this sentence only applies to Meilisearch versions < 0.21


maryamsulemani97 (commented):
We do have a note telling them to skip this step if they're using >v0.21

dichotommy:
I know, just didn't want users to think that dumps still use `GET /documents` since it's no longer the case :)

maryamsulemani97 (approved):
bors merge

PR #1896, opened 2022-09-21
standardize formatting

### LINKED ISSUE #831: Standardize searchable attributes formatting
The searchable attributes list is sometimes referred to as `searchableAttributes` in the documentation, because that's the name of the field in the settings object. Other times it's referred to as "the searchable attributes list". On the other hand, the route itself is searchable-attributes. We should be more consistent and intentional with these spellings.

Need to take a closer look at the pages in question and ensure that the correct format is used in the appropriate place.

maryamsulemani97:
closes #831

### DIFF
--- a/learn/configuration/settings.md
+++ b/learn/configuration/settings.md
@@ -18,7 +18,7 @@ This page describes the **index-level settings** available in Meilisearch and ho
 
 ## Displayed attributes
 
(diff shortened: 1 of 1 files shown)


dichotommy (approved):
LGTM!

maryamsulemani97:
bors merge

PR #1899, opened 2022-09-22
Highlight behavior regarding searchableAttributes

### LINKED ISSUE #1256: Highlight behavior regarding `searchableAttributes`
If I'm not wrong this behavior is not described in the documentation and I don't really know where it should be added so I open an issue instead of a PR.
Feel free to not document this behavior if you think it brings too much complexity for a little help.

----

Currently, even if attributes are not set in `searchableAttributes`, the highlight and `matches` still apply to them.

Example using MeiliSearch v0.23.1

- Add this document
```json
{
  "id": 1,
  "title_fr": "Le Petit Prince",
  "title_en": "The Little Prince"
}
```

- Set only `title_fr` as `searchableAttributes`

- Do the following search 
```json
{
    "q": "prince",
    "attributesToHighlight": ["*"],
    "matches": true
}
```
Here is the following response:
```json
{
    "hits": [
        {
            "title_fr": "Le Petit Prince",
            "id": 1,
            "title_en": "The Little Prince",
            "_formatted": {
                "title_fr": "Le Petit <em>Prince</em>",
                "id": "1",
                "title_en": "The Little <em>Prince</em>"
            },
            "_matchesInfo": {
                "title_en": [
                    {
                        "start": 11,
                        "length": 6
                    }
                ],
                "title_fr": [
                    {
                        "start": 9,
                        "length": 6
                    }
                ]
            }
        }
    ],
    "nbHits": 2,
    "exhaustiveNbHits": false,
    "query": "prince",
    "limit": 20,
    "offset": 0,
    "processingTimeMs": 0
}
```

Despite only `title_fr` is considered as searchable, `title_en` is also highlighted and "matched".

Of course the row document is still available outside the `_formatted` field. Also, the users can set only `title_fr` as `attributesToHighlight` if they want to only highlight `title_fr`

maryamsulemani97:
closes #1256 

- Note vs warning. We agreed to use warnings when "missing the message could lead to really bad consequences". We won't have any bad consequences in this case, but users may not be aware of this so a warning sounds better
- I think a warning/note is enough to explain this behavior. But if we want to add code samples, we could add one to  `learn/configuration/displayed_searchable_attributes.md`. WDYT?

### DIFF
--- a/learn/configuration/displayed_searchable_attributes.md
+++ b/learn/configuration/displayed_searchable_attributes.md
@@ -25,8 +25,12 @@ Meilisearch uses an ordered list to determine which attributes are searchable. T
 
 In other words, the `searchableAttributes` list serves two purposes:
 
-1. It designates the fields that are searchable.
-2. It dictates the [attribute ranking order](/learn/core_concepts/relevancy.md#attribute-ranking-order).
+1. It designates the fields that are searchable
+2. It dictates the [attribute ranking order](/learn/core_concepts/relevancy.md#attribute-ranking-order)
+
+::: warning
+Documents with matching query terms will still [highlight](/reference/api/search.md#attributes-to-highlight) and [match](/reference/api/search.md#show-matches-position) attributes even if they are not part of `searchableAttributes`.
+:::
 
 There are two possible modes for the `searchableAttributes` list.
 
--- a/learn/configuration/displayed_searchable_attributes.md
+++ b/learn/configuration/displayed_searchable_attributes.md
@@ -2,8 +2,8 @@
 
 By default, whenever a document is added to Meilisearch, all new attributes found in it are automatically added to two lists:
 
-- [`searchableAttributes`](/learn/configuration/displayed_searchable_attributes.md#the-searchableattributes-list): attributes whose values are searched for matching query words.
-- [`displayedAttributes`](/learn/configuration/displayed_searchable_attributes.md#displayed-fields): attributes whose fields are displayed in documents.
+- [`searchableAttributes`](/learn/configuration/displayed_searchable_attributes.md#the-searchableattributes-list): Attributes whose values are searched for matching query words
+- [`displayedAttributes`](/learn/configuration/displayed_searchable_attributes.md#displayed-fields): Attributes whose fields are displayed in documents
 
 This means that by default, every field in a document is **searchable** and **displayed**. These properties can be modified in the [settings](/reference/api/settings.md).
 
@@ -27,10 +27,6 @@ In other words, the `searchableAttributes` list serves two purposes:
 
 1. It designates the fields that are searchable
 2. It dictates the [attribute ranking order](/learn/core_concepts/relevancy.md#attribute-ranking-order)
-
-::: warning
-Documents with matching query terms will still [highlight](/reference/api/search.md#attributes-to-highlight) and [match](/reference/api/search.md#show-matches-position) attributes even if they are not part of `searchableAttributes`.
-:::
 
 There are two possible modes for the `searchableAttributes` list.
 
--- a/reference/api/search.md
+++ b/reference/api/search.md
@@ -565,7 +565,7 @@ By default highlighted elements are enclosed in `<em>` and `</em>` tags. You may
 :::
 
 ::: warning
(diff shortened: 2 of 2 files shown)


reference/api/search.md
```diff
@@ -638,6 +642,10 @@ Though it is not necessary to use `highlightPreTag` and `highlightPostTag` in co
 
 Adds a `_matchesPosition` object to the search response that contains the location of each occurrence of queried terms across all fields. This is useful when you need more control than offered by our [built-in highlighting](#attributes-to-highlight). `showMatchesPosition` only works for strings, numbers, and arrays of strings and numbers.
 
+::: warning
+`showMatchesPosition` will consider all attributes that contain words matching the query terms even if they are not set as `searchableAttributes`.
```
guimachiavelli, line 646: "consider" is a bit unclear here. Can we find a better verb, or group of words?

I'm contrasting it to the new warning on L#568, where "highlight" tells me what happens even if I haven't read anything else about that parameter.

guimachiavelli (dismissed):
Minor comment on the verb we're using for `showMatchesPosition`. Not sure if there's anything better, though.

reference/api/search.md
maryamsulemani97: I can only think of "takes into account", but that's not helpful either 🤔

guimachiavelli: What about "returns match location"?

```suggestion
`showMatchesPosition` returns match location for all attributes that contain words matching the query terms even if they are not set as `searchableAttributes`.
```

learn/configuration/displayed_searchable_attributes.md
```diff
@@ -25,8 +25,12 @@ Meilisearch uses an ordered list to determine which attributes are searchable. T
 
 In other words, the `searchableAttributes` list serves two purposes:
 
-1. It designates the fields that are searchable.
-2. It dictates the [attribute ranking order](/learn/core_concepts/relevancy.md#attribute-ranking-order).
+1. It designates the fields that are searchable
+2. It dictates the [attribute ranking order](/learn/core_concepts/relevancy.md#attribute-ranking-order)
+
+::: warning
```
dichotommy, line 31: I would vote for removing this warning. Although this behavior involves `searchableAttributes`, it only affects people using attributes to highlight and matches. Therefore, we only need to give a warning in pages/sections related to those parameters.

reference/api/search.md
```diff
@@ -564,6 +564,10 @@ By default highlighted elements are enclosed in `<em>` and `</em>` tags. You may
 `attributesToHighlight` also highlights terms configured as [synonyms](/reference/api/settings.md#synonyms) and [stop words](/reference/api/settings.md#stop-words).
 :::
 
+::: warning
+`attributesToHighlight` will highlight all attributes added to the `attributesToHighlight` array even if they are not set as `searchableAttributes`.
```
dichotommy, line 568:
```suggestion
`attributesToHighlight` will highlight matches within all attributes added to the `attributesToHighlight` array, even if those attributes are not set as [`searchableAttributes`](/learn/configuration/displayed_searchable_attributes.md#searchable-fields).
```

reference/api/search.md
```diff
@@ -638,6 +642,10 @@ Though it is not necessary to use `highlightPreTag` and `highlightPostTag` in co
 
 Adds a `_matchesPosition` object to the search response that contains the location of each occurrence of queried terms across all fields. This is useful when you need more control than offered by our [built-in highlighting](#attributes-to-highlight). `showMatchesPosition` only works for strings, numbers, and arrays of strings and numbers.
 
+::: warning
+`showMatchesPosition` returns match location for all attributes that contain words matching the query terms even if they are not set as `searchableAttributes`.
```
dichotommy, line 646:
```suggestion
`showMatchesPosition` returns the location of matched query terms within all attributes, even attributes that are not set as `searchableAttributes`.
```

learn/configuration/displayed_searchable_attributes.md
maryamsulemani97: That makes sense

dichotommy (approved):
LGTM 🌈

maryamsulemani97:
bors merge

maryamsulemani97:
bors merge

PR #1905, opened 2022-09-27
Known limitations: add maximum number of concurrent requests

### LINKED ISSUE #1870: Known limitation: maximum number of requests handled
I'm here to share a known limitation of Meilisearch, it's up to you to put it in the docs or not 😇 

Meilisearch can handle maximum 126 requests at the same time. "Handle" not "process". It means the users can send 126 requests maximum in the request queue.

More info here: https://github.com/meilisearch/meilisearch/issues/2648

Pinging also @Kerollmops who will be the best to review the PR

guimachiavelli:
Closes #1870

### DIFF
--- a/learn/advanced/known_limitations.md
+++ b/learn/advanced/known_limitations.md
@@ -57,6 +57,12 @@ If your query is `Hello - World`:
 
 **Explanation:** This limit is enforced for performance and storage reasons. Overly large internal data structures—resulting from documents with too many fields—lead to overly large databases on disk, and slower search performance.
 
+## Maximum number of concurrent requests
+
+**Limitation:** Meilisearch can handle a maximum of 126 requests at the same time. Requests exceeding this limit will cause Meilisearch to thrown an internal error.
+
+**Explanation:** This limit is set by [LMDB](/learn/advanced/storage.md#lmdb), Meilisearch's storage engine.
+
 ## Maximum number of documents in an index
 
 **Limitation:** An index can contain no more than 4,294,967,296 documents.
--- a/learn/advanced/known_limitations.md
+++ b/learn/advanced/known_limitations.md
@@ -59,9 +59,9 @@ If your query is `Hello - World`:
 
 ## Maximum number of concurrent requests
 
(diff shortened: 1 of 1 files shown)


learn/advanced/known_limitations.md
```diff
@@ -57,6 +57,12 @@ If your query is `Hello - World`:
 
 **Explanation:** This limit is enforced for performance and storage reasons. Overly large internal data structures—resulting from documents with too many fields—lead to overly large databases on disk, and slower search performance.
 
+## Maximum number of concurrent requests
+
+**Limitation:** Meilisearch can receive a maximum of 126 API requests at the same time. Requests exceeding this limit will cause Meilisearch to thrown an internal error.
```
Kerollmops, line 62:
```suggestion
**Limitation:** Meilisearch can receive a maximum of 126 API requests at the exact same time. Requests exceeding this limit will cause Meilisearch to thrown an internal error.
```

I would like to add this extra precision because I want people to fill this situation to be nearly impossible to trigger.

learn/advanced/known_limitations.md
```diff
@@ -57,6 +57,12 @@ If your query is `Hello - World`:
...
+
+**Explanation:** This limit applies only to receiving requests, not to processing or adding them to the Tasks queue. It is set by [LMDB](/learn/advanced/storage.md#lmdb), Meilisearch's storage engine.
```
Kerollmops, line 64: It does apply to the write calls, too, if you are trying to send more than 126 HTTP calls at the exact same time (quite complex) this error can be triggered.

Kerollmops (dismissed):

learn/advanced/known_limitations.md
guimachiavelli: Apologies for my ignorance, but what would be the difference between a request and a write call in this case? Is there a term we can use to refer to both things at the same time?

Kerollmops: I was a little bit too picky here, you can keep what you wrote. I was referring that write operations (like inserting tasks) can also throw errors, but it is, in practice impossible, to do so.

maryamsulemani97 (approved):
👻

guimachiavelli:
bors merge

PR #1906, opened 2022-09-27
Known limitations: large datasets

guimachiavelli:
Closes 1874

### DIFF
--- a/learn/advanced/known_limitations.md
+++ b/learn/advanced/known_limitations.md
@@ -100,3 +100,9 @@ user = 1 OR user = 2 […] OR user = 1500 OR user = 1501 […] OR user = 2000 OR
 **Limitation:** By default, Meilisearch returns up to 1000 documents per search.
 
 **Explanation:** Meilisearch limits the maximum amount of returned search results to protect your database from malicious scraping. You may change this by using the `maxTotalHits` property of the [pagination index settings](/reference/api/settings.md#pagination-object). `maxTotalHits` only applies to the [search route](/reference/api/search.md) and has no effect on the [get documents endpoint](/reference/api/documents.md#get-documents).
+
+## Large datasets and internal errors
+
+**Limitation:** Meilisearch might throw an internal error when indexing large datasets.
+
+**Explanation:** Indexing datasets over 2GB in size can result in Meilisearch opening too many file descriptors. Raise your machine's resource usage limits with `ulimit` or a similar tool available in your environment.


learn/advanced/known_limitations.md
```diff
@@ -100,3 +100,9 @@ user = 1 OR user = 2 […] OR user = 1500 OR user = 1501 […] OR user = 2000 OR
 **Limitation:** By default, Meilisearch returns up to 1000 documents per search.
 
 **Explanation:** Meilisearch limits the maximum amount of returned search results to protect your database from malicious scraping. You may change this by using the `maxTotalHits` property of the [pagination index settings](/reference/api/settings.md#pagination-object). `maxTotalHits` only applies to the [search route](/reference/api/search.md) and has no effect on the [get documents endpoint](/reference/api/documents.md#get-documents).
+
+## Large datasets and internal errors
+
+**Limitation:** Meilisearch might throw an internal error when indexing large datasets.
```
Kerollmops, line 106:
```suggestion
**Limitation:** Meilisearch might throw an internal error when indexing large batches of documents.
```

It is not related to the final size of the dataset but the size of one batch of incoming documents.

learn/advanced/known_limitations.md
```diff
@@ -100,3 +100,9 @@ user = 1 OR user = 2 […] OR user = 1500 OR user = 1501 […] OR user = 2000 OR
...
+
+**Explanation:** Indexing datasets over 2GB in size can result in Meilisearch opening too many file descriptors. Depending on your system, this might reach your system's resource usage limits and trigger an internal error. Use `ulimit` or a similar tool to increase resource consumption limits.
```
Kerollmops, line 108:
```suggestion
**Explanation:** Indexing a batch of documents that is over 2GB in size can result in Meilisearch opening too many file descriptors. Depending on your system, this might reach your system's default resource usage limits and trigger an internal error. Use `ulimit` or a similar tool to increase resource consumption limits e.g. call `ulimit -Sn 3000` just before running Meilisearch.
```

Kerollmops (changes requested):

learn/advanced/known_limitations.md
maryamsulemani97, line 106: We should quantify "large" as we do in the other Known limitations

learn/advanced/known_limitations.md
maryamsulemani97, line 108: - The issue mentions 2GB for CSV and 3.5GB for JSON. Should we mention this as well? 
- Should we add a link or maybe a line or two on what `ulimit` is and how to use it?

learn/advanced/known_limitations.md
guimachiavelli: The problem is that we don't have an unambiguous answer as in other items. This limit can be 2GB CSV/3.5GB JSON in a system, but might be higher in some machines and lower in others.

My idea was to be sort of vague at first and then give a generic explanation in the following paragraph.

learn/advanced/known_limitations.md
guimachiavelli: > The issue mentions 2GB for CSV and 3.5GB for JSON. Should we mention this as well?

I think mentioning the two is not super useful (they are the same dataset, but in different formats), but you're right that we should make it a bit more explicit.

maryamsulemani97 (approved):
🛸

Kerollmops (approved):
Perfect, thank you!

guimachiavelli:
don't worry, buddy, it's fine, we're all struggling. let's try again

bors merge

PR #1907, opened 2022-09-29
update node to ensure compatibility with node 16

### LINKED ISSUE #1657: Make netlify use node 16
**Context**
Node 14 is now in maintenance. The two current active supported versions are node 16 and node 18.
See [issue here](https://github.com/meilisearch/integration-guides/issues/196) for more explanation.

** Issue **
The documentation is still built in node 14 and should be updated to node 16 for the following reasons: 
- For security reasons
- To stay up to date with your own local node environment (which should be updated to a higher version) 
- To be compatible with the dependencies that might update to versions using features that are not in node 14.

I assume because the CI builts the documentation in node 14 that netlify also builds it in node 14.

I would suggest that Netlify builds the documentation using `node 16` and not node 14 anymore.

maryamsulemani97:
closes #1657 + removes support for node 12 + renames  `gh-pages-scarping.yml`(not very helpful) to `deploy.yml`

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "vuex": "^3.6.2"
   },
   "engines": {
-    "node": ">=16"
+    "node": ">=12 <=18"
   },
   "resolutions": {
     "vuepress-plugin-check-md/check-md": "https://github.com/bidoubiwa/check-md#add_ignore_pattern"
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "vuex": "^3.6.2"
   },
   "engines": {
-    "node": ">=12 <=18"
+    "node": ">=14 <=16"
   },
   "resolutions": {
     "vuepress-plugin-check-md/check-md": "https://github.com/bidoubiwa/check-md#add_ignore_pattern"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -44,7 +44,7 @@
     "vuex": "^3.6.2"
   },
   "engines": {
-    "node": ">=12 <=16"
+    "node": ">=16"
```
maryamsulemani97, line 47:
```suggestion
    "node": ">=12 <=18"
```

bidoubiwa:
Can you change the title with something in the lines of `Ensure compatibility with node 16` ?
Otherwise looks good to me

bidoubiwa (approved):

dichotommy:
Correct me if I'm wrong @bidoubiwa but I think this change means that the site will use any Node version from 12 to 18 inclusive to build. Would it be better to set an explicit version number, or increase the lower limit so we're sure it uses Node 16?

dichotommy:
@maryamsulemani97 , could you also change our CIs so they use Node 16? Thanks 🙏🏻

bidoubiwa:
Hey @dichotommy we just had a talk to discuss everything and this Pr is going to be updated accordingly :)

bidoubiwa (approved):
LGTM 🔥

maryamsulemani97:
bors merge

Issue #840, opened 2021-03-03
Documentation around How to Handle Dates / Best Practices


cryptoroo:
**Is your feature request related to a problem? Please describe.**
Having used other search engines I am used to have a datetime type. 
Usually ISO-8601 is the preferred format. 
I haven't seen any recommendations on how to deal with dates in MeiliSearch. 
It would be useful to understand what your thoughts as the developers of the product are around this. 

**Describe the solution you'd like**
Some clear examples on how you can use dates for search filters but also for facets. 
Any roadmaps for including native date parsing? 

**Describe alternatives you've considered**
The cleanest solution to date has been to use Linux Timestamps as they are Integers which are very easily represented in MeiliSearch

react-learner:
This is indeed something that should be documented, ideally in a tutorial or guide. Thanks for bringing it to our attention, @cryptoroo !

lukasa1993:
so timestamps are only eay forward ? no datetime support timezones and stuff ?

curquiza:
For the moment only timestamps are handle.

I don't think the datetime support is in our current roadmap. However we have a public roadmap where our users can share their ideas and their needs. It would be awesome if you can add your idea so that other users will upvote it if they need it as well.
https://roadmap.meilisearch.com/tabs/5-ideas

alfonsrv:
Not sure how much overhead this would add to the search speed, but being able to declare a data-type for each `filterableAttributes` could be one idea.

maryamsulemani97:
fixed in #1917

cryptoroo:
thanks all. Unix timestamps are absolutely fine :D

PR #1917, opened 2022-10-05
New guide: Filtering and sorting with chronological values

### LINKED ISSUE #982: Create guide on converting dates to unix timestamps, filtering + sorting by date
Related to #840 

Would cover essentially the same material as [this Algolia guide](https://www.algolia.com/doc/guides/managing-results/refine-results/filtering/how-to/filter-by-date/).

guimachiavelli:
## Reviewers

Code samples are embedded in the guide to make the review process easier. As usual, we will move them to `.code-samples.yaml` before merging.

Closes #982 and #840


learn/advanced/chronological_filtering_sorting.md
```diff
@@ -0,0 +1,109 @@
...
+const timestamp = Date.parse(game.year);
+
+game = {
+  "id": 0,
+  "title": "Return of the Obra Dinn",
+  "genre": "adventure",
+  "year": 2018,
+  "release_date": timestamp
+};
+```
```
guimachiavelli, line 53: Is this a necessary? I want to show the conversion process, but completely isolated from an actual  script for data parsing/formatting.

learn/advanced/chronological_filtering_sorting.md
```diff
@@ -0,0 +1,109 @@
+# Filtering and sorting with chronological values
+
+In this guide, you will learn about Meilisearch's approach to date and time values, how to prepare your dataset for indexing, and how to chronologically sort and filter search results.
+
+## Preparing your documents
```
guimachiavelli, line 5: I present the ideal case (dataset contains timestamps) first, then the likely scenario of a dataset using datetime.

Alternatively, we could separate these two cases in two sections, or even make the datetime scenario a subsection of `## Preparing your documents`. 

Any opinions?

learn/advanced/chronological_filtering_sorting.md
```diff
@@ -0,0 +1,109 @@
...
+};
+```
+
+:::tip
+When preparing your dataset, it can be useful to leave the date and time fields in your documents intact. In the example above, we keep the `year` field because it is more readable than the raw `timestamp`.
+:::
+
+After adding a numeric timestamp to all documents, index your dataset as usual.
+
+## Chronological filtering
```
guimachiavelli, line 61: I have chosen to use "Chronological" throughout the article to avoid entering the pitfall of date (YYYY-MM-DD) vs time (18:27).

Is "chronological" too complex a word?

learn/advanced/chronological_filtering_sorting.md
```diff
@@ -0,0 +1,109 @@
...
+  "year": 2018,
+  "release_date": timestamp
+};
+```
+
+:::tip
+When preparing your dataset, it can be useful to leave the date and time fields in your documents intact. In the example above, we keep the `year` field because it is more readable than the raw `timestamp`.
+:::
+
+After adding a numeric timestamp to all documents, index your dataset as usual.
```
guimachiavelli, line 59: Am I silly or does this sentence feel a bit lost and alone here?

learn/advanced/chronological_filtering_sorting.md
```diff
@@ -0,0 +1,109 @@
+# Filtering and sorting with chronological values
+
+In this guide, you will learn about Meilisearch's approach to date and time values, how to prepare your dataset for indexing, and how to chronologically sort and filter search results.
+
+## Preparing your documents
+
+To filter and sort search results based on chronological values, your documents must have a numeric field containing a UNIX timestamp. In this videogames dataset, the release year is formatted as a timestamp:
```
maryamsulemani97, line 7: Pretty sure everyone reading would probably know what UNIX timestamps are, but maybe add a link for it? We do that for the ISO 8601 and RFC 3339 formats too.

learn/advanced/chronological_filtering_sorting.md
maryamsulemani97: You can ignore this if you want
Doesn't chronological mean arranging in the order it happened? Chronological sorting makes sense to me; you're sorting by oldest to newest or the other way round, but not chronological filtering. I didn't find any search results for `chronological filtering`. Should we go with something like Filtering or sorting by date? It will also get rid of any complexity "chronological" brings

learn/advanced/chronological_filtering_sorting.md
maryamsulemani97: I don't get this 👀

learn/advanced/chronological_filtering_sorting.md
maryamsulemani97: Doesn't filtering by date, generally, mean you can filter by time too? At least, that's how I see it

learn/advanced/chronological_filtering_sorting.md
```diff
@@ -0,0 +1,116 @@
+# Filtering and sorting by date
+
+In this guide, you will learn about Meilisearch's approach to date and time values, how to prepare your dataset for indexing, and how to chronologically sort and filter search results.
+
+## Preparing your documents
+
+To filter and sort search results based on chronological values, your documents must have a numeric field containing a [UNIX timestamp](https://kb.narrative.io/what-is-unix-time). In this videogames dataset, the release year is formatted as a timestamp:
```
dichotommy, line 7:
```suggestion
To filter and sort search results chronologically, your documents must have at least one numeric field containing a [UNIX timestamp](https://kb.narrative.io/what-is-unix-time).

As an example, consider a database of video games. In this dataset, the release year is formatted as a timestamp:
```

learn/advanced/chronological_filtering_sorting.md
```diff
@@ -0,0 +1,116 @@
...
+    "genre": "adventure",
+    "release_date": 1640991600
+  },
+  {
+    "id": 2,
+    "title": "Bayonetta 2",
+    "genre": "action",
+    "release_date": 1388530800
+  }
+]
```
dichotommy, line 29: After sticking these dates into a unix timestamp converter, I was a bit confused that they didn't match up with the actual release dates of the games in question. Then I realized that they are supposed to be the release year, not the release date. I think it would make more sense for this guide to use specific dates as examples, since filtering or sorting by year is easy even without using a timestamp.

learn/advanced/chronological_filtering_sorting.md
```diff
@@ -0,0 +1,116 @@
...
+:::
+
+After adding a numeric timestamp to all documents, index your data as usual:
+
+```sh
+curl \
+  -X POST 'http://localhost:7700/indexes/games/documents' \
+  -H 'Content-Type: application/json' \
+  --data-binary @games.json
+```
```
dichotommy, line 66: This example is not related to the specific aims of this guide, and will not run for the user since there is no games.json file. My initial reaction was that it should be removed. However, I see how it can be helpful for "continuity", so to speak, since later commands are using the `games` index. So if we keep it, it needs to be clear that it is an example.

````suggestion
After adding a numeric timestamp to all documents, [index your data](/reference/api/documents.md#add-or-replace-documents) as usual. See below for an example using our nonexistent video games dataset.

```sh
curl \
  -X POST 'http://localhost:7700/indexes/games/documents' \
  --data-binary @games.json
```
````

File: learn/advanced/chronological_filtering_sorting.md [hunk 9]
```diff
@@ -0,0 +1,116 @@
...
+  -H 'Content-Type: application/json' \
+  --data-binary '{
+    "q": "",
+    "filter": "release_date >= 1514761200 AND release_date < 1672527600"
+  }'
+```
+
+## Sorting by timestamp
+
+To sort search results chronologically, add your document's timestamp field to the `sortableAttributes` index setting:
```
dichotommy, line 95: I prefer the wording you used above. I think it sounds a bit strange to talk about **adding** a field **to an index setting**, whereas **adding an element to a list** is very intuitive.

We should also provide a link for sortable attributes, as we did for filterable attributes above.

```suggestion
To sort search results chronologically, add your document's timestamp field to the list of [`sortableAttributes`](/reference/api/settings.md#update-sortable-attributes):
```

learn/advanced/chronological_filtering_sorting.md
```diff
@@ -0,0 +1,116 @@
...
+
+```sh
+curl \
+  -X POST 'http://localhost:7700/indexes/games/search' \
+  -H 'Content-Type: application/json' \
+  --data-binary '{
+    "q": "adventure",
+    "sort": ["release_date:desc"]
+  }'
+```
```
dichotommy, line 116: I would suggest removing the adventure query and just return all games sorted from most recent to oldest. Simpler examples are better!

````suggestion
Once you have configured `sortableAttributes`, you can sort your search results based on their timestamp. The following query returns all games sorted from most recent to oldest:

```sh
curl \
  -X POST 'http://localhost:7700/indexes/games/search' \
  --data-binary '{
    "q": "",
    "sort": ["release_date:desc"]
  }'
```
````

**Review by dichotommy — CHANGES_REQUESTED** [review]
Hey, the structure of this guide looks great to me. I think the examples in the beginning can be improved a bit; we should work with a specific date instead of just a year, since years are fully numeric values that are already easy to filter and sort.

Other than that, looks great to me 👌🏻 💯

**Comment by dichotommy** [comment]
Two more things:

- I like how descriptive the title of the guide is, but perhaps it should be more general. What do you think about renaming it "Working with dates"?
- Don't forget to add the new page to the sidebar 😉

File: learn/advanced/working_with_dates.md [hunk 11]
```diff
@@ -0,0 +1,118 @@
...
+
+```sh
+curl \
+  -X POST 'http://localhost:7700/indexes/games/search' \
+  -H 'Content-Type: application/json' \
+  --data-binary '{
+    "q": "",
+    "filter": "release_timestamp >= 1514761200 AND release_timestamp < 1672527600"
+  }'
+```
```
curquiza, line 93: Maybe an example in JS would be more convenient to read than curl. I mean, maybe we could have a code part ([with `Date`](https://stackoverflow.com/questions/9873197/how-to-convert-date-to-timestamp)?) where we transform `2018` and `2022` into a timestamp. It's easier to understand we are searching between 2018 and 2022 with the date written than seeing `1514761200` and `1672527600` and guessing it means `2018` and `2022`

@bidoubiwa WDYT? Maybe it's not performant to do this?

bidoubiwa: I agree that javascript is easier to read; the issue is that we need to import meilisearch-js.
Are you thinking about something like that: 

```js
const { MeiliSearch } = require('meilisearch')
const client = new MeiliSearch({ host: '...' })

let documents = [ { id: 1, release_date: "2018-10-18T00:00Z" },  {}] // etc

documents = documents.map(document => ({
  ...document,
  release_timestamp: Date.parse(document.release_date) / 1000,
}))
await client.index('games').addDocuments(documents)

// wait for tasks?

const searchResults = client.index('games').search('', {
  filters: 'release_timestamp >= 1514761200 AND release_timestamp < 1672527600',
})  // Where 1514761200 is Sunday, December 31, 2017  and 1514761200 is Saturday, December 31, 2022

```

My bad if I misunderstood. The code can be made more readable btw.

learn/advanced/working_with_dates.md
```diff
@@ -0,0 +1,118 @@
...
+  {
+    "id": 2,
+    "title": "Bayonetta 2",
+    "genre": "action",
+    "release_timestamp": 1411164000
+  }
+]
+```
+
+If your chronological data is expressed in another format, like [ISO 8601](https://www.iso.org/iso-8601-date-and-time-format.html), you must convert it to a numeric value before indexing it with Meilisearch.
```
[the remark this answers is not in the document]
guimachiavelli: I like that. Made a couple of edits directly in your suggestion, committing it now.

learn/advanced/working_with_dates.md
```diff
@@ -0,0 +1,118 @@
...
+
+```js
+let game = {
+  "id": 0,
+  "title": "Return of the Obra Dinn",
+  "genre": "adventure",
+  "release_date": "2018-10-18T00:00Z"
+};
+
+const timestamp = Date.parse(game.year);
```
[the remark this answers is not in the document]
guimachiavelli: Duh, thanks for that.

To build up on it, what about this:

```suggestion
const timestampInMilliseconds = Date.parse(game.year); // Javascript's Date.parse returns the timestamp in milliseconds
const timestamp = timestampInMilliseconds / 1000; // UNIX timestamps must be in seconds
```

learn/advanced/working_with_dates.md
bidoubiwa: That's oke for me!

learn/advanced/working_with_dates.md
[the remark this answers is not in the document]
bidoubiwa: I don't think `Javascript's` in front of `Date.parse` is useful but the rest is 👌

curquiza (dismissed):

curquiza:
Btw, I forgot to say: I think this is definitely a really valuable guide that will help the support team and the users 👍

learn/advanced/working_with_dates.md
```diff
@@ -0,0 +1,119 @@
...
+  "release_date": "2018-10-18T00:00Z",
+  "release_timestamp": timestamp
+};
+```
+
+:::tip
+When preparing your dataset, it can be useful to leave the original date and time fields in your documents intact. In the example above, we keep the `release_date` field because it is more readable than the raw `timestamp`.
+:::
+
+After adding a numeric timestamp to all documents, [index your data](/reference/api/documents.md#add-or-replace-documents) as usual. See below for an example using our nonexistent video games dataset.
```
maryamsulemani97, line 62: Technically, we do have a dataset with 3 games, so idk if we should say "nonexistent". We have a similar dataset for the geosearch guide as well

dichotommy: If we refer to "our video games dataset" but don't provide a downloadable `videogamesdataset.json` file anywhere on the page, I can only see that confusing and misleading users. I think either we keep it how it is (rewording possible), or we provide an actual video games dataset.

Personally I support the former option, because it's less work and I don't think that a downloadable dataset would add much value to this guide. But maybe I'm wrong?

guimachiavelli: Perhaps `hypothetical` is a better word?

maryamsulemani97: How about the wording we used in geosearch:
>Using our [example dataset](https://docs.meilisearch.com/restaurants.json),

guimachiavelli: Followed @maryamsulemani97's suggestion, re-requesting reviews

learn/advanced/working_with_dates.md
```diff
@@ -0,0 +1,84 @@
...
+  "release_date": "2018-10-18T00:00Z",
+  "release_timestamp": timestamp
+};
+```
+
+:::tip
+When preparing your dataset, it can be useful to leave the original date and time fields in your documents intact. In the example above, we keep the `release_date` field because it is more readable than the raw `timestamp`.
+:::
+
+After adding a numeric timestamp to all documents, [index your data](/reference/api/documents.md#add-or-replace-documents) as usual. The example below adds a <a id="downloadRestaurants" href="/restaurants.json" download="restaurants.json">videogame dataset</a> to a `games` index:
```
maryamsulemani97, line 62:
```suggestion
After adding a numeric timestamp to all documents, [index your data](/reference/api/documents.md#add-or-replace-documents) as usual. The example below adds a <a id="downloadVideogames" href="/videogames.json" download="videogames.json">videogame dataset</a> to a `games` index:
```

guimachiavelli: oops! That's what happens when I'm on a rush 😓

maryamsulemani97 (approved):
🏃🏻‍♀️

PR #1933, opened 2022-10-13
Sync versions with docker pull and docker run commands to v0.29

jrinder42:
## What does this PR do?
Syncs the versions with the docker pull and docker run commands to v0.29 in the quickstart guide. Otherwise, an error will be occur.


maryamsulemani97 (approved):
Thanks for catching this @jrinder42! 🙏

bors merge

PR #1721, opened 2022-06-13
add commas after `e.g.` and `i.e.`

maryamsulemani97:

PR #1937, opened 2022-10-17
update api reference/overview

### LINKED ISSUE #1901: Update `/reference/api/overview.md`
I looked at what other docs have in their version of the “Overview” page, and we can improve ours too. Here’s what we could start with:
- There were different names, but I think something like “API conventions” or “Document conventions” would be a good heading. We could have specific headings for these “conventions” like [Zendesk](https://developer.zendesk.com/api-reference/introduction/doc-conventions/):
  - Endpoints. This was very basic, something like: “We use an HTTP method and a path”
  - Placeholder values. Explain the format and give an example. We can use this for `MATSER_KEY` and path parameters in `{}`
  - There were more, but it depends on how thorough we want to be, I guess. We could have one heading with a bullet for each convention or dedicate headings to each
- ~~API rate limits: The number of requests a client can make to the APIs in a specific time frame. Idk if we have these. I didn’t know this was a thing~~ These do not apply to us
- Request/Response formats: Some were [brief](https://developer.zendesk.com/api-reference/introduction/requests/) and some were [very thorough](https://developer.vimeo.com/api/reference). It could be helpful to explain everything, but I guess that’s not the point of the docs. But maybe briefly explain what path parameters and all of this is for beginners? 
- Data format. I found it for [Ring](https://developers.ringcentral.com/api-reference/Data-Types), and it’s a table with all the data types they use in their docs
- Pagination: This was very common. Just a few paragraphs on the kind of pagination used. We could do something like [Stripe](https://stripe.com/docs/api/pagination) or have a guide like [Zendesk](https://developer.zendesk.com/api-reference/introduction/pagination/)
- Links to OpenAPI specification files
- Details on error codes and statuses. We could move this to a separate page like the “Overview” for Errors, but many websites have this here. I guess it depends on what we want
- I don’t know if this could go here, but something on Updating. We could explain partial updates here instead of doing that on every page for PATCH requests
- API versioning, something we might after v1.0

maryamsulemani97:
closes #1901 

### For reviewers:
- The first paragraph needs help. Most API references either have something like `Our API has predictable resource-oriented URLs, accepts form-encoded request bodies, returns JSON-encoded responses, and uses standard HTTP response codes, authentication, and verbs.` or a brief description of what the API is
- Does OpenAPI need to be a section, or should I add it at the end of the "introduction"?
  - Will be part of the introduction ✅ 
- Can the current order be improved?
- Should "Data types" be a section on its own? I think it makes it more visible
- I added two small sections on request and response bodies
  - Should they be merged into one?
  - Is there enough information?


reference/api/overview.md
```diff
@@ -1,45 +1,115 @@
...
+
+| Name         | Type    | Description                  |
+| :----------- | :------ | :--------------------------- |
+| **`offset`** | Integer | Number of resources skipped  |
+| **`limit`**  | Integer | Number of resources returned |
+| **`total`**  | Integer | Total number of resources    |
+
+### `/tasks` endpoint
+
+The `/tasks` endpoint uses a different type of pagination; hence the response contains different fields. You can read more about it in the [tasks API reference](/reference/api/tasks.md#get-tasks).
```
maryamsulemani97, line 51:
```suggestion
Since the `/tasks` endpoint uses a different type of pagination, the response contains different fields. You can read more about it in the [tasks API reference](/reference/api/tasks.md#get-tasks).
```

reference/api/overview.md
```diff
@@ -1,45 +1,115 @@
...
+The Meilisearch OpenAPI specifications:
+
+- [JSON](https://bump.sh/doc/meilisearch.json)
+- [YAML](https://bump.sh/doc/meilisearch.yaml)
+
+## Document conventions
+
+The API documentation uses the following conventions:
+
+- Curly braces (`{}`) represent path parameters, e.g., GET `/indexes/{index_uid}`
```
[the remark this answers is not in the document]
maryamsulemani97: I think we should use the comma as it's used everywhere else in the documentation. 
https://github.com/meilisearch/documentation/pull/1721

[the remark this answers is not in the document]
guimachiavelli: I…never knew `e.g.,` was an actual thing and completely missed #1721. I'm against it because I always find it very awkward when two punctuation marks meet.  I mean, I know it's inevitable, but the less I see `",`, `—,`, `.,`, and so on, the happier I am. Will follow the majority, though.

Also, regarding `e.g.` specifically, I try not to use it a lot when writing for Meilisearch. [According to gov.uk](https://insidegovuk.blog.gov.uk/2016/07/20/changes-to-the-style-guide-no-more-eg-and-ie-etc/), `e.g.` can be difficult to  understand for people not super comfortable with English and screen readers don't always parse it correctly. They recommend using `for example`, `such as`, `like`, and `including` depending on the sentence.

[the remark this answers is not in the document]
maryamsulemani97: I agree we should get rid of it eventually and create a Vale rule so we use the full form instead.

reference/api/overview.md
```diff
@@ -1,45 +1,115 @@
...
+| **`limit`**  | Integer | Number of resources returned |
+| **`total`**  | Integer | Total number of resources    |
+
+### `/tasks` endpoint
+
+Since the `/tasks` endpoint uses a different type of pagination, the response contains different fields. You can read more about it in the [tasks API reference](/reference/api/tasks.md#get-tasks).
+
+## Parameters
+
+Parameters are options passed to the endpoint to influence the response. They are added at the end of the string for GET requests. For POST requests, they are added to the body.
```
guimachiavelli, line 55: Ok, this is a big edit in multiple parts. Parameters can be sort of confusing, so I hope I'm not saying anything wrong 😅 

I think it's confusing to describe query parameters (`http://example.com?a=b&c=d`) both here, mixed with request body parameters, and below, under `### Query parameters`.

My suggestions are:
- use this paragraph to list the types of parameters we use in Meilisearch's API, but refrain from explaining them;
- create a `### Request body parameters` for POST/PATCH/PUT requests
- move the stuff about GET requests here to `### Query parameters`

```suggestion
Parameters are options you can pass to an API endpoint to modify its response. There are three main types of parameter in Meilisearch's API: request body parameters, path parameters, and query parameters.

### Request body parameters

These parameters are mandatory parts of POST, PUT, and PATCH requests. They accept a wide variety of values and data types depending on the resource you're modifying. You must add these parameters to your request's data payload.

```

reference/api/overview.md
```diff
@@ -1,45 +1,115 @@
...
+
+### Path parameters
+
+These are parameters you pass to the API in the endpoint's path. They are used to identify a resource uniquely. You can have multiple path parameters, e.g., `/indexes/{index_uid}/documents/{document_id}`.
+
+We omit this section if an endpoint does not take any path parameters.
+
+### Query parameters
+
+These optional parameters are a sequence of key-value pairs and appear after the question mark (`?`) in the endpoint. You can list multiple query parameters by separating them with an ampersand (`&`). The order of query parameters does not matter.
```
guimachiavelli, line 65:
```suggestion
These optional parameters are a sequence of key-value pairs and appear after the question mark (`?`) in GET endpoints. You can list multiple query parameters by separating them with an ampersand (`&`). The order of query parameters does not matter.
```

reference/api/overview.md
```diff
@@ -1,45 +1,115 @@
...
+
+We omit this section if an endpoint does not take any query parameters.
```
guimachiavelli, line 67: I think it's good to specify where exactly we omit the query parameters section.

```suggestion
If an endpoint does not take any query parameters, this section is not present in that endpoint's documentation.
```

reference/api/overview.md
```diff
@@ -1,45 +1,115 @@
...
+Any API request with a payload (`--data-binary`) requires a `Content-Type` header. Content type headers indicate the media type of the resource, helping the client process the response body correctly.
 
-Any API request with a payload (`--data-binary`) requires a `Content-Type` header. Meilisearch currently supports the following formats:
+Meilisearch currently supports the following formats:
 
 - `Content-Type: application/json` for JSON
 - `Content-Type: application/x-ndjson` for NDJSON
 - `Content-Type: text/csv` for CSV
 
 The only endpoints which currently accept NDJSON and CSV `Content-Type` are the [add documents](/reference/api/documents.md#add-or-replace-documents) and [update documents](/reference/api/documents.md#add-or-update-documents) endpoints. For all others, use `Content-Type: application/json`.
```
guimachiavelli, line 81: Editing for brevity and directness.

```suggestion
Only the [add documents](/reference/api/documents.md#add-or-replace-documents) and [update documents](/reference/api/documents.md#add-or-update-documents) endpoints accept  NDJSON and CSV. For all others, use `Content-Type: application/json`.
```

guimachiavelli (changes requested):
I'm proposing a couple of significant changes on this one in hopes of clarifying the distinction between the types of API parameters we accept. You might have to edit them slightly in case I'm saying something not quite right 😅

reference/api/overview.md
guimachiavelli: Yeap, you're right, just saw this while reviewing the changelog. I still think it's a good idea to highlight query parameters are particularly useful for GET requests. Perhaps you can add a short sentence about it?

maryamsulemani97: Done ✅

guimachiavelli (dismissed):
🥾

reference/api/overview.md
```diff
@@ -1,44 +1,94 @@
...
+
+Any API request with a payload (`--data-binary`) requires a `Content-Type` header. Content type headers indicate the media type of the resource, helping the client process the response body correctly.
+
+Meilisearch currently supports the following formats:
+
+- `Content-Type: application/json` for JSON
+- `Content-Type: application/x-ndjson` for NDJSON
+- `Content-Type: text/csv` for CSV
+
+Only the [add documents](/reference/api/documents.md#add-or-replace-documents) and [update documents](/reference/api/documents.md#add-or-update-documents) endpoints accept  NDJSON and CSV. For all others, use `Content-Type: application/json`.
```
maryamsulemani97, line 85:
```suggestion
Only the [add documents](/reference/api/documents.md#add-or-replace-documents) and [update documents](/reference/api/documents.md#add-or-update-documents) endpoints accept NDJSON and CSV. For all others, use `Content-Type: application/json`.
```

guimachiavelli (approved):
🉑

maryamsulemani97:
bors merge

PR #1941, opened 2022-10-19
update settings page

### LINKED ISSUE #824: Update settings before adding documents
Updating the settings causes all documents to be reindexed. This is normal for a search engine database.

This means that the best practice for starting a new MeiliSearch index is to update the index settings first, then add documents. This halves RAM consumption, since you only index the documents once (on add), vs. if you add documents then update settings you are actually indexing your documents twice (1. on add 2. on settings update).

This info should be added to the docs.

maryamsulemani97:
The settings page says updating all settings will cause re-indexing. This is not true. 
Only updating these 5 settings triggers re-indexing: 
https://github.com/meilisearch/specifications/blob/main/text/0123-settings-api.md#41-triggering-documents-re-indexing

Also closes #824

### DIFF
--- a/reference/api/settings.md
+++ b/reference/api/settings.md
@@ -306,7 +306,7 @@ You can use this `taskUid` to get more details on [the status of the task](/refe
 The distinct attribute is a field whose value will always be unique in the returned documents.
 
 ::: warning
-Updating distinct attributes will re-index all documents in the index, which can take some time.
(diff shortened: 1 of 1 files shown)


guimachiavelli (approved):
⛸️

maryamsulemani97:
bors merge

PR #1950, opened 2022-10-23
Update settings API uses PATCH instead of PUT

teramotodaiki:
According to https://docs.meilisearch.com/reference/api/settings.html#update-settings , Update settings API is PATCH method. It is not PUT method.


maryamsulemani97 (approved):
Thanks @teramotodaiki! 🙏

bors merge

PR #1953, opened 2022-10-24
merge note and tip for stop words

maryamsulemani97:
We have a note, tip, and warning at the beginning of the [stop words](https://docs.meilisearch.com/reference/api/settings.html#stop-words) section of the API reference. This PR merges the content of the note and tip to make it less overwhelming.


maryamsulemani97:
bors merge

PR #1955, opened 2022-10-25
add markdown files as supported dataset format

### LINKED ISSUE #1819: add markdown files as supported dataset format
https://docs.meilisearch.com/learn/core_concepts/documents.html#dataset-format currently lists JSON, CSV NDJSON as the supported types for Meiisearch. 
In case you have data in Markdown file and you use Gatsby you can refer to the Gatsby Meilisearch plugin (https://www.gatsbyjs.com/plugins/gatsby-plugin-meilisearch/)
By using the Gatsby Meilisearch plugin, if we have dataset in markdown format, the data can be indexed in Meilisearch. 

Note: I understand this is a very peculiar case and could potentially be added as a note.

maryamsulemani97:
closes #1819

### DIFF
--- a/learn/core_concepts/documents.md
+++ b/learn/core_concepts/documents.md
@@ -63,6 +63,10 @@ Meilisearch accepts datasets in the following formats:
 - [NDJSON](#ndjson)
 - [CSV](#csv)
 
+::: note
+If you use the [Gatsby](https://github.com/meilisearch/gatsby-plugin-meilisearch/) plugin, Meilisearch accepts datasets in Markdown format.
+:::
+
 #### JSON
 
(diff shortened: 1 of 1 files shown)


guimachiavelli:
I'm not sure we should add this info to the docs.

If a plugin alters Meilisearch's default behaviour, then the altered behaviour should be documented on the plugin, no? This way we're protected against things like the plugin or Gatsby disappearing/becoming deprecated, updating the docs in case behaviour changes, and even from justifying why we talk about Gatsby but not about [another library/tool/SDK] that does something similar.

Additionally, I think the original issue misunderstands what's going on with Gatsby and Meilisearch. I might be wrong, but I think the plugin doesn't index markdown files directly. Instead, it seems to use GraphQL to query Gatsby and get data in an accepted format like JSON.

PR #1991, opened 2022-11-01
v.30: pass config file to meilisearch

### LINKED ISSUE #1962: v.30: Pass config file to Meilisearch
Users can now send Meilisearch a configuration file instead of command line options at launch

- Update the "[Customization](https://docs.meilisearch.com/learn/getting_started/quick_start.html#customization)" section of the quick start
- Add new section to the [Configuration](https://docs.meilisearch.com/learn/configuration/instance_options.html) guide

### References
- Issue on core: https://github.com/meilisearch/meilisearch/issues/2558, https://github.com/meilisearch/meilisearch/issues/2800, https://github.com/meilisearch/meilisearch/issues/2801
- Spec: https://github.com/meilisearch/specifications/pull/185
- SME: @curquiza

maryamsulemani97:
closes #1962


learn/configuration/instance_options.md
```diff
@@ -46,6 +137,15 @@ Environment variables are always identical to the corresponding command-line opt
 
 ## All instance options
 
+### Configuration file
```
curquiza, line 140: This option cannot be used in the config file. Maybe could we mention it here again?

learn/configuration/instance_options.md
```diff
@@ -4,10 +4,101 @@ sidebarDepth: 2
...
-You can configure Meilisearch at launch with **environment variables** and **command-line options**.
+You can configure Meilisearch at launch with **command-line options**, **environment variables**, or a **configuration file**.
 
 These startup options affect your entire Meilisearch instance, not just a single index. For settings that affect search within a single index, see [index settings](/learn/configuration/settings.md).
 
+## Configuration file
+
+Meilisearch accepts a configuration file in the `.toml` format as an alternative to command-line options and environment variables. Configuration files can be easily shared and versioned, and allow you to define multiple options.
+
+**Environment variables can overwrite the configuration file, and command-line options can overwrite environment variables.**
```
curquiza, line 15: I would add some context, like "using environment variables and config file at the same time would lead to..."

learn/configuration/instance_options.md
```diff
@@ -4,10 +4,101 @@ sidebarDepth: 2
...
+
+You can download the configuration file using the following command:
+
+<CodeSamples id="instance_options_config_file_download_1" />
+
+Once downloaded, Meilisearch will use it as the default configuration file. You will see it when you launch Meilisearch:
```
curquiza, line 21: I would write Meilisearch will read the `./config.toml` file by default, I mean the file in the working directory named `config.toml`

learn/configuration/instance_options.md
```diff
@@ -4,10 +4,101 @@ sidebarDepth: 2
...
+888Y88888P888  .d88b.  888 888 888 .d8888b   .d88b.   8888b.  888d888 .d8888b 88888b.
+888 Y888P 888 d8P  Y8b 888 888 888 88K      d8P  Y8b     "88b 888P"  d88P"    888 "88b
+888  Y8P  888 88888888 888 888 888 "Y8888b. 88888888 .d888888 888    888      888  888
+888   "   888 Y8b.     888 888 888      X88 Y8b.     888  888 888    Y88b.    888  888
+888       888  "Y8888  888 888 888  88888P'  "Y8888  "Y888888 888     "Y8888P 888  888
+
+Config file path:       "./config.toml"
+```
+
+If no configuration file is provided, the `Config file path` is set to `none`.
```
curquiza, line 36: I would put it at the end of the section, because if no `./config.toml` file are found OR if you haven't provided any file with `--config-file-path`, `none` is displayed.
This sentence only helps users for debugging but is not necessary to understand the feature.

learn/configuration/instance_options.md
```diff
@@ -4,10 +4,101 @@ sidebarDepth: 2
...
+
+### Configuration file format
+
+You can configure all environment variables and CLI options using a dedicated key in the configuration file. These keys follow the [snake case](https://en.wikipedia.org/wiki/Snake_case) convention. For example, `--import-dump` must be named `import_dump` in the file, `import_dumps` will throw an error.
+
+::: warning
+The only exception is the `config_file_path` key. Specifying the `config_file_path` key in the configuration file will throw an [error](/learn/configuration/instance_options.md#defining-a-configuration-file-with-a-syntax-error).
+:::
+
+By default, the file is divided into sections separated by comments. Each key has a brief description followed by a link to the relevant section of the documentation. Some keys are commented out as they are not required when launching Meilisearch, for example, `import_snapshot`, or because the value provided in the file is not the default value like `max_indexing_memory`.
```
curquiza, line 48: Hum, not sure this is useful to talk about the default file we provide. The users do not need to provide a file like this exactly to make it work. It's really confusing to me to explain this

learn/configuration/instance_options.md
```diff
@@ -4,10 +4,101 @@ sidebarDepth: 2
...
+
+### Configuration file errors
+
```
curquiza, line 51: Do we really need to define errors in the config file feature? Is it useful for the users? Looks to much like a specification, not a guide.

curquiza: My opinion is the full section can be removed, let me know what you think @maryamsulemani97 and @guimachiavelli

curquiza (changes requested):

learn/configuration/instance_options.md
maryamsulemani97: I added it here because I though it might be useful, but we can get rid of it

learn/configuration/instance_options.md
maryamsulemani97: It might be helpful for first-time users who just downloaded the file. Maybe we could add this part:
>Some keys are commented out as they are not required when launching Meilisearch, for example, `import_snapshot`, or because the value provided in the file is not the default value like `max_indexing_memory`.

to the config file, if you think it's useful

learn/configuration/instance_options.md
maryamsulemani97: Yes, good idea!

learn/configuration/instance_options.md
curquiza: From my user POV I would say it's not useful for me to know all the behavior of the features in case of errors. Errors should be explicit enough to guide me, but the docs don't need to describe them. It makes the docs heavier.
But I would rather have the opinion of gui, maybe he does not agree with me.

learn/configuration/instance_options.md
curquiza: The problem is we don't have the context here, we don't know we are talking about the default config file here.

learn/configuration/instance_options.md
```diff
@@ -12,13 +12,15 @@ These startup options affect your entire Meilisearch instance, not just a single
...
 
-**Environment variables can overwrite the configuration file, and command-line options can overwrite environment variables.**
+**If used simultaneously, environment variables can overwrite the configuration file, and command-line options can overwrite environment variables.**
 
 You can download the configuration file using the following command:
 
 <CodeSamples id="instance_options_config_file_download_1" />
 
-Once downloaded, Meilisearch will use it as the default configuration file. You will see it when you launch Meilisearch:
+This default configuration file is divided into sections separated by comments. Each key has a brief description followed by a link to the relevant section of the documentation. Some keys are commented out as they are not required when launching Meilisearch, for example, `import_snapshot`, or because the value provided in the file is not the default value like `max_indexing_memory`.
```
curquiza, line 21: "Some keys are commented out as they are not required when launching Meilisearch" => No key are required when launching Meilisearch.

From my POV this is not really useful to "justify" our default config file, but only my opinion

.code-samples.meilisearch.yaml
```diff
@@ -1038,3 +1038,5 @@ date_guide_sort_1: |
       "q": "",
       "sort": ["release_timestamp:desc"]
     }'
+instance_options_config_file_download_1: |
+  curl https://raw.githubusercontent.com/meilisearch/meilisearch/main/config.toml > config.toml
```
guimachiavelli, line 1042: Unless I'm mistaken, instance options cannot be configured with an SDK, so perhaps it's better to leave it outside the code sample component?

@brunoocasali, what do you think?

learn/configuration/instance_options.md
```diff
@@ -4,10 +4,99 @@ sidebarDepth: 2
...
 # Configure Meilisearch at launch
 
-You can configure Meilisearch at launch with **environment variables** and **command-line options**.
+You can configure Meilisearch at launch with **command-line options**, **environment variables**, or a **configuration file**.
 
 These startup options affect your entire Meilisearch instance, not just a single index. For settings that affect search within a single index, see [index settings](/learn/configuration/settings.md).
 
+## Configuration file
+
+Meilisearch accepts a configuration file in the `.toml` format as an alternative to command-line options and environment variables. Configuration files can be easily shared and versioned, and allow you to define multiple options.
```
guimachiavelli, line 13: I find it a bit weird to say that configuration files are alternatives to cli/env vars, but have it as the first configuration method. I suggest we move it below environment variables.

learn/configuration/instance_options.md
```diff
@@ -4,10 +4,99 @@ sidebarDepth: 2
...
+
+**If used simultaneously, environment variables can overwrite the configuration file, and command-line options can overwrite environment variables.**
```
guimachiavelli, line 15: I recommend dropping the "can"s so statements are simple and leave no room for interpretation: env vars overwrite config files. You might also be able to get away with dropping the first clause, but it's your call.

```suggestion
**When used simultaneously, environment variables overwrite the configuration file, and command-line options overwrite environment variables.**
```

learn/configuration/instance_options.md
guimachiavelli: I believe "key" here stands for "option", no? In which case, I recommend saying option to avoid misunderstandings.

In any case, I tend to agree with with @curquiza: not sure this paragraph is necessary since we're only describing the file.

learn/configuration/instance_options.md
```diff
@@ -4,10 +4,99 @@ sidebarDepth: 2
...
+
+You can download the configuration file using the following command:
+
+<CodeSamples id="instance_options_config_file_download_1" />
+
+This default configuration file is divided into sections separated by comments. Each key has a brief description followed by a link to the relevant section of the documentation. Some keys are commented out as they are not required when launching Meilisearch, for example, `import_snapshot`, or because the value provided in the file is not the default value like `max_indexing_memory`.
+
+By default, Meilisearch will use the file `./config.toml` in the working directory as the default configuration file. You will see it when you launch Meilisearch:
```
guimachiavelli, line 23: I think we can drop the article. Also, we don't need to use both `./` and working directory, in my opinion because the working directory is `./`.

```suggestion
By default, Meilisearch will use the `config.toml` in the working directory as the default configuration file. You will see it when you launch Meilisearch:
```

learn/configuration/instance_options.md
```diff
@@ -4,10 +4,99 @@ sidebarDepth: 2
...
+888       888  "Y8888  888 888 888  88888P'  "Y8888  "Y888888 888     "Y8888P 888  888
+
+Config file path:       "./config.toml"
+```
+
+You can override the default configuration file using the `MEILI_CONFIG_FILE_PATH` environment variable or the `--config-file-path` CLI option.
+
+### Configuration file format
+
+You can configure all environment variables and CLI options using a dedicated key in the configuration file. These keys follow the [snake case](https://en.wikipedia.org/wiki/Snake_case) convention. For example, `--import-dump` must be named `import_dump` in the file, `import_dumps` will throw an error.
```
guimachiavelli, line 42: Last bit is unnecessary and doesn't connect with the snake_case convention, imo. `import_dumps` throws an error not because it's not following the convention, but because that option doesn't exist.

```suggestion
You can configure all environment variables and CLI options using a dedicated key in the configuration file. These keys follow the [snake case](https://en.wikipedia.org/wiki/Snake_case) convention. For example, `--import-dump` must be named `import_dump` in the file.
```

learn/configuration/instance_options.md
```diff
@@ -4,10 +4,99 @@ sidebarDepth: 2
...
+
+::: warning
+The only exception is the `config_file_path` key. Specifying the `config_file_path` key in the configuration file will throw an [error](/learn/configuration/instance_options.md#defining-a-configuration-file-with-a-syntax-error).
```
guimachiavelli, line 45: The first sentence lacks context: the only exception to what? the snake_case bit and the warning box make create too much distance between this statement and  "You can configure all environment variables".

```suggestion
Specifying the `config_file_path` key in the configuration file will throw an [error](/learn/configuration/instance_options.md#defining-a-configuration-file-with-a-syntax-error).
```

learn/configuration/instance_options.md
guimachiavelli: I agree with @curquiza, I also think this section makes the article too long. All the errors seem to work pretty much like other Meilisearch errors, so there's nothing unexpected that justifies talking about them here, I think.

learn/getting_started/quick_start.md
```diff
@@ -279,6 +279,10 @@ These allow you to configure your Meilisearch instance at launch with [environme
 
 **Instance options affect your entire Meilisearch instance**, not just a single index.
 
+### Configuration file
```
guimachiavelli, line 282: The config file only accepts instance options, right? If so, I think it's better to have it as subheading of `### Instance options` instead of a self-sufficient section like `### Index settings`.

```suggestion
#### Configuration file
```

guimachiavelli (changes requested):

learn/getting_started/quick_start.md
maryamsulemani97: It is a subheading of `Instance options` 👀

guimachiavelli: Oh, but aren't both of them on the same heading level?

`### Instance options` on L#276

`### Configuration file` on L#282

Both are level three (`###`), but for configuration file to be a subheading it would have to be level four (`####`), no?

maryamsulemani97: I get it now! Yes, you're right 🙏

learn/configuration/instance_options.md
```diff
@@ -44,8 +44,84 @@ Environment variables for command-line flags accept `n`, `no`, `f`, `false`, `of
...
+888Y88888P888  .d88b.  888 888 888 .d8888b   .d88b.   8888b.  888d888 .d8888b 88888b.
+888 Y888P 888 d8P  Y8b 888 888 888 88K      d8P  Y8b     "88b 888P"  d88P"    888 "88b
+888  Y8P  888 88888888 888 888 888 "Y8888b. 88888888 .d888888 888    888      888  888
+888   "   888 Y8b.     888 888 888      X88 Y8b.     888  888 888    Y88b.    888  888
+888       888  "Y8888  888 888 888  88888P'  "Y8888  "Y888888 888     "Y8888P 888  888
+
+Config file path:       "./config.toml"
+```
+
+You can override the default configuration file using the `MEILI_CONFIG_FILE_PATH` environment variable or the `--config-file-path` CLI option:
```
guimachiavelli, line 72:
```suggestion
You can override the default location of the configuration file using the `MEILI_CONFIG_FILE_PATH` environment variable or the `--config-file-path` CLI option:
```

guimachiavelli (changes requested):

learn/configuration/instance_options.md
```diff
@@ -44,8 +44,84 @@ Environment variables for command-line flags accept `n`, `no`, `f`, `false`, `of
...
+::: tab Environment variable
+
+Linux/MacOS:
+
+```sh
+export MEILI_CONFIG_FILE_PATH="./config.toml"
+./meilisearch
+```
+
+Windows:
```
curquiza, line 92: Just by curiosity? Why do we now explain how to set an environment variable with Windows since we don't do it [here](https://docs.meilisearch.com/learn/configuration/instance_options.html#environment-variables)?

maryamsulemani97: Good catch. I'll update it here too

.code-samples.meilisearch.yaml
brunoocasali: Yes, that's correct! The SDK uses an already-initialized instance :)

learn/getting_started/quick_start.md
```diff
@@ -279,6 +279,10 @@ These allow you to configure your Meilisearch instance at launch with [environme
 
 **Instance options affect your entire Meilisearch instance**, not just a single index.
 
+#### Configuration file
+
+Instead of providing command-line options or environment variables, you can provide a `.toml` [configuration file](/learn/configuration/instance_options.md#configuration-file). Once downloaded, move it to your working directory and it will be loaded at launch. Like instance options, it affects your entire instance.
```
guimachiavelli, line 284: Trying to avoid repeating the same verb twice in quick succession.

```suggestion
Instead of providing command-line options or environment variables, you can use a `.toml` [configuration file](/learn/configuration/instance_options.md#configuration-file). Once downloaded, move it to your working directory and it will be loaded at launch. Like instance options, it affects your entire instance.
```

learn/configuration/instance_options.md
```diff
@@ -32,20 +32,114 @@ The above flag disables analytics for the Meilisearch instance and does not acce
 
 To configure a Meilisearch instance using environment variables, set the environment variable prior to launching the instance. If you are unsure how to do this, read more about [setting and listing environment variables](https://linuxize.com/post/how-to-set-and-list-environment-variables-in-linux/), or [use a command-line option](#command-line-options-and-flags) instead.
 
-```bash
+:::: tabs
+::: tab Linux/MacOS
```
guimachiavelli, line 36: Would it be more accurate/inclusive to say UNIX instead of Linux? What do you think, @curquiza?

learn/configuration/instance_options.md
```diff
@@ -32,20 +32,114 @@ The above flag disables analytics for the Meilisearch instance and does not acce
...
+set MEILI_CONFIG_FILE_PATH="./config.toml"
+./meilisearch
+```
+
+:::
+::::
+
+### Configuration file format
+
+You can configure all environment variables and CLI options using a dedicated option in the configuration file. These options follow the [snake case](https://en.wikipedia.org/wiki/Snake_case) convention. For example, `--import-dump` must be named `import_dump` in the file.
```
guimachiavelli, line 122: What do you think about removing the "using a dedicated option" bit?

```suggestion
You can configure all environment variables and CLI options using the configuration file. These options must be written in [snake case](https://en.wikipedia.org/wiki/Snake_case). For example, `--import-dump` must be named `import_dump` in the file.
```

guimachiavelli (changes requested):
Ok, sorry, I think this is the last review.

learn/configuration/instance_options.md
maryamsulemani97: Note to self: Update [this](https://docs.meilisearch.com/learn/security/master_api_keys.html#protecting-a-meilisearch-instance) if we go with UNIX

learn/configuration/instance_options.md
maryamsulemani97: We're talking about the file format here 🤔  So wouldn't it be better to talk about "options"? We don't mention them anywhere else.

learn/configuration/instance_options.md
curquiza: Yes, in this case `Unix` is better

curquiza (approved):

learn/advanced/updating.md
```diff
@@ -141,7 +141,7 @@ If you're unsure where your Meilisearch directory is located, try this:
 
 :::: tabs
 
-::: tab macOS / Linux
+::: tab UNIX / macOS
```
curquiza, line 144: Unix, includes Macos

curquiza, line 144:
```suggestion
::: tab UNIX
```

curquiza (changes requested):

learn/advanced/updating.md
maryamsulemani97: Thanks @curquiza! 🙏

learn/configuration/instance_options.md
```diff
@@ -32,20 +32,114 @@ The above flag disables analytics for the Meilisearch instance and does not acce
...
+
+**When used simultaneously, environment variables overwrite the configuration file, and command-line options overwrite environment variables.**
+
+You can download the configuration file using the following command:
+
+```sh
+curl https://raw.githubusercontent.com/meilisearch/meilisearch/main/config.toml > config.toml
+```
+
+By default, Meilisearch will use `config.toml` in the working directory as the default configuration file. You will see it when you launch Meilisearch:
```
dichotommy, line 75: What if you launch Meilisearch w/o a configuration file, does the `Config file path` field still appear in the startup message? If so, what is its value?

```suggestion
By default, Meilisearch will look for a `config.toml` file in the working directory. If it is present, it will be used as the configuration file. You can verify this when you launch Meilisearch:
```

learn/configuration/instance_options.md
```diff
@@ -32,20 +32,114 @@ The above flag disables analytics for the Meilisearch instance and does not acce
...
+:::: tabs
+::: tab CLI
+
+```sh
+./meilisearch --config-file-path="./config.toml"
+```
+
+:::
+
+::: tab Environment variable
```
dichotommy, line 101:
```suggestion
::: tab Env
```

learn/configuration/instance_options.md
dichotommy: I agree w/ Gui's suggestion, and would go a little further.

I also think that after this paragraph we should provide an example of a full line from a config file. Of course, anyone curious enough can look at the default config file we provide and figure it out from that... but people are lazy! This should make the formatting completely clear.

````suggestion
You can configure any environment variable or CLI option using a configuration file. In configuration files, options must be written in [snake case](https://en.wikipedia.org/wiki/Snake_case). For example, `--import-dump` would be written as `import_dump`.

```toml
import_dump = "./example.dump"
```
````

File: learn/configuration/instance_options.md [hunk 23]
```diff
@@ -32,20 +32,114 @@ The above flag disables analytics for the Meilisearch instance and does not acce
...
+
+:::
+::::
+
+### Configuration file format
+
+You can configure all environment variables and CLI options using a dedicated option in the configuration file. These options follow the [snake case](https://en.wikipedia.org/wiki/Snake_case) convention. For example, `--import-dump` must be named `import_dump` in the file.
+
+::: warning
+Specifying the `config_file_path` option in the configuration file will throw an error.
```
dichotommy, line 125:
```suggestion
Specifying the `config_file_path` option within the configuration file will throw an error. This is the only configuration option that cannot be set within a configuration file.
```

learn/configuration/instance_options.md
```diff
@@ -32,20 +32,114 @@ The above flag disables analytics for the Meilisearch instance and does not acce
...
+
+```sh
+set MEILI_CONFIG_FILE_PATH="./config.toml"
+./meilisearch
+```
+
+:::
+::::
+
+### Configuration file format
```
dichotommy, line 120: "file format" = `jpg`, `toml`, etc.
"file formatting" = rules for arranging text
TMYK 😉 

```suggestion
### Configuration file formatting
```

learn/configuration/instance_options.md
```diff
@@ -32,20 +32,114 @@ The above flag disables analytics for the Meilisearch instance and does not acce
...
+
+**Environment variable**: `MEILI_CONFIG_FILE_PATH`
+**CLI option**: `--config-file-path`
+**Default**: `./config.toml`
+**Expected value**: a filepath
+
+Designates the location of the configuration file to load at launch.
+
+::: warning
+Specifying this option in the configuration file will throw an error.
```
dichotommy, line 140:
```suggestion
Specifying this option in the configuration file itself will throw an error (assuming Meilisearch is able to find your configuration file).
```

learn/getting_started/quick_start.md
```diff
@@ -279,6 +279,10 @@ These allow you to configure your Meilisearch instance at launch with [environme
 
 **Instance options affect your entire Meilisearch instance**, not just a single index.
 
+#### Configuration file
+
+Instead of providing command-line options or environment variables, you can use a `.toml` [configuration file](/learn/configuration/instance_options.md#configuration-file). Once downloaded, move it to your working directory and it will be loaded at launch. Like instance options, it affects your entire instance.
```
dichotommy, line 284: It's not necessary to download a premade configuration file, since users can also easily construct their own based on the formatting rules we provide.

```suggestion
As an alternative to providing command-line options or environment variables at launch, you can configure your Meilisearch instance using a [configuration file](/learn/configuration/instance_options.md#configuration-file) in `.toml` format. If a configuration file is present at the root of your working directory, it will be loaded at launch.
```

curquiza (approved):

curquiza:
Nice guide @maryamsulemani97 👏

learn/configuration/instance_options.md
maryamsulemani97: We use "Environment variable" [here](https://docs.meilisearch.com/learn/security/master_api_keys.html#protecting-a-meilisearch-instance)

dichotommy: I see. Please ignore my suggestion then—it's just some needless optimization 😅

learn/configuration/instance_options.md
```diff
@@ -32,20 +32,118 @@ The above flag disables analytics for the Meilisearch instance and does not acce
...
+88888b.d88888              888                                                888
+888Y88888P888  .d88b.  888 888 888 .d8888b   .d88b.   8888b.  888d888 .d8888b 88888b.
+888 Y888P 888 d8P  Y8b 888 888 888 88K      d8P  Y8b     "88b 888P"  d88P"    888 "88b
+888  Y8P  888 88888888 888 888 888 "Y8888b. 88888888 .d888888 888    888      888  888
+888   "   888 Y8b.     888 888 888      X88 Y8b.     888  888 888    Y88b.    888  888
+888       888  "Y8888  888 888 888  88888P'  "Y8888  "Y888888 888     "Y8888P 888  888
+
+Config file path:       "./config.toml"
+```
+
```
dichotommy, line 89:
```suggestion

If the `Config file path` is anything other than `"none"`, it means that a config file was successfully located and used to start Meilisearch.

```

learn/configuration/instance_options.md
```diff
@@ -32,20 +32,118 @@ The above flag disables analytics for the Meilisearch instance and does not acce
...
+import_dump = "./example.dump"
+```
+
+::: warning
+Specifying the `config_file_path` option within the configuration file will throw an error. This is the only configuration option that cannot be set within a configuration file.
+:::
+
 ## All instance options
 
+### Configuration file
```
dichotommy, line 134:
```suggestion
### Configuration file path
```

dichotommy (approved):
Added two small comments, let me know if you need another approval after committing 🏂🏻 ⛰️

maryamsulemani97:
bors merge

PR #2018, opened 2022-11-28
v0.30: update quick start+instance options

maryamsulemani97:
Updates quick start+ instance options


dichotommy (commented):
Thanks for fixing that in instance options.

Could you take another look at the customization section? A few comments:

- Just an idea, but perhaps instead of giving configuration options as bullet points and using separate paragraphs for customization options, we could give the three customization options as bullet points (i.e. index settings, search parameters, instance options).
  - Communicating the different types of customization possible is more important than specifics of how to customize.
- Rewrite intro paragraph so it's less redundant w/ following content

Thanks!

dichotommy (approved):
I'm approving this for scheduling reasons (would like to approve the v0.30 PR before we start our meeting at 2pm). I think we can make further changes once it's live on the site 👌🏻

maryamsulemani97:
bors merge

PR #2023, opened 2022-11-29
Fix index swap code sample json payload

### LINKED ISSUE #2022: Invalid swap indexes example
Hey team 👋,

[There is a little issue in the example of the index swap](https://docs.meilisearch.com/reference/api/indexes.html#example-6). The JSON body is wrong as it is an object with two _indexes_ key. It should be of [the form that you can find in the spec](https://github.com/meilisearch/specifications/blob/main/text/0191-swap-indexes-api.md#3411-payload-definition).

Have a nice day 😃

PS: I stole the issue 2022, don't be mad...

dichotommy:
Fixes #2022 

Thanks @Kerollmops for spotting it! 🙌🏻

### DIFF
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -1077,17 +1077,19 @@ cancel_tasks_1: |
     -X POST 'http://localhost:7700/tasks/cancel?uids=1,2'
 swap_indexes_1: |
   curl \
-    -X POST 'http://localhost:7700/swap-indexes'\
+    -X POST 'http://localhost:7700/swap-indexes' \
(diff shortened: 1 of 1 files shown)


curquiza (approved):

curquiza:
Only my opinion, but it's easier for me to read in a single line like in the specification, but again only my opinion 😊

dichotommy:
@curquiza I think I agree in this case. But we try to follow a consistent style, and in other cases it's harder to read (e.g. ranking rules array). Thanks for the feedback en tout cas! 😄

dichotommy:
bors merge

PR #2024, opened 2022-11-29
Replace Slack links + mentions w/ discord

dichotommy:
Note that I did not replace the "Public Slack community size" row in the comparison to alternatives. I think we should probably wait til our Discord server fills up a bit then change it to just say "Public community size" and replace the number.


maryamsulemani97 (commented):
We should update the readme (line 91) as well

ferdi05 (approved):
All lgtm, thanks @dichotommy and good catch @maryamsulemani97

guimachiavelli:
bors merge

PR #2026, opened 2022-12-01
Use the new latest git tag instead of the stable branch

curquiza:
We don't use `stable` git branch anymore, but the `latest` git tag 😇


dichotommy:
bors merge

PR #2029, opened 2022-12-05
Improve messaging around large datasets

dichotommy:
Early in Meilisearch's development, the engine was not equipped to index or search through very large datasets. Since then, the engine team has done a lot of work on indexation and search performance. Today, we can say with confidence that Meilisearch is effective even on large datasets.

For reviewers: I tried to find any area where we imply that Meilisearch is not suited for large datasets, but I may have missed some. Let me know if you spot an area that should be improved.

In response to [this thread](https://meilisearch.slack.com/archives/C010VV3SHQF/p1669912907905059) (private link)

### DIFF
--- a/learn/what_is_meilisearch/comparison_to_alternatives.md
+++ b/learn/what_is_meilisearch/comparison_to_alternatives.md
@@ -151,14 +151,15 @@ Can't find a client you'd like us to support? [Submit your idea or vote for it](
 
 ### Meilisearch vs Elasticsearch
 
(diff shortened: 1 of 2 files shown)


dureuill (approved):
Looks good to me, thanks dichotommy!

> For reviewers: I tried to find any area where we imply that Meilisearch is not suited for large datasets, but I may have missed some. Let me know if you spot an area that should be improved.

I grep'd "large" and "many" in the documentation repository but could not find any other relevant location to change.

ferdi05 (approved):
lgtm, thanks @dichotommy

PR #2030, opened 2022-12-06
Fix routeHighlighter content for cancel + delete tasks

dichotommy:
Both of these endpoints require at least one query parameter, but it doesn't necessarily have to be a `task_uid`. In fact, if you try to call these endpoints giving only a task uid for parameters, it will throw a deserialize error. Task uids must follow the `uids` query parameter, e.g. `?uids=5,6`

Related to [this thread](https://meilisearch.slack.com/archives/C010VV3SHQF/p1670322127291219) (private link)

### DIFF
--- a/reference/api/tasks.md
+++ b/reference/api/tasks.md
@@ -319,7 +319,7 @@ If you try retrieving a deleted task, Meilisearch will return a [`task_not_found
 
 ## Cancel tasks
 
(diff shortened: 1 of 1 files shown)


gmourier (approved):
LGTM 🙇‍♂️

maryamsulemani97 (approved):
bors merge

PR #2031, opened 2022-12-06
Make name of section similar to name of option

dichotommy:
Small, simple change to bring the `Dumps location` section of the instance options page closer to the name of the actual option (`dumps-dir`).


guimachiavelli (approved):
🧛

PR #2032, opened 2022-12-06
No /indexes in swap-indexes endpoint route

dichotommy:
Mistake in the `routeHighlighter` element for swap indexes; `/indexes` should not be part of the route.


bidoubiwa (approved):
Very nice 🔥

dichotommy:
bors merge

PR #2033, opened 2022-12-06
Clarify swap indexes payload format

dichotommy:
Payload format was incorrect, so I fixed it, and while I was at it cleaned up the explanation a bit.


CaroFG (approved):

dichotommy:
bors merge

PR #2040, opened 2022-12-12
add vale rule for latin terms

### LINKED ISSUE #1974: Remove `e.g.` and `i.e` from the docs
- [x] We should replace `e.g.` with `for example` and `i.e.` with `meaning` or `that is` as everyone does not understand them
 - [ ] Create a Vale rule to throw an error when `e.g.` or `i.e.` is used

maryamsulemani97:
closes #1974 

Adds Vale rule to replace `i.e` and `e.g` with `that is` and `for example`, respectively. 

## For reviewers
`that is` isn't the best replacement for `i.e`, but most style guides use it. I did add `consider rewriting the sentence` to the error message. Should we go with `meaning` instead?

### DIFF
--- a/.vale/styles/style_guide/LatinTerms.yml
+++ b/.vale/styles/style_guide/LatinTerms.yml
@@ -1,0 +1,17 @@
+---
+# Error: style_guide.LatinTerms.yml
+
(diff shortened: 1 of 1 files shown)


maryamsulemani97:
bors merge

PR #2043, opened 2022-12-13
comparison of implicit vs. explicit index creation

### LINKED ISSUE #873: Add comparison of implicit vs. explicit index creation
In the main concept guide, we should definitely add an implicit vs. explicit comparison section similar to [MongoDB's](https://docs.mongodb.com/manual/core/databases-and-collections/#create-a-collection).

- Implicit: default option, simple, less code.
- Explicit: much more error-proof. This is because implicit index creation is actually two actions (1. Create index, 2. Add documents to said index) tied to one command (add documents). Unlike all other actions in MeiliSearch which are atomic, it is possible for only one of these two actions to complete successfully (index is created, but documents are not added). This can cause errors or missing documents.

Inspired by #583

maryamsulemani97:
closes #873

### DIFF
--- a/learn/core_concepts/indexes.md
+++ b/learn/core_concepts/indexes.md
@@ -18,11 +18,15 @@ One index's settings do not impact other indexes. For example, you could use a d
 
 ## Index creation
 
+### Implicit index creation
+
 If you try to add documents or settings to an index that does not already exist, Meilisearch will automatically create it for you.
 
 ### Explicit index creation
 
-You can explicitly create an index with custom search settings using the [create index endpoint.](/reference/api/indexes.md#create-an-index) Once created, you can add documents using the [add documents endpoint.](/reference/api/documents.md#add-or-update-documents) Though accomplished using a single command, the index creation and document addition actions are separate. Hence, it is possible that one of them might not complete successfully. This can cause errors or missing documents.
+You can explicitly create an index using the [create index endpoint](/reference/api/indexes.md#create-an-index). Once created, you can add documents using the [add documents endpoint](/reference/api/documents.md#add-or-update-documents). 
+
+While implicit index creation is more convenient, requiring only a single API request, **explicit index creation is considered safer for production**. This is because implicit index creation bundles multiple actions into a single task. If one action completes successfully while the other fails, the problem can be difficult to diagnose.
 
 ## Index UID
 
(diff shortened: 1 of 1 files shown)


learn/core_concepts/indexes.md
```diff
@@ -18,7 +18,11 @@ One index's settings do not impact other indexes. For example, you could use a d
 
 ## Index creation
 
-If you try to add documents or settings to an index that does not already exist, Meilisearch will automatically create it for you. You can also create indexes manually using the [create index endpoint](/reference/api/indexes.md#create-an-index).
+If you try to add documents or settings to an index that does not already exist, Meilisearch will automatically create it for you.
```
dichotommy, line 21: Using this phrase makes it a bit easier to describe the differences between the two options.

```suggestion
### Implicit index creation

If you try to add documents or settings to an index that does not already exist, Meilisearch will automatically create it for you.
```

learn/core_concepts/indexes.md
```diff
@@ -18,7 +18,11 @@ One index's settings do not impact other indexes. For example, you could use a d
...
+
+### Explicit index creation
+
+You can explicitly create an index with custom search settings using the [create index endpoint.](/reference/api/indexes.md#create-an-index) Once created, you can add documents using the [add documents endpoint.](/reference/api/documents.md#add-or-update-documents) Though accomplished using a single command, the index creation and document addition actions are separate. Hence, it is possible that one of them might not complete successfully. This can cause errors or missing documents.
```
dichotommy, line 25:
```suggestion
You can explicitly create an index using the [create index endpoint](/reference/api/indexes.md#create-an-index). Once created, you can add documents using the [add documents endpoint](/reference/api/documents.md#add-or-update-documents). 

While implicit index creation is more convenient, requiring only a single API request, **explicit index creation is considered safer for production**. This is because implicit index creation bundles multiple actions into a single task. If one action completes successfully while the other fails, the problem can be difficult to diagnose.
```

dichotommy (changes requested):
While it's true that implicit index creation "can cause errors or missing documents", this phrasing could scare users away from the feature. We don't want that! I proposed changes aiming to illustrate the benefits of using one vs. the other, while gently pushing users towards explicit index creation as the safer option.

learn/core_concepts/indexes.md
maryamsulemani97: This is more helpful. Thanks 🤡

dichotommy (approved):
LGTM 🍨

maryamsulemani97:
bors merge

PR #2044, opened 2022-12-13
Remove`learn/configuration/settings.md`

### LINKED ISSUE #1614: Remove/simplify `learn/configuration/settings.md`
We currently have three separate spots where users can find information about a setting:
1. `learn/configuration/settings.md`
2. `reference/api/settings.md`
3. one article for each index setting under `reference/api/`

I believe this is bad because: 
1. how can a user who searches for e.g. `displayedAttributes` which one of these pages is more likely to have the information they seek?
2. problem #1 is arguably mitigated if all three possible pages with info on a setting share the same content.  But, if they share the same content, there is no point in having these three pages;
3. maintenance and updates become problematic: the same way a user needs to guess where the info they seek is, we need to guess where to place a certain piece of info.

From these pages, `learn/configuration/settings.md` seems like the least useful right now. It's neither a reference, nor a guide. At first glance it looks a bit useless and out of sync with our recent updates, but it's possible there's some unique content to it.

maryamsulemani97:
closes #1614 

## Update

- We decided to go with 3 
## What is this PR doing?
- Removes all unused code samples
- Updates links from `/learn/configuration/settings.md` to API reference 
- Uses the same table in the API reference and `/learn/configuration/settings.md`
- Adds "developer-friendly syntax" to the "Body" of each setting in the API reference
~- Moves "developer-friendly syntax"(`displayedAttributes=[<String>, <String>, ...]`) to a new column "Format" in the table in `/learn/configuration/settings.md`~
  ~- I don't like how the table looks because of the "Format" column, and "Type" and "Format" are talking about the same thing:~
     ~- Should we remove "Type"? I think it's a lot easier to process compared to "Format"~
     ~- Orrr we could have the same table as the[ API reference](https://docs.meilisearch.com/reference/api/settings.html#body)? It will be easier to update too. Update one table and copy it to the other~
     ~WDYT @guimachiavelli @dichotommy?~

<hr>

1. We get rid of `/learn/configuration/settings.md`

Screen Shot 2022-12-13 at 15 46 40
  
  - The guides seem to pop out of nowhere. We could rename the "Index settings" heading to something more suitable.
2. Only keep `/learn/configuration/instance_options.md` under Configuration. Move the guides to "Advanced topics"

Screen Shot 2022-12-13 at 15 54 12

 - The guides are very different from the current content under "Advanced topics". Should we create a new heading?
 - Makes it look like there isn't any content under "Configuration"

3. We only keep the ["Index settings"](/learn/configuration/settings.html#index-settings) table in  `/learn/configuration/settings.md`
- The reader will get a summary of the different settings without having to go to another page, and we can keep the guides here


learn/configuration/instance_options.md
```diff
@@ -6,7 +6,7 @@ sidebarDepth: 2
 
 You can configure Meilisearch at launch with **command-line options**, **environment variables**, or a **configuration file**.
 
-These startup options affect your entire Meilisearch instance, not just a single index. For settings that affect search within a single index, see [index settings](/learn/configuration/settings.md).
+These startup options affect your entire Meilisearch instance, not just a single index. For settings that affect search within a single index, see [index settings](/reference/api/settings.md).
```
dichotommy, line 9: I think it makes more sense to link to the overview table here, since they are in the same site section (configuration).

learn/core_concepts/documents.md
```diff
@@ -24,7 +24,7 @@ Every field has a data type dictated by its value. Every value must be a valid [
 
 If a field contains an object, Meilisearch flattens it during indexing using dot notation and brings the object's keys and values to the root level of the document itself. This flattened object is only an intermediary representation—you will get the original structure upon search. You can read more about this in our [dedicated guide](/learn/advanced/datatypes.md#objects).
 
-With [ranking rules](/learn/core_concepts/relevancy.md#ranking-rules), you can decide what fields are more relevant than others. For example, you may decide recent movies should be more relevant than older ones. You can also configure how Meilisearch handles certain fields at an [index level](/learn/configuration/settings.md) in the settings.
+With [ranking rules](/learn/core_concepts/relevancy.md#ranking-rules), you can decide what fields are more relevant than others. For example, you may decide recent movies should be more relevant than older ones. You can also configure how Meilisearch handles certain fields at an [index level](/reference/api/settings.md) in the settings.
```
dichotommy, line 27: Did not think the link was adding anything here—distracting, if anything—so replaced it with a segue to the next section.

Also, not the right PR to fix this... but this is a bad example! Making recent movies more relevant than older ones is an example of sorting on a single field. "Deciding which fields are more relevant than others," in the first sentence, alludes to attribute ranking order.

I may ask you to review this guide next week @maryamsulemani97 .

```suggestion
With [ranking rules](/learn/core_concepts/relevancy.md#ranking-rules), you can decide which fields are more relevant than others. For example, you may decide recent movies should be more relevant than older ones. You can also designate certain fields as displayed or searchable.
```

learn/getting_started/algolia_migration.md
```diff
@@ -160,15 +160,15 @@ const BATCH_SIZE = 1000;
 
 Meilisearch's default settings are designed to deliver a fast and relevant search experience that works for most use-cases.
 
-To customize your index settings, we recommend following [this guide](/learn/configuration/settings.md). To learn more about the differences between settings in Algolia and Meilisearch, read on.
+To customize your index settings, we recommend following [this guide](/reference/api/settings.md). To learn more about the differences between settings in Algolia and Meilisearch, read on.
```
dichotommy, line 163: API reference ≠ a guide, replace with link to settings section of index core concepts page.

In general, when linking from the getting started, we should prefer "explain"-y guides like those in the Learn section. There are of course exceptions, like the below example.

learn/getting_started/filtering_and_sorting.md
```diff
@@ -18,7 +18,7 @@ The `settings` object contains two arrays for this purpose: `filterableAttribute
 
 The above code sample adds `mass` and `_geo` to `filterableAttributes` and `sortableAttributes` setting you up for the next sections.
 
-To learn more about the `settings` object and how to configure it, refer to our [dedicated guide](/learn/configuration/settings.md).
+To learn more about the `settings` object and how to configure it, refer to our [dedicated guide](/reference/api/settings.md).
```
dichotommy, line 21: "Dedicated guide" is not just go-to link text; it actually means something. We want our users to understand what they're clicking on.

```suggestion
To learn more about the `settings` object and how to configure it, check out the [API reference](/reference/api/settings.md).
```

learn/security/tenant_tokens.md
```diff
@@ -16,7 +16,7 @@ Tenant tokens are small packages of encrypted data presenting proof a user can a
 
 To use tokens in Meilisearch, you only need to have a system for token generation in place. The quickest method to generate tenant tokens is [using one of our official SDKs](#generating-tenant-tokens-with-an-sdk). It is also possible to [generate a token from scratch](#generating-tenant-tokens-without-a-meilisearch-sdk).
 
-Tenant tokens do not require you to configure any specific [instance options](/learn/configuration/instance_options.md) or [index settings](/learn/configuration/settings.md). They are also meant to be short-lived—Meilisearch does not store nor keeps track of generated tokens.
+Tenant tokens do not require you to configure any specific [instance options](/learn/configuration/instance_options.md) or [index settings](/reference/api/settings.md). They are also meant to be short-lived—Meilisearch does not store nor keeps track of generated tokens.
```
dichotommy, line 19:
```suggestion
Tenant tokens do not require you to configure any specific [instance options](/learn/configuration/instance_options.md) or [index settings](/learn/configuration/settings.md). They are also meant to be short-lived—Meilisearch does not store nor keeps track of generated tokens.
```

reference/api/settings.md
```diff
@@ -801,13 +821,23 @@ Update the ranking rules of an index.
...
 
 To create a custom ranking rule, give an attribute followed by a colon (`:`) and either `asc` for ascending order or `desc` for descending order.
 
 - To apply an **ascending sort** (results sorted by increasing value): `attribute_name:asc`
 - To apply a **descending sort** (results sorted by decreasing value): `attribute_name:desc`
 
+::: warning
+If some documents do not contain the attribute defined in a custom ranking rule, the application of the ranking rule is undefined and the search results might not be sorted as you expected.
+
+We recommend that all your documents contain any attribute used in a custom ranking rule. For example, if you set the custom ranking rule `desc(year)`, make sure that all your documents contain the attribute `year`.
```
dichotommy, line 838:
```suggestion
Make sure that any attribute used in a custom ranking rule is present in all of your documents. For example, if you set the custom ranking rule `desc(year)`, make sure that all your documents contain the attribute `year`.
```

dichotommy (changes requested):
I really like the addition of the developer-friendly syntax to the settings API reference. I think it adds a lot in terms of drawing the reader's eye.

learn/configuration/instance_options.md
maryamsulemani97, line 9:
```suggestion
These startup options affect your entire Meilisearch instance, not just a single index. For settings that affect search within a single index, see [index settings](/learn/configuration/settings.md).
```

learn/getting_started/algolia_migration.md
maryamsulemani97, line 163:
```suggestion
To customize your index settings, we recommend following [this guide](/learn/core_concepts/indexes.md#index-settings). To learn more about the differences between settings in Algolia and Meilisearch, read on.
```

dichotommy (approved):
🧘🏻

maryamsulemani97:
bors merge

PR #2059, opened 2022-12-22
Fix cURL code-sample where POST method was in lowercase

bidoubiwa:
The `date_guide_index_1` code sample method flag is set as `post`, but it should be uppercased `POST`. 

https://github.com/meilisearch/documentation/blob/fc03ae707cb7890a0ce11be15165205234e26cee/.code-samples.meilisearch.yaml#L933-L937


dichotommy (approved):
Thank you!! 🍰

dichotommy:
bors merge

PR #2102, opened 2023-01-11
v1.0: remove two-step dump import process

### LINKED ISSUE #2067: v1.0: remove two-step dump import
Currently, users with v0.20 and below need to follow a two-step process for updating Meilisearch due to different dump versions. With v1.0, all dumps created before and up to Meilisearch v1.0 can be imported into Meilisearch v1.X.Y.

## To do

- Update `updating.md` to remove content on the two-step process
- Update `learn/advanced/dumps.md`

## References
- Spec: https://github.com/meilisearch/specifications/pull/203
- Issue on core: https://github.com/meilisearch/meilisearch/issues/2985
- SME: @dureuill

maryamsulemani97:
closes #2067

## Docs team review

[This comment](https://github.com/meilisearch/meilisearch/issues/2985#issuecomment-1333840498) mentions some version-specific update FYIs that we can document. I added the first (`enqueued` updates are not exported) and last (DB creation date isn't exported) points. The other two seem too dump version specific. Please let me know if I should add them as well.

### DIFF
--- a/learn/advanced/updating.md
+++ b/learn/advanced/updating.md
@@ -22,6 +22,8 @@ This section contains instructions for upgrading from specific versions. Most ve
 
 - If you are updating **from v0.19 or below**, please ensure all updates finish processing before creating the dump. `enqueued` updates will not be exported and may result in data loss.
 
+- If you are updating from **v0.20 or below**, the database creation date will not be imported to the new version.
+
 - If you are using **v0.24 or below**, use the `X-Meili-API-Key: API_KEY` authorization header:
 
 <CodeSamples id="updating_guide_check_version_old_authorization_header" />
--- a/learn/advanced/updating.md
+++ b/learn/advanced/updating.md
@@ -20,6 +20,8 @@ This guide only works for versions v0.15 and above. If you are using an older ve
 
 This section contains instructions for upgrading from specific versions. Most versions don't require version-specific steps and you should be able to upgrade directly. If the version you are upgrading from isn't listed here, no additional steps are required.
 
+- If you are updating **from v0.19 or below**, please ensure all updates finish processing before creating the dump. `enqueued` updates will not be exported and may result in data loss.
+
 - If you are using **v0.24 or below**, use the `X-Meili-API-Key: API_KEY` authorization header:
 
 <CodeSamples id="updating_guide_check_version_old_authorization_header" />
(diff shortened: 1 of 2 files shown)


learn/advanced/updating.md
```diff
@@ -20,6 +20,10 @@ This guide only works for versions v0.15 and above. If you are using an older ve
 
 This section contains instructions for upgrading from specific versions. Most versions don't require version-specific steps and you should be able to upgrade directly. If the version you are upgrading from isn't listed here, no additional steps are required.
 
+- If you are updating **from v0.19 or below**, please ensure all updates finish processing before creating the dump. `enqueued` updates will not be exported and may result in data loss.
```
dichotommy, line 23: @dureuill I know you already reviewed this, but I wanted to double-check because it's a bit surprising!

Is this behavior specific to exporting a dump from Meilisearch versions < v0.20? Not including v0.20?

I'm surprised because I know that v0.21 is when we introduced v2 of dumps, so I would think that it would also affect v0.20.

Thanks 🙏🏻

learn/advanced/updating.md
```diff
@@ -20,6 +20,10 @@ This guide only works for versions v0.15 and above. If you are using an older ve
...
+
+- If you are updating from **v0.20 or below**, the database creation date will not be imported to the new version.
```
dichotommy, line 25: My opinion is that this is not important enough to users to document here.

For one thing, it only affects users that are ten or more versions behind the latest release. More importantly, I personally cannot imagine any situation where losing the database creation date would have an impact on existing code processes.

Since it's distracting in a guide that is already very information and warning-heavy, I am suggesting we remove it. @dureuill Do you object?

learn/advanced/updating.md
dureuill: Thank you @dichotommy for the double-check!

You would be correct [I wrote versions <0.20](https://github.com/meilisearch/meilisearch/issues/2985#issuecomment-1333840498), but what I meant was really <=0.20. Every version of Meilisearch making use of the dumps v1 has the same issue of not exporting the necessary data to complete any pending task.

That must be why they say that the two hard problems of computing are naming, cache invalidation, and off-by-one errors 😬

learn/advanced/updating.md
dureuill: No objection from here

learn/advanced/updating.md
dichotommy: > That must be why they say that the two hard problems of computing are naming, cache invalidation, and off-by-one errors 😬

Hahaha, that's a good one 😆

dichotommy (approved):
LGTM! 🎆 🐰

guimachiavelli:
bors merge

PR #2104, opened 2023-01-11
v1.0: Primary key inference

### LINKED ISSUE #2075: v1.0: primary key inference
v1.0 changes the behavior of primary key inference
## To do
Some may be missing
- [ ] Update the "Add documents" section of the quick start
- [ ] Update [primary key guide](https://docs.meilisearch.com/learn/core_concepts/primary_key.html)
- [ ] Update the "update index" endpoint
- [ ] Update error codes

## References
- Issue on core: https://github.com/meilisearch/meilisearch/issues/3233
- Notion document: https://www.notion.so/meilisearch/Proto-documentation-402680403ae5460ea180cc00a5ed08c0
- Spec:
- SME: @dureuill

maryamsulemani97:
closes #2075

### DIFF
--- a/learn/core_concepts/primary_key.md
+++ b/learn/core_concepts/primary_key.md
@@ -165,7 +165,7 @@ This error occurs when you add documents for the first time and none of them hav
     "indexedDocuments":5
   },
   "error":{
-    "message": "The primary key inference process failed because the engine did not find any field ending with `id` in its name. Please specify the primary key manually using the `primaryKey` query parameter.",
+    "message": "The primary key inference failed as the engine did not find any field ending with `id` in its name. Please specify the primary key manually using the `primaryKey` query parameter.",
     "code": "index_primary_key_no_candidate_found",
     "type": "invalid_request",
     "link": "https://docs.meilisearch.com/errors#index-primary-key-no-candidate-found"
@@ -249,7 +249,7 @@ This error occurs when Meilisearch fails to guess your primary key as the index 
     "indexedDocuments":5
   },
   "error":{
-    "message": "The primary key inference process failed because the engine found 2 fields ending with `id` in their name, such as 'id' and 'author_id'. Please specify the primary key manually using the `primaryKey` query parameter.",
+    "message": "The primary key inference failed as the engine found 2 fields ending with `id` in their names: 'id' and 'author_id'. Please specify the primary key manually using the `primaryKey` query parameter.",
     "code": "index_primary_key_multiple_candidates_found",
     "type": "invalid_request",
     "link": "https://docs.meilisearch.com/errors#index-primary-key-multiple-candidates-found"
--- a/learn/core_concepts/primary_key.md
+++ b/learn/core_concepts/primary_key.md
@@ -141,7 +141,7 @@ The code below updates the primary key to `title`:
 
 ### Meilisearch guesses your primary key
 
-Suppose you add documents to an index without previously setting its primary key. In this case, Meilisearch will automatically look for an attribute containing the string `id` in a case-insensitive manner (for example, `uid`, `BookId`, `ID`, `123id123`) in your first document and set it as the index's primary key.
+Suppose you add documents to an index without previously setting its primary key. In this case, Meilisearch will automatically look for an attribute ending with the string `id` in a case-insensitive manner (for example, `uid`, `BookId`, `ID`) in your first document and set it as the index's primary key. If it finds multiple attributes ending with `id`, it will throw an error and ask you to manually set the primary key.
 
 If Meilisearch cannot find a suitable attribute, the document addition process will be interrupted and no documents will be added to your index.
 
--- a/learn/core_concepts/primary_key.md
+++ b/learn/core_concepts/primary_key.md
@@ -149,6 +149,34 @@ If Meilisearch finds [multiple attributes ending with `id`](#index-primary-key-m
 
 This section covers some primary key errors and how to resolve them.
 
+### `index_primary_key_multiple_candidates_found`
+
+This error occurs when you add documents to an index for the first time and Meilisearch finds multiple attributes ending with `id`. It can be resolved by [manually setting the index's primary key](#setting-the-primary-key-on-document-addition).
+
+```json
+{
+  "uid": 4,
+  "indexUid": "books",
+  "status": "failed",
+  "type": "documentAdditionOrUpdate",
+  "canceledBy":null,
+  "details":{
+    "receivedDocuments":5,
+    "indexedDocuments":5
+  },
+  "error":{
+    "message": "The primary key inference failed as the engine found 2 fields ending with `id` in their names: 'id' and 'author_id'. Please specify the primary key manually using the `primaryKey` query parameter.",
+    "code": "index_primary_key_multiple_candidates_found",
+    "type": "invalid_request",
+    "link": "https://docs.meilisearch.com/errors#index-primary-key-multiple-candidates-found"
+  },
+  "duration": "PT0.006002S",
+  "enqueuedAt": "2023-01-17T10:44:42.625574Z",
+  "startedAt": "2023-01-17T10:44:42.626041Z",
+  "finishedAt": "2023-01-17T10:44:42.632043Z"
+}
+```
+
 ### `index_primary_key_no_candidate_found`
 
 This error occurs when you add documents to an index for the first time and none of them have an attribute ending with `id`. It can be resolved by [manually setting the index's primary key](#setting-the-primary-key-on-document-addition), or ensuring that all documents you add possess an `id` attribute.
@@ -175,34 +203,6 @@ This error occurs when you add documents to an index for the first time and none
   "startedAt": "2023-01-17T10:19:14.465369Z",
   "finishedAt": "2023-01-17T10:19:14.471948Z"
   }
-```
-
-### `missing_document_id`
-
-This error occurs when your index already has a primary key, but one of the documents you are trying to add is missing this attribute.
-
-```json
-{
-    "uid":1,
-    "indexUid": "books",
-    "status": "failed",
-    "type": "documentAdditionOrUpdate",
-    "canceledBy": null,
-    "details":{
-        "receivedDocuments":1,
-        "indexedDocuments":null
-        },
-    "error":{
-        "message": "Document doesn't have a `id` attribute: `{\"title\":\"Solaris\",\"author\":\"Stanislaw Lem\",\"genres\":[\"science fiction\"],\"price\":5.0.",
-        "code": "missing_document_id",
-        "type": "invalid_request",
-        "link": "https://docs.meilisearch.com/errors#missing_document_id"
-        },
-    "duration": "PT0.007899S",
-    "enqueuedAt": "2021-12-30T11:23:52.304689Z",
-    "startedAt": "2021-12-30T11:23:52.307632Z",
-    "finishedAt": "2021-12-30T11:23:52.312588Z"
-}
 ```
 
 ### `invalid_document_id`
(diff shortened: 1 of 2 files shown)


reference/errors/error_codes.md
```diff
@@ -54,6 +54,10 @@ An index with this UID was not found, check out our guide on [index creation](/l
 
 The requested index already has a primary key that cannot be changed.
 
+## `index_primary_key_multiple_candidates_found`
+
+[Primary key inference](/learn/core_concepts/primary_key.md#meilisearch-guesses-your-primary-key) failed as the index contains multiple fields ending with `id`. Use the [update index endpoint](/reference/api/indexes.md#update-an-index) to manually set a primary key or add `id` to one of your fields so it can be used as the primary key during inference. We recommend manually setting the primary key.
```
dureuill, line 59: AFAIK adding `id` won't work in this case because we already have multiple candidates, and `id` is not given any priority.

As it probably doesn't make sense for users to remove fields from their documents, I'd say that manually specifying the field representing the primary key is their only recourse when there are multiple candidates.

Adding `id` to one of the fields **does** work in the `index_primary_key_no_candidate_found` case, though.

learn/core_concepts/primary_key.md
```diff
@@ -141,7 +141,7 @@ The code below updates the primary key to `title`:
 
 ### Meilisearch guesses your primary key
 
-Suppose you add documents to an index without previously setting its primary key. In this case, Meilisearch will automatically look for an attribute containing the string `id` in a case-insensitive manner (for example, `uid`, `BookId`, `ID`, `123id123`) in your first document and set it as the index's primary key.
+Suppose you add documents to an index without previously setting its primary key. In this case, Meilisearch will automatically look for an attribute ending with the string `id` in a case-insensitive manner (for example, `uid`, `BookId`, `ID`) in your first document and set it as the index's primary key. If it finds multiple attributes ending with `id`, it will throw an error and ask you to manually set the primary key.
```
dureuill, line 144: I wonder if we could describe the error case with multiple attributes maybe in the next paragraph as that would emphasize that the behavior of Meilisearch is similar when there is no candidate and when there are multiple candidates found.

> If Meilisearch cannot find a suitable attribute, or if it finds multiple attributes, ...

dureuill (changes requested):
Thanks for the PR! I have one technical comment about adding `id` to a field in the case where multiple candidates are found (see inline comments)

learn/core_concepts/primary_key.md
```diff
@@ -141,40 +141,40 @@ The code below updates the primary key to `title`:
...
+  "indexUid": "books",
+  "status": "failed",
+  "type": "documentAdditionOrUpdate",
+  "canceledBy": null,
+  "details":{
+    "receivedDocuments":5,
+    "indexedDocuments":5
+  },
+  "error":{
+    "message": "The primary key inference process failed because the engine did not find any field ending with `id` in its name. Please specify the primary key manually using the `primaryKey` query parameter.",
```
dureuill, line 168: Ah, sorry for not noticing this earlier, but it looks like the error message is not the latest version we decided upon together (the [specification PR](https://github.com/meilisearch/specifications/pull/208) is [up to date](https://github.com/meilisearch/specifications/pull/208#issuecomment-1372005125))

```suggestion
    "message": "The primary key inference failed as the engine did not find any field ending with `id` in its name. Please specify the primary key manually using the `primaryKey` query parameter.",
```

learn/core_concepts/primary_key.md
```diff
@@ -232,3 +232,31 @@ This happens when your document id does not have the correct [format](#formattin
...
+  "indexUid": "books",
+  "status": "failed",
+  "type": "documentAdditionOrUpdate",
+  "canceledBy":null,
+  "details":{
+    "receivedDocuments":5,
+    "indexedDocuments":5
+  },
+  "error":{
+    "message": "The primary key inference process failed because the engine found 2 fields ending with `id` in their name, such as 'id' and 'author_id'. Please specify the primary key manually using the `primaryKey` query parameter.",
```
dureuill, line 252: Same here:

```suggestion
    "message": "The primary key inference failed as the engine found 2 fields ending with `id` in their names: 'id' and 'author_id'. Please specify the primary key manually using the `primaryKey` query parameter.",
```

dureuill (changes requested):
I have further comments, I'm sorry about the second round of review, the error messages somehow escaped me the first time around.

dureuill (approved):
Looks good to me! Thank you for the PR 👍

learn/core_concepts/primary_key.md
```diff
@@ -141,40 +141,40 @@ The code below updates the primary key to `title`:
...
+Suppose you add documents to an index without previously setting its primary key. In this case, Meilisearch will automatically look for an attribute ending with the string `id` in a case-insensitive manner (for example, `uid`, `BookId`, `ID`) in your first document and set it as the index's primary key.
 
-If Meilisearch cannot find a suitable attribute, the document addition process will be interrupted and no documents will be added to your index.
+If Meilisearch finds multiple attributes ending with `id`, it will throw the [`index_primary_key_multiple_candidates_found`](#index-primary-key-multiple-candidates-found) error and ask you to manually set the primary key. If it cannot find a suitable attribute, it will throw the [`index_primary_key_no_candidate_found`](#index-primary-key-no-candidate-found) error. In both cases, the document addition process will be interrupted and no documents will be added to your index.
```
dichotommy, line 146: I felt that the error code names are unnecessary here, since they make the text more congested and are also mentioned further down on the page. Please push back if you disagree.

```suggestion
If Meilisearch finds [multiple attributes ending with `id`](#index-primary-key-multiple-candidates-found) or [cannot find a suitable attribute](#index-primary-key-no-candidate-found), it will throw an error. In both cases, the document addition process will be interrupted and no documents will be added to your index.
```

learn/core_concepts/primary_key.md
```diff
@@ -232,3 +232,31 @@ This happens when your document id does not have the correct [format](#formattin
     "finishedAt": "2021-12-30T11:28:59.084803Z"
 }
 ```
+
+### `index_primary_key_multiple_candidates_found`
```
dichotommy, line 236: Should this follow directly the other new error, since they both have to do with primary key inference?

learn/core_concepts/primary_key.md
```diff
@@ -141,40 +141,40 @@ The code below updates the primary key to `title`:
...
 
 ## Primary key errors
 
 This section covers some primary key errors and how to resolve them.
 
-### `primary_key_inference_failed`
+### `index_primary_key_no_candidate_found`
 
-This error occurs when you add documents for the first time and Meilisearch [fails to guess your primary key](#meilisearch-guesses-your-primary-key). It can be resolved by [manually setting the index's primary key](#setting-the-primary-key-on-document-addition), or ensuring that all documents you add possess an `id` attribute.
+This error occurs when you add documents for the first time and none of them have an attribute ending with `id`. It can be resolved by [manually setting the index's primary key](#setting-the-primary-key-on-document-addition), or ensuring that all documents you add possess an `id` attribute.
```
dichotommy, line 154:
```suggestion
This error occurs when you add documents to an index for the first time and none of them have an attribute ending with `id`. It can be resolved by [manually setting the index's primary key](#setting-the-primary-key-on-document-addition), or ensuring that all documents you add possess an `id` attribute.
```

learn/core_concepts/primary_key.md
```diff
@@ -232,3 +232,31 @@ This happens when your document id does not have the correct [format](#formattin
     "finishedAt": "2021-12-30T11:28:59.084803Z"
 }
 ```
+
+### `index_primary_key_multiple_candidates_found`
+
+This error occurs when Meilisearch fails to guess your primary key as the index has multiple attributes containing `id`. It can be resolved by [manually setting the index's primary key](#setting-the-primary-key-on-document-addition).
```
dichotommy, line 238: Do you see why I am suggesting these changes?

```suggestion
This error occurs when you add documents to an index for the first time and Meilisearch finds multiple attributes ending with `id`. It can be resolved by [manually setting the index's primary key](#setting-the-primary-key-on-document-addition).
```

reference/errors/error_codes.md
```diff
@@ -54,6 +54,10 @@ An index with this UID was not found, check out our guide on [index creation](/l
 
 The requested index already has a primary key that cannot be changed.
 
+## `index_primary_key_multiple_candidates_found`
+
+[Primary key inference](/learn/core_concepts/primary_key.md#meilisearch-guesses-your-primary-key) failed as the index contains multiple fields ending with `id`. Use the [update index endpoint](/reference/api/indexes.md#update-an-index) to manually set a primary key.
```
dichotommy, line 59:
```suggestion
[Primary key inference](/learn/core_concepts/primary_key.md#meilisearch-guesses-your-primary-key) failed because the received documents contain multiple fields ending with `id`. Use the [update index endpoint](/reference/api/indexes.md#update-an-index) to manually set a primary key.
```

reference/errors/error_codes.md
```diff
@@ -154,9 +158,9 @@ The `minWordSizeForTypos` object is invalid. The value for both `oneTypo` and `t
 
 The field you are trying to modify is immutable.
 
-## `primary_key_inference_failed`
+## `index_primary_key_no_candidate_found`
 
-The first provided document contains no fields with the substring `id`. [Manually designate the primary key](/learn/core_concepts/primary_key.md#setting-the-primary-key) or add `id` to one of your fields so it can be used as the primary key during inference. We recommend manually setting the primary key.
+[Primary key inference](/learn/core_concepts/primary_key.md#meilisearch-guesses-your-primary-key) failed as the index does not contain any fields ending with `id`. [Manually designate the primary key](/learn/core_concepts/primary_key.md#setting-the-primary-key) or add `id` to one of your fields so it can be used as the primary key during inference. We recommend manually setting the primary key.
```
dichotommy, line 163: I don't think we should make this recommendation, as their dataset might not contain a field suitable to be used as a primary key (i.e. present in every document and containing unique values).

```suggestion
[Primary key inference](/learn/core_concepts/primary_key.md#meilisearch-guesses-your-primary-key) failed as the received documents do not contain any fields ending with `id`. [Manually designate the primary key](/learn/core_concepts/primary_key.md#setting-the-primary-key), or add some field ending with `id` to your documents.
```

dichotommy (changes requested):

learn/core_concepts/primary_key.md
```diff
@@ -141,40 +141,40 @@ The code below updates the primary key to `title`:
...
-    "finishedAt": "2021-12-30T11:17:49.716303Z"
-}
+  "uid": 1,
+  "indexUid": "books",
+  "status": "failed",
+  "type": "documentAdditionOrUpdate",
+  "canceledBy": null,
+  "details":{
+    "receivedDocuments":5,
+    "indexedDocuments":5
```
maryamsulemani97, line 165:
```suggestion
    "receivedDocuments": 5,
    "indexedDocuments": null
```

learn/core_concepts/primary_key.md
maryamsulemani97: The sentence paraphrases the error and isn't useful

learn/core_concepts/primary_key.md
maryamsulemani97: It would be better to list these errors in alphabetical order as we do in `error_codes.md`

dichotommy: I disagree.

Alphabetical order makes sense for a reference file like `error_codes.md`, where people will be searching for a specific error. Being alphabetical helps them find that error more quickly.

Unlike `error_codes.md`, this page is intended to be read sequentially. Because of that, these errors should be grouped by topic/conceptually.

learn/core_concepts/primary_key.md
dichotommy: The main thing is that the difference between "containing id" and "ending with id" is potentially huge; it's the difference between having an accurate understanding and expectation of the inference, and not.

dichotommy: The other thing is that saying "the index has multiple attributes containing `id`" is incorrect and potentially confusing. When the error occurs, the document addition fails, so the index is empty.

learn/core_concepts/primary_key.md
dichotommy: In this case, it seems like alphabetical order puts the errors in an intuitive order anyway, so further changes may not be required. Just something to think about in the future.

dichotommy (approved):
LGTM! 🧑‍🍳

maryamsulemani97:
bors merge

PR #2111, opened 2023-01-16
v1.0: update to add  `*` for date filters in task api

### LINKED ISSUE #2109: v1.0: update to add `*` for date filters in task API
Update task API to add `*` for the date filters

## To do
- [x] Update task API
- [x] Update async operations guide

## References
Issue on core: https://github.com/meilisearch/meilisearch/issues/3342
Spec: https://github.com/meilisearch/specifications/pull/218
SME: @irevoire

maryamsulemani97:
closes #2109

### DIFF
--- a/reference/api/tasks.md
+++ b/reference/api/tasks.md
@@ -357,10 +357,10 @@ A valid `uids`, `statuses`, `types`, `indexUids`, or date(`beforeXAt` or `afterX
 | **`statuses`**         | Cancel tasks based on `status`. Separate multiple `statuses` with a comma (`,`). Use `statuses=*` for all `statuses`                 |
 | **`types`**            | Cancel tasks based on `type`. Separate multiple `types` with a comma (`,`). Use `types=*` for all `types`                            |
 | **`indexUids`**        | Cancel tasks based on `indexUid`. Separate multiple `uids` with a comma (`,`). Use `indexUids=*` for all `indexUids`. Case-sensitive |
-| **`beforeEnqueuedAt`** | Cancel tasks **before** a specified `enqueuedAt` date. Use `beforeEnqueuedAt=*` for all tasks enqueued before the specified date     |
-| **`beforeStartedAt`**  | Cancel tasks **before** a specified `startedAt` date. Use `beforeStartedAt=*` for all tasks started before the specified date        |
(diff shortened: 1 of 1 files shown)


guimachiavelli (commented):
I'm not sure the added wildcard explanation makes sense. 

For example: we say `beforeEnqueuedAt=*` returns all tasks enqueued before the specified date. However, there is no specified date—the value is not a date, but `*`. 

The spec states:

> when `beforeEnqueuedAt` query parameter is not set or set to `*`, all task are selected.

Which, AFAIU, would mean that `beforeEnqueuedAt=*` is functionally the same as not adding `beforeEnqueuedAt`.

Am I missing something? Or would a better way of explaining the wildcard default be something like:

> `*` (all enqueued tasks)

guimachiavelli (commented):
Don't we need to propagate the "all tasks" change to the response fields? (lines 360-363 and 422-427)

guimachiavelli (approved):
📟

reference/api/tasks.md
```diff
@@ -357,10 +357,10 @@ A valid `uids`, `statuses`, `types`, `indexUids`, or date(`beforeXAt` or `afterX
...
 | **`types`**            | Cancel tasks based on `type`. Separate multiple `types` with a comma (`,`). Use `types=*` for all `types`                            |
 | **`indexUids`**        | Cancel tasks based on `indexUid`. Separate multiple `uids` with a comma (`,`). Use `indexUids=*` for all `indexUids`. Case-sensitive |
-| **`beforeEnqueuedAt`** | Cancel tasks **before** a specified `enqueuedAt` date                                                                                |
-| **`beforeStartedAt`**  | Cancel tasks **before** a specified `startedAt` date                                                                                 |
-| **`afterEnqueuedAt`**  | Cancel tasks **after** a specified `enqueuedAt` date                                                                                 |
-| **`afterStartedAt`**   | Cancel tasks **after** a specified `startedAt` date                                                                                  |
+| **`beforeEnqueuedAt`** | Cancel tasks **before** a specified `enqueuedAt` date. Use `beforeEnqueuedAt=*` to cancel all tasks                                  |
```
irevoire, line 360: Shouldn't we write ``` `*` (all tasks)```like we did for all the other parameters?

Or at least I think we should use the word `select` instead of `cancel`. Because you can select everything in one field and then constraint the selection to only the `documentAddition` for example.

guimachiavelli:
bors merge

Issue #851, opened 2021-03-10
Change relevancy examples images to text


fharper:
Change [relevancy examples](https://docs.meilisearch.com/learn/core_concepts/relevancy.html#examples) images to text so it's accessible for people with vision impairment and right now, the contrast for bold words isn't good for colorblind folks.

guimachiavelli:
Changing these images to text would negatively impact users—they're gifs illustrating a process already described in the accompanying text. Describing these gifs would result in content duplication.

We could improve the contrast of the images to address some of the accessibility concerns, but it's unclear where the original interface is located, and recreating it would use resources better spent elsewhere.

PR #2126, opened 2023-01-24
reduce vale errors

maryamsulemani97:
This PR updates:
- `accept.txt`
  - add words used in the docs that Vale considers typos
- `Headings.yml`
  - add words used with a capital letter in headings
  - add `"(.*?)"` to ignore words in quotes
- `FirstPerson.yml`
  -  update the scope to sentence. This means the rule only applies to sentences and not headings.

Total error count: 0 errors, 2 warnings and 8 suggestions in 75 files.


.vale/styles/Meilisearch/Headings.yml
```diff
@@ -51,6 +51,9 @@ exceptions:
   - Bleve & Tantivy
   - Google Cloud Console
   - AMIs
+  - Droplet
+  - Previous
+  - Next
```
dichotommy, line 56: That's interesting! I'm not surprised that "Next" is throwing a warning, but "Previous" _is_ the first word in the title. Do you think the quotation mark is throwing off the rule?

> "Previous" and "Next" buttons

I wonder if it would be possible to alter the rule so that it ignores capitals immediately following quotation marks. I would prefer that, versus adding these ordinary words to the proper noun list.

.vale/styles/Vocab/word_list/accept.txt
```diff
@@ -78,3 +78,27 @@ AMIs
...
+npm
+html
+Maneu
+Hoxie
+Bleve
+Yonik
+Seeley
+MBs
+url
+masterKey
```
dichotommy, line 97: Doesn't this rule ignore text wrapped in `<code>` elements? I'm curious about this one, as well as a few others that normally would be capitalized (`toc`, `gcp`, `faq`, `npm`, `url`).

.vale/styles/Vocab/word_list/accept.txt
```diff
@@ -78,3 +78,27 @@ AMIs
...
+SDK's
```
dichotommy, line 98: Is it necessary to specify the `s`? If we simply changed the earlier `SDKs` to `SDK`, would that solve the problem? Not sure how smart this rule is (i.e. does it form plurals of words added to the list and ignore those too?)

Mostly curious to better our understanding of Vale rules—not a problem if `SDK's` makes it onto the list.

.vale/styles/Vocab/word_list/accept.txt
```diff
@@ -78,3 +78,27 @@ AMIs
...
+Bardem
+LMDB's
+vogli
+belgium
```
dichotommy, line 102: Should probably be capitalized wherever it is used, in which case it would probably not need to be added to this list?

.vale/styles/Vocab/word_list/accept.txt
maryamsulemani97: gcp, faq, npm, url are all used in alt text. You can't use backticks in alt text, so I added them here to reduce the warnings.
toc isn't used in backticks. It's [toc]. We only have one page that uses it. We should remove it as well.

.vale/styles/Vocab/word_list/accept.txt
maryamsulemani97: >does it form plurals of words added to the list and ignore those too?

It does not. So if `SDK` isn't a word, neither is `SDKs` or `SDK's`.

.vale/styles/Meilisearch/Headings.yml
maryamsulemani97: Good catch! I didn't notice the quotes

.vale/styles/Vocab/word_list/accept.txt
dichotommy: I'm confused by your answer. It implies that if `SDK` is added to the list, `SDKs` and `SDK's` should also be accepted.

.vale/styles/Vocab/word_list/accept.txt
dichotommy: Can we change it to be capitalized in the search query and remove this from the list?

.vale/styles/Meilisearch/Headings.yml
```diff
@@ -52,8 +52,7 @@ exceptions:
   - Google Cloud Console
   - AMIs
   - Droplet
-  - Previous
-  - Next
+  - "(.*?)" #ignore words in quotes
```
dichotommy, line 55: Awesome, thanks! ❤️

dichotommy:
AWS is our most-used deploy cookbook by a wide margin (followed by DO and GCP). It's also quite long. We want users to be able to return to it and quickly find the information they're looking for. 

Since we're removing the `[toc]`s from AWS and GCP cookbooks, I would recommend changing the header level displayed in the sidebar for this whole section. we may want to . These are our longest and most-used cookbooks.

.vale/styles/Vocab/word_list/accept.txt
dichotommy: Ok, that makes sense. However there's something going on here related to acronyms/case-sensitivity, no? I mean, we have `GCP` in an article title and it's not setting off any flags. So why is the lowercase causing problems? And shouldn't the lowercase `gcp` in the alt-text be uppercase, anyway (same for `faq` and `url` IMO)?

.vale/styles/Vocab/word_list/accept.txt
maryamsulemani97: Sorry if it was clear. If `SDK` isn't a word according to Vale, neither is `SDKs`, `SDK's`, or any other version of `SDK`. We will have to add all variations of the word separately to `accept.txt`.

.vale/styles/Vocab/word_list/accept.txt
maryamsulemani97: - faq, npm, synon: we're using GIFs that search for these terms in lowercase
- gcp: only used once in the telemetry table in the "Example" column. We cannot use uppercase there
- url: "url" is a field in Postman. We added `"(.*?)"` so this won't be an issue anymore

.vale/styles/Vocab/word_list/accept.txt
maryamsulemani97: We'll have to update the image: 
Screen Shot 2023-02-13 at 14 33 01

If we update the "Attribute" tab, we'll have to update the rest as well. This can be done in a separate PR as we have an [open issue for updating this example ](https://github.com/meilisearch/documentation/issues/851)

maryamsulemani97:
>Since we're removing the [toc]s from AWS and GCP cookbooks, I would recommend changing the sidebarDepth for this whole Deploy section.

I didn't add sidebarDepth for Railway, Qovery, and Koyeb as they don't have as many headings and the pages are very short compared to the other cookbooks

.vale/styles/Vocab/word_list/accept.txt
dichotommy: I see. Is there a way to make this rule ignore alt text?

maryamsulemani97: We could, but then Vale would ignore all alt text which isn't a good idea

dichotommy: The reason I am pushing for this is that adding `url, `gcp`, and `faq` to our exceptions list will make our linter less effective since we never actually want to write them like this in body text.

.vale/styles/Vocab/word_list/accept.txt
dichotommy: Huh. I think there is still a little miscommunication—I am wondering about what would happen if `SDK` _is_ a word according to Vale, not if it _isn't_. 

It would seem logical to me that if SDK is defined as a word, `SDK's` would be accepted as a natural grammatical extension of that word. Less sure about `SDKs` but it may also work. Could you test this, please?

.vale/styles/Vocab/word_list/accept.txt
dichotommy: Ok, not worth doing it just for this. I'll resolve this conversation, since it's related to the above one: finding a way to ignore alt text for linting would solve both problems.

dichotommy:
Could you convert this into a draft while you look into disabling Vale for alt text, please?

dichotommy:
Hi @maryamsulemani97 , could you please give an update on the status of this PR? I saw a commit titled "ignore alt text" but when I looked into it I wasn't sure if we have found a way to make Vale ignore alt text yet or not. Thanks!

maryamsulemani97:
`!\[(.*?)\]` should ignore alt text, but it's not. I think we should merge this PR for now and open a new one for alt text

.vale/styles/Vocab/word_list/accept.txt
maryamsulemani97: If it is a word, then we don't need to add `SDKs`, `SDK's` or any other variation

guimachiavelli (approved):
👯

maryamsulemani97:
bors merge

Issue #2125, opened 2023-01-19
Restructure quick start: Installation page


maryamsulemani97:
As part of restructuring the quick start, we will start with a new dedicated Installation page. This will move the content from the ["Setup and installation"](https://docs.meilisearch.com/learn/getting_started/quick_start.html#setup-and-installation) section of the quick start and the [Supported OS](https://github.com/meilisearch/documentation/pull/2112#pullrequestreview-1261354509) page to the Installation page.
We should also mention the Meilisearch cloud in the quick start and Installation page.

dichotommy:
The new installation page will centralize information related to installation, but some instructions will still need to remain in the quick start so that it can stand alone. The question is, how much?

What is the "bare minimum" so that users can install Meilisearch in a simple and rapid way?

I would say:
- Curl
- Homebrew
- Docker
- Meilisearch Cloud

I feel that eliminating the cloud deploy section and putting everything in tabs is a good idea. WDYT?

I chose not to include AWS, GCP, and DigitalOcean because they are redundant and in some ways in direction competition with Meilisearch Cloud. I think we would prefer users to try the free tier of MC if possible.

On the other hand, that may require users to register an account, whereas they may already have accounts for AWS/GCP/DO. Something to consider!

We could also include APT, if we have space in the tab bar without creating scrolling.

With the "Windows" tab, it's not an installation method in and of itself, so I feel it may make things unnecessarily complicated. I think it makes much more sense to keep the Windows-related information in the installation page. It seems like Windows users will still be able to follow the quick start  using Docker. (But will they know that?)

I also think we may want to get rid of the ASCII `MEILISEARCH`, if only to reduce the amount of vertical scrolling users have to do to get to the meat of the quick start.

Ultimately, we want to reduce the amount of time it takes for users to reach the "A-HA" moment of executing their first search. All changes to the quick start should be done with that in mind.

PR #2140, opened 2023-02-01
Add installation page

### LINKED ISSUE #2139: Following docker instructions in the quickstart lead to unauthorized error
Hello there,

## Issue

Following the instructions from the quickstart guide leads to a 401 error because one code sample enforces admin key usage, but the other code samples do not use that key it.

## Reproduction

Following the instructions in quickstart guide (i.e. copy-pasting all the code), I

1) Launched Meilisearch with Docker 

```sh
docker run -it --rm \
    -p 7700:7700 \
    -e MEILI_MASTER_KEY='MASTER_KEY'\
    -v $(pwd)/meili_data:/meili_data \
    getmeili/meilisearch:v0.30 \
    meilisearch --env="development"
```

(This code sample provides a master key for the meili db)

2) Used the JavaScript example to import the movies.json sample data.

```js
const client = new MeiliSearch({
  host: 'http://localhost:7700'
})

client.index('movies').addDocuments(movies)
   .then((res) => console.log(res))
```

(This code sample does not provide an API key to authenticate)

3) Received a 401 `missing_authorization_header` error when running my JS script. This error gets fixed when I specify the API key.

maryamsulemani97:
closes #2139, #2125  
This PR:
- adds a dedicated installation page to the Getting started and shortens the "Installation" section of the quick start
- updates the Docker code sample to remove the master key

## For reviewers

Instead of putting everything in tabs, I added three headings: Local installation, Meilisearch Cloud, and Cloud Deploy.
>From the [issue](https://github.com/meilisearch/documentation/issues/2125#issuecomment-1397303675):
I chose not to include AWS, GCP, and DigitalOcean because they are redundant and in some ways in direction competition with Meilisearch Cloud. I think we would prefer users to try the free tier of MC if possible.

I think we should document all the possible installation methods, especially when we have a dedicated page.

As mentioned in the issue, the [Windows tab isn't an installation method](https://github.com/meilisearch/documentation/issues/2125#issuecomment-1397303675), but we should keep it since we have a dedicated Homebrew tab for macOS/Linux users.

>From the [issue](https://github.com/meilisearch/documentation/issues/2125#issuecomment-1397303675):
I also think we may want to get rid of the ASCII MEILISEARCH, if only to reduce the amount of vertical scrolling users have to do to get to the meat of the quick start.

We _could_ get rid of it. If it's to reduce the scrolling, we did remove content for the Installation page. I think we should keep it. It's reassuring to see your response is exactly the same as the one in the docs 😅 (though this one isn't as helpful as the code sample responses)

### DIFF
--- a/learn/getting_started/installation.md
+++ b/learn/getting_started/installation.md
@@ -1,0 +1,150 @@
+# Installation
+
+We'll start with downloading and installing Meilisearch. You have the option to install Meilisearch locally or deploy it over a cloud service.
+
+## Local installation
+
+:::: tabs
+
+::: tab cURL
+Download the **latest stable release** of Meilisearch with **cURL**.
+
+Launch Meilisearch to start the server.
+
+```bash
+# Install Meilisearch
+curl -L https://install.meilisearch.com | sh
+
+# Launch Meilisearch
+./meilisearch
+```
+
+:::
+
+::: tab Homebrew
+Download the **latest stable release** of Meilisearch with **[Homebrew](https://brew.sh/)**, a package manager for MacOS.
+
+Launch Meilisearch to start the server.
+
+```bash
+# Update brew and install Meilisearch
+brew update && brew install meilisearch
+
+# Launch Meilisearch
+meilisearch
+```
+
+:::
+
+::: tab Docker
+When using **Docker**, you can run [any available tag](https://hub.docker.com/r/getmeili/meilisearch/tags).
+
+These commands launch the **latest stable release** of Meilisearch.
+
+```bash
+# Fetch the latest version of Meilisearch image from DockerHub
+docker pull getmeili/meilisearch:v0.30
+
+# Launch Meilisearch in development mode with a master key
+docker run -it --rm \
+    -p 7700:7700 \
+    -e MEILI_MASTER_KEY='MASTER_KEY'\
+    -v $(pwd)/meili_data:/meili_data \
+    getmeili/meilisearch:v0.30 \
+    meilisearch --env="development"
+```
+
+You can learn more about [using Meilisearch with Docker in our dedicated guide](/learn/cookbooks/docker.md).
+:::
+
+::: tab APT
+
+Download the **latest stable release** of Meilisearch with **APT**.
+
+Launch Meilisearch to start the server.
+
+```bash
+# Add Meilisearch package
+echo "deb [trusted=yes] https://apt.fury.io/meilisearch/ /" | sudo tee /etc/apt/sources.list.d/fury.list
+
+# Update APT and install Meilisearch
+sudo apt update && sudo apt install meilisearch-http
+
+# Launch Meilisearch
+meilisearch
+```
+
+:::
+
+::: tab Source
+
+Meilisearch is written in `Rust`. To compile it, [install the Rust toolchain](https://www.rust-lang.org/tools/install).
+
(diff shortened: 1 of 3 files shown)


dichotommy (changes requested):
Looks pretty good to me! You copied basically everything that was in the Quick Start installation section over to the install page, right? That's what I was trying to suggest in my comments; all the bits you quoted were referring to the install options we should show in the quick start!

That being said, now that I'm seeing it a part of me thinks... why not show only curl for installation? We have a link to the other install options... it's more straightforward, users aren't asked to make a decision right off the bat. I dunno, I see pros + cons.

🚨 ATTENTION: check out the redirects file. I'm pretty sure there's an existing redirect from our old installation page, which probably had the same URL, to the quick start. We will want to delete that so that people can actually visit the new installation page 💁🏻 😂

dichotommy:
Oh, I see you altered the Docker code sample a bit from @ManyTheFish 's suggestion. Many suggested an env variable (-e MEILI_ENV='development'), you used `meilisearch --env="development"`.

I suppose there's probably not much of a difference, but I'm sort of thinking it may be better to use Many's suggestion, since it seems that Docker has an `--env` option, so it could be unclear if we are passing a Meili option or a Docker one.

maryamsulemani97:
> all the bits you quoted were referring to the install options we should show in the quick start!

My bad! I thought they were for the installation page. We could keep this for now, see how it goes, and add more tabs if needed?

>why not show only curl for installation? We have a link to the other install options... it's more straightforward

It's supposed to be quick 🤷‍♀️

dichotommy (dismissed):
🪂

ManyTheFish:
@dichotommy @maryamsulemani97,
for more info about `-e MEILI_ENV='development'` vs `meilisearch --env="development"`,
as you said, there is no real difference between both versions, one uses an environment variable `MEILI_ENV`, whereas the other uses the CLI option `env`.
The reason that I suggested `-e MEILI_ENV='development'` is to avoid specifying the command to use while running the docker image (`meilisearch` in this context).
So it's not a big suggestion. 🙂

learn/getting_started/installation.md
```diff
@@ -0,0 +1,149 @@
...
+:::
+
+::: tab Docker
+When using **Docker**, you can run [any available tag](https://hub.docker.com/r/getmeili/meilisearch/tags).
+
+These commands launch the **latest stable release** of Meilisearch.
+
+```bash
+# Fetch the latest version of Meilisearch image from DockerHub
+docker pull getmeili/meilisearch:v0.30
```
dichotommy, line 46:
```suggestion
docker pull getmeili/meilisearch:v1.0
```

learn/getting_started/installation.md
```diff
@@ -0,0 +1,149 @@
...
+
+# Launch Meilisearch in development mode with a master key
+docker run -it --rm \
+    -p 7700:7700 \
+    -e MEILI_ENV='development' \
+    -v $(pwd)/meili_data:/meili_data \
+    getmeili/meilisearch:v0.30
```
dichotommy, line 53:
```suggestion
    getmeili/meilisearch:v1.0
```

dichotommy (approved):
🏄🏻 🌊

dichotommy (dismissed):
A p p r o v e d 👍🏻

dichotommy (approved):
🤞🏻

dichotommy:
bors merge

PR #2142, opened 2023-02-01
v1.0: remove stable v1 mentions

guimachiavelli:
We found four old mentions reflecting our hopes and wishes for v1 around the documentation. Search was conducted using a grep-like tool with the following terms:

- `v1`
- `stable`
- `unstable`
- `in development`
- `major`
- `alpha`
- `beta`

Two mentions on the updating guide will be fixed by #2102.

This PR removes the remaining two mentions on the FAQ.

### DIFF
--- a/resources/faq.md
+++ b/resources/faq.md
@@ -82,7 +82,7 @@ Meilisearch major versions introduce breaking changes and are not compatible wit
 
 ## What are the recommended requirements for hosting a Meilisearch instance?
 
-Indexing documents is a highly complex process, making it difficult to accurately estimate the size and memory use of a Meilisearch database. There are a few aspects to keep in mind when optimizing your instance.
+Indexing documents is a complex process, making it difficult to accurately estimate the size and memory use of a Meilisearch database. There are a few aspects to keep in mind when optimizing your instance.
 
 ### Memory usage
 
--- a/resources/faq.md
+++ b/resources/faq.md
@@ -78,15 +78,11 @@ For more accurate features and issues, everything is detailed in the issues of a
 
 ## I have just updated Meilisearch, and I am getting an error: "Cannot open database, expected Meilisearch engine version..."
 
-Until our first stable release (v1.0), Meilisearch minor versions are not compatible with each other. This means **every new version is considered breaking** with the small exception of bug-fixing patches. To fix this error, simply delete your database folder (`data.ms` by default) and re-index your documents with the current-version engine. See [updating Meilisearch](/learn/advanced/updating.md) for more information.
+Meilisearch major versions introduce breaking changes and are not compatible with each other. Meilisearch version 0 releases, such as v0.30 or v0.24, are also not compatible with each other. Consult the [updating guide](/learn/advanced/updating.md) for more information on how to safely migrate your data between Meilisearch releases.
 
 ## What are the recommended requirements for hosting a Meilisearch instance?
 
-**The short answer: we don't know yet!**
-
-Indexing documents is a highly complex process, making it difficult to accurately estimate the size and memory use of a Meilisearch database. At this stage, our engine is changing constantly, which means benchmarks are constantly out of date.
-
-Once our engine is completely stable (v1), we will provide detailed benchmarks with different datasets. Until then, read on to learn how to optimize your Meilisearch instance.
+Indexing documents is a highly complex process, making it difficult to accurately estimate the size and memory use of a Meilisearch database. There are a few aspects to keep in mind when optimizing your instance.
 
 ### Memory usage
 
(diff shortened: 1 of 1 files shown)


maryamsulemani97 (approved):
🐼

resources/faq.md
```diff
@@ -78,15 +78,11 @@ For more accurate features and issues, everything is detailed in the issues of a
 
 ## I have just updated Meilisearch, and I am getting an error: "Cannot open database, expected Meilisearch engine version..."
 
-Until our first stable release (v1.0), Meilisearch minor versions are not compatible with each other. This means **every new version is considered breaking** with the small exception of bug-fixing patches. To fix this error, simply delete your database folder (`data.ms` by default) and re-index your documents with the current-version engine. See [updating Meilisearch](/learn/advanced/updating.md) for more information.
+Meilisearch major versions introduce breaking changes and are not compatible with each other. Meilisearch version 0 releases, such as v0.30 or v0.24, are also not compatible with each other. Consult the [updating guide](/learn/advanced/updating.md) for more information on how to safely migrate your data between Meilisearch releases.
```
curquiza, line 81: Hum I'm not sure this is the right answer to the right question.

First the new error the users will get is "Error: Your database version (0.30.5) is incompatible with your current engine version (1.0.0).
To migrate data between Meilisearch versions, please follow our guide on ...:

Secondly, this is not related to major versions. Users will need to use dump (and follow the guide) even for minor version.
I'm not sure this is related to talk about breaking changes here.

curquiza (changes requested):

resources/faq.md
dichotommy: Question: do we even need this FAQ anymore? 🤔 

I started to think about how I would rewrite it... but considering how much the wording of the error has been improved, I'm not sure if the answer would really add anything!

dichotommy: In the end, we'll just be explaining that you migrate your database between versions using a dump, and redirecting users to the update guide, right?

resources/faq.md
```diff
@@ -78,15 +78,11 @@ For more accurate features and issues, everything is detailed in the issues of a
...
 
 ## What are the recommended requirements for hosting a Meilisearch instance?
 
-**The short answer: we don't know yet!**
-
-Indexing documents is a highly complex process, making it difficult to accurately estimate the size and memory use of a Meilisearch database. At this stage, our engine is changing constantly, which means benchmarks are constantly out of date.
-
-Once our engine is completely stable (v1), we will provide detailed benchmarks with different datasets. Until then, read on to learn how to optimize your Meilisearch instance.
+Indexing documents is a complex process, making it difficult to accurately estimate the size and memory use of a Meilisearch database. There are a few aspects to keep in mind when optimizing your instance.
```
dichotommy, line 85: Would suggest we copy some text from the Crisp helpdesk article (which itself copies a lot of text from this answer). Here's my adaptation:

```suggestion
**The short answer:**

The recommended requirements for hosting a Meilisearch instance will depend on many factors, such as the number of documents, the size of those documents, the number of filters/sorts you will need, and more. For a quick estimate to start with, try to use a machine that has at least ten times the disk space of your dataset.

**The long answer:**

Indexing documents is a complex process, making it difficult to accurately estimate the size and memory use of a Meilisearch database. There are a few aspects to keep in mind when optimizing your instance.
```

Ideally these would be headers, not simply bolded text. The only reason I didn't suggest it that way is because it would probably require changing the following H3s to H4s.

resources/faq.md
curquiza: > Question: do we even need this FAQ anymore? 🤔

I cannot answer, the stats can haha
As a developer, and also as a person doing technical I like FAQ, but only my opinion. It's cool to have a "naive" approach to the product.

> In the end, we'll just be explaining that you migrate your database between versions using a dump, and redirecting users to the update guide, right?

Yes that's only the purpose of this question.
But if you estimate the error message is clear enough from v1, maybe this question in the FAQ is not relevant anymore.

resources/faq.md
guimachiavelli: I think the short answer is actually all we could reasonably need here.

Given v1 is just around the corner, I'll accept @dichotommy's suggestion. Then, at a later date, I'll move unique info to a better location and delete the rest.

curquiza (approved):
Ok for me to remove it! We will re-put it if support gets a lot of issues with it, but from my point of view, the new error message in v1 is clear enough and redirect to the most approrioate page of the docs (i.e the "upgrading meilisearch" guide)

dichotommy (approved):
🏄🏻

guimachiavelli:
bors merge

PR #2144, opened 2023-02-02
Use latest instead of main when requesting the default config file

curquiza:
Latest = the latest stable git tag (so v0.30.5 currently)
main = the branch containing the most recent commit (so more commits than in v0.30.5 currently)

### DIFF
--- a/learn/configuration/instance_options.md
+++ b/learn/configuration/instance_options.md
@@ -69,7 +69,7 @@ Meilisearch accepts a configuration file in the `.toml` format as an alternative
 You can download a default configuration file using the following command:
 
 ```sh
-curl https://raw.githubusercontent.com/meilisearch/meilisearch/main/config.toml > config.toml
+curl https://raw.githubusercontent.com/meilisearch/meilisearch/latest/config.toml > config.toml
 ```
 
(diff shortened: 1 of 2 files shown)


curquiza:
@meilisearch/engine-team FYI 😇

guimachiavelli (changes requested):
One request and one question:

- Could you also change L#87 of `/learn/cookbooks/running_production.md` so it also uses `latest`?
- I noticed one page uses  `wget` and the other uses `curl` and `>` to download the config file. I'd like to choose one approach and make it consistent for both code samples. Do you have any preferences?

curquiza:
> Could you also change L#87 of /learn/cookbooks/running_production.md so it also uses latest?

✅ 

> I noticed one page uses wget and the other uses curl and > to download the config file. I'd like to choose one approach and make it consistent for both code samples. Do you have any preferences?

I discussed with Kero, and we agreed `curl` is more used and more present by default on the machines than `wget`

guimachiavelli (approved):
Brilliant, thanks! ❤️

guimachiavelli:
bors merge

PR #2173, opened 2023-02-16
Updated Typesense comparison table

ferdi05:
Some notes:

**Don't trust me**
- I tried to make informed changes, but I also relied on info from [these tweets](https://twitter.com/typesense/status/1620825236055932928) and [this Hacker News thread](https://news.ycombinator.com/item?id=34707727) (although it's really painful to use)

**Language support**
- Meilisearch [supports any language](https://docs.meilisearch.com/learn/what_is_meilisearch/language.html), and provides optimized support for any language that uses whitespace to separate words, plus Chinese, Japanese, Korean, Hebrew & Thai.
- Typesense [claims to support any language](https://news.ycombinator.com/item?id=34715656), and possibly [optimized support for Chinese, Japanese and Korean](https://twitter.com/typesense/status/1620825236055932928)
- Algolia provides [the longest list of supported languages](https://www.algolia.com/doc/guides/managing-results/optimize-search-results/handling-natural-languages-nlp/in-depth/supported-languages/) and indicates specific support for plurals, stop words, and segmentation.
- Elasticsearch [is being Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-lang-analyzer.html): way too much details, and I didn't understand anything
- Being slightly lost, I added All languages to Typesense, but it's very arguable

** Upgrade path **
- I decided to remove this row as it looks like we're the only solution requiring a re-indexing. So it makes us look harder to use. If you think it's unfair, I'll happily undo this.

PR #2174, opened 2023-02-16
Update wrong / outdated statements about Typesense

jasonbosco:
# Pull Request

Hey team! 👋

Following up from @Kerollmops's comment here: https://github.com/typesense/typesense-website/pull/163#issuecomment-1433101898, I'm opening this PR to update some of the statements about Typesense, based on recent features we've launched. 

## Related issue

https://github.com/typesense/typesense-website/pull/163#issuecomment-1433101898

## What does this PR do?
Updates statements about Typesense in your comparison tables. 

## PR checklist
- [x] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [x] Have you read the contributing guidelines?

Thank you so much for contributing to Meilisearch!

### DIFF
--- a/learn/what_is_meilisearch/comparison_to_alternatives.md
+++ b/learn/what_is_meilisearch/comparison_to_alternatives.md
@@ -44,15 +44,15 @@ Can't find a client you'd like us to support? [Submit your idea or vote for it](
 | [Python client](https://github.com/meilisearch/meilisearch-python)              | ✅          | ✅      |        ✅   |       ✅        |
 | [Ruby client](https://github.com/meilisearch/meilisearch-ruby)              | ✅          | ✅      |        ✅   |       ✅        |
 | [Java client](https://github.com/meilisearch/meilisearch-java)              | ✅          | ✅      |        ✅   |       ✅        |
-| [Swift client](https://github.com/meilisearch/meilisearch-swift)              | ✅          | ✅      |   🔶 <br> WIP  |       ❌        |
-| [.NET client](https://github.com/meilisearch/meilisearch-dotnet)               | ✅          | ✅      |        ❌   |       ✅        |
+| [Swift client](https://github.com/meilisearch/meilisearch-swift)              | ✅          | ✅      |   ✅  |       ❌        |
+| [.NET client](https://github.com/meilisearch/meilisearch-dotnet)               | ✅          | ✅      |        ✅   |       ✅        |
 | [Rust client](https://github.com/meilisearch/meilisearch-rust)             | ✅          | ❌      |        🔶 <br> WIP  |       ✅        |
 | [Go client](https://github.com/meilisearch/meilisearch-go)              | ✅          | ✅     |        ✅   |       ✅        |
 | [Dart client](https://github.com/meilisearch/meilisearch-dart)               | ✅          | ✅     |        ✅   |       ❌        |
-| [Symfony](https://github.com/meilisearch/meilisearch-symfony) | ✅ | ✅ | ❌ | ❌ |
+| [Symfony](https://github.com/meilisearch/meilisearch-symfony) | ✅ | ✅ | ✅ | ❌ |
 | [Django](https://roadmap.meilisearch.com/c/60-django) | ❌ | ✅ | ❌ | ❌ |
 | [Rails](https://github.com/meilisearch/meilisearch-rails) | ✅ | ✅ | 🔶 <br>WIP | ✅ ||
-| [Official Laravel Scout Support](https://github.com/laravel/scout) | ✅ | ✅ | ❌ | ❌ |
+| [Official Laravel Scout Support](https://github.com/laravel/scout) | ✅ | ✅ | ❌ <br> Available as a standalone module | ❌ <br> Available as a standalone module |
 | [UI Search Kit](https://github.com/meilisearch/instant-meilisearch) | ✅ | ✅ | ✅ | ✅ |
 | [Docsearch](https://github.com/meilisearch/docs-scraper) | ✅ | ✅ | ✅ | ❌ |
 | [Strapi](https://github.com/meilisearch/strapi-plugin-meilisearch) | ✅ | ✅ | ❌ | ❌ |
@@ -65,8 +65,8 @@ Can't find a client you'd like us to support? [Submit your idea or vote for it](
 
 |         | Meilisearch | Algolia | Typesense | Elasticsearch |
 |---|:---:|:----:|:---:|:---:|
-| Schemaless | ✅ | ✅ | 🔶 <br>Automatic schema detection is supported but needs to be specified | ✅ |
-| Nested field support | ✅ | ✅ | ❌ | ✅ |
+| Schemaless | ✅ | ✅ | ✅ <br>Automatic schema detection needs to be enabled | ✅ |
+| Nested field support | ✅ | ✅ | ✅ | ✅ |
 | Automatic document ID detection | ✅ | ❌ | ❌ | ❌ |
 | Native document formats | `JSON`, `NDJSON`, `CSV` | `JSON` | `JSON`, `NDJSON` | `JSON`, `NDJSON`, `CSV` |
 
@@ -75,20 +75,20 @@ Can't find a client you'd like us to support? [Submit your idea or vote for it](
 |   | Meilisearch | Algolia | Typesense | Elasticsearch |
 |---|:---:|:----:|:---:|:---:|
 | Typo tolerant  | ✅ | ✅ | ✅ | 🔶 <br>Needs to be specified by fuzzy queries |
-| Orderable ranking rules | ✅ | ✅ | 🔶 <br>Tie-breaking order is limited by a unique scoring rule | ❌|
-| Custom rules | ✅ | ✅ | 🔶 <br>Limited to one default sorting rule | 🔶 <br>Function score query
+| Orderable ranking rules | ✅ | ✅ | 🔶 <br>Field priority can be changed, but ranking rules order cannot be changed. | ❌|
+| Custom rules | ✅ | ✅ | ✅ | 🔶 <br>Function score query
 | Query field weights | ✅ | ✅ | ✅ | ✅ |
 | Synonyms | ✅ | ✅ | ✅ | ✅ |
 | Stop words | ✅ | ✅ | ❌ | ✅ |
 | Automatic language detection | ✅ | ✅ | ❌ | ❌ |
-| All language supports | ✅ | ✅ | ❌ <br> Only space separated | ✅ |
+| All language supports | ✅ | ✅ | ✅ | ✅ |
 
 ##### Security
 
 |   | Meilisearch | Algolia | Typesense | Elasticsearch |
 |---|:---:|:----:|:---:|:---:|
 | API Key Management | ✅ | ✅ | ✅ | ✅ |
-| Tenant tokens & multi-tenant indexes | ✅ <br> [Multitenancy support](/learn/security/tenant_tokens.md) | ✅  <br> Hard filters are not configurable per index for an end-user tenant key | ✅ <br> Hard filters are not configurable per index for an end-user tenant key | ✅ <br> Role-based |
+| Tenant tokens & multi-tenant indexes | ✅ <br> [Multitenancy support](/learn/security/tenant_tokens.md) | ✅  <br> Hard filters are not configurable per index for an end-user tenant key | ✅ | ✅ <br> Role-based |
 
 ##### Search
 
@@ -96,10 +96,10 @@ Can't find a client you'd like us to support? [Submit your idea or vote for it](
 |---|:---:|:----:|:---:|:---:|
 | Placeholder search | ✅ | ✅ | ✅ | ✅ |
 | Multi-index search | [2023](https://github.com/meilisearch/product/issues/74) | ✅ | ✅ | ✅ |
-| Exact phrase search | ✅ | ✅ | ❌ | ✅ |
+| Exact phrase search | ✅ | ✅ | ✅ | ✅ |
 | Geo search |  ✅  | ✅ | ✅ | ✅ |
-| Sort by  |  ✅  | 🔶 <br> Limited to one `sort_by` rule per index. Indexes may have to be duplicated for each sort field and sort order | 🔶 <br> Does not support sort on string field | ✅ |
-| Filtering |  ✅ <br> Support complex filter queries with an SQL-like syntax.  | 🔶 <br> Does not support `OR` operation across multiple fields | 🔶 <br> Does not support `OR` operation across multiple fields | ✅ |
+| Sort by  |  ✅  | 🔶 <br> Limited to one `sort_by` rule per index. Indexes may have to be duplicated for each sort field and sort order | ✅ | ✅ |
+| Filtering |  ✅ <br> Support complex filter queries with an SQL-like syntax.  | 🔶 <br> Does not support `OR` operation across multiple fields | ✅ | ✅ |
 | Faceted search |  ✅ | ✅ | ✅ | ✅ |
 | Distinct attributes <br><div style="color:#A9A9A9;font-size:0.9em;">De-duplicate documents by a field value</div>| ✅ | ✅ | ✅  | ✅ |
 | Grouping <br><div style="color:#A9A9A9;font-size:0.9em;">Bucket documents by field values</div> | ❌ | ✅ | ✅  | ✅ |
@@ -120,7 +120,7 @@ Can't find a client you'd like us to support? [Submit your idea or vote for it](
 | High availability | Available with [Meilisearch Cloud](https://cloud.meilisearch.com/register) | ✅ | ✅ | ✅ |
 | Run-time dependencies | None | N/A | None | None |
 | Backward compatibility | ✅ | N/A | ✅ | ✅ |
-| Upgrade path | Documents need to be reindexed | N/A  | Documents need to be reindexed | Documents need to be reindexed |
+| Upgrade path | Documents need to be reindexed | N/A  | Documents need not be reindexed  | Documents need to be reindexed |
 
 ### Limits
 


learn/what_is_meilisearch/comparison_to_alternatives.md
```diff
@@ -65,8 +65,8 @@ Can't find a client you'd like us to support? [Submit your idea or vote for it](
 
 |         | Meilisearch | Algolia | Typesense | Elasticsearch |
 |---|:---:|:----:|:---:|:---:|
-| Schemaless | ✅ | ✅ | 🔶 <br>Automatic schema detection is supported but needs to be specified | ✅ |
-| Nested field support | ✅ | ✅ | ❌ | ✅ |
+| Schemaless | ✅ | ✅ | ✅ <br>Automatic schema detection needs to be enabled | ✅ |
+| Nested field support | ✅ | ✅ | ✅ | ✅ |
```
jasonbosco, line 69: Launched in v0.24.0

learn/what_is_meilisearch/comparison_to_alternatives.md
```diff
@@ -75,31 +75,31 @@ Can't find a client you'd like us to support? [Submit your idea or vote for it](
 |   | Meilisearch | Algolia | Typesense | Elasticsearch |
 |---|:---:|:----:|:---:|:---:|
 | Typo tolerant  | ✅ | ✅ | ✅ | 🔶 <br>Needs to be specified by fuzzy queries |
-| Orderable ranking rules | ✅ | ✅ | 🔶 <br>Tie-breaking order is limited by a unique scoring rule | ❌|
-| Custom rules | ✅ | ✅ | 🔶 <br>Limited to one default sorting rule | 🔶 <br>Function score query
+| Orderable ranking rules | ✅ | ✅ | 🔶 <br>Field priority can be changed, but ranking rules order cannot be changed. | ❌|
```
jasonbosco, line 78: I initially didn't understand what you meant by the earlier statement, but I read up on this and I get it now :)

So I've updated the description to be more precise.

learn/what_is_meilisearch/comparison_to_alternatives.md
```diff
@@ -75,31 +75,31 @@ Can't find a client you'd like us to support? [Submit your idea or vote for it](
...
+| Custom rules | ✅ | ✅ | ✅ | 🔶 <br>Function score query
 | Query field weights | ✅ | ✅ | ✅ | ✅ |
 | Synonyms | ✅ | ✅ | ✅ | ✅ |
 | Stop words | ✅ | ✅ | ❌ | ✅ |
 | Automatic language detection | ✅ | ✅ | ❌ | ❌ |
-| All language supports | ✅ | ✅ | ❌ <br> Only space separated | ✅ |
```
jasonbosco, line 84: Launched in v0.24.0

learn/what_is_meilisearch/comparison_to_alternatives.md
```diff
@@ -75,31 +75,31 @@ Can't find a client you'd like us to support? [Submit your idea or vote for it](
...
+| All language supports | ✅ | ✅ | ✅ | ✅ |
 
 ##### Security
 
 |   | Meilisearch | Algolia | Typesense | Elasticsearch |
 |---|:---:|:----:|:---:|:---:|
 | API Key Management | ✅ | ✅ | ✅ | ✅ |
-| Tenant tokens & multi-tenant indexes | ✅ <br> [Multitenancy support](/learn/security/tenant_tokens.md) | ✅  <br> Hard filters are not configurable per index for an end-user tenant key | ✅ <br> Hard filters are not configurable per index for an end-user tenant key | ✅ <br> Role-based |
```
jasonbosco, line 91: We addressed this in v0.23

learn/what_is_meilisearch/comparison_to_alternatives.md
```diff
@@ -75,31 +75,31 @@ Can't find a client you'd like us to support? [Submit your idea or vote for it](
...
 ##### Search
 
 |   | Meilisearch | Algolia | Typesense | Elasticsearch |
 |---|:---:|:----:|:---:|:---:|
 | Placeholder search | ✅ | ✅ | ✅ | ✅ |
 | Multi-index search | [2023](https://github.com/meilisearch/product/issues/74) | ✅ | ✅ | ✅ |
-| Exact phrase search | ✅ | ✅ | ❌ | ✅ |
+| Exact phrase search | ✅ | ✅ | ✅ | ✅ |
 | Geo search |  ✅  | ✅ | ✅ | ✅ |
-| Sort by  |  ✅  | 🔶 <br> Limited to one `sort_by` rule per index. Indexes may have to be duplicated for each sort field and sort order | 🔶 <br> Does not support sort on string field | ✅ |
```
jasonbosco, line 101: Addressed this in v0.23

learn/what_is_meilisearch/comparison_to_alternatives.md
```diff
@@ -75,31 +75,31 @@ Can't find a client you'd like us to support? [Submit your idea or vote for it](
...
-| Filtering |  ✅ <br> Support complex filter queries with an SQL-like syntax.  | 🔶 <br> Does not support `OR` operation across multiple fields | 🔶 <br> Does not support `OR` operation across multiple fields | ✅ |
```
jasonbosco, line 102: Addressed this in v0.24

learn/what_is_meilisearch/comparison_to_alternatives.md
```diff
@@ -75,31 +75,31 @@ Can't find a client you'd like us to support? [Submit your idea or vote for it](
...
+| Sort by  |  ✅  | 🔶 <br> Limited to one `sort_by` rule per index. Indexes may have to be duplicated for each sort field and sort order | ✅ | ✅ |
+| Filtering |  ✅ <br> Support complex filter queries with an SQL-like syntax.  | 🔶 <br> Does not support `OR` operation across multiple fields | ✅ | ✅ |
```
jasonbosco, line 102: Addressed this in v0.24

learn/what_is_meilisearch/comparison_to_alternatives.md
```diff
@@ -120,7 +120,7 @@ Can't find a client you'd like us to support? [Submit your idea or vote for it](
 | High availability | Available with [Meilisearch Cloud](https://cloud.meilisearch.com/register) | ✅ | ✅ | ✅ |
 | Run-time dependencies | None | N/A | None | None |
 | Backward compatibility | ✅ | N/A | ✅ | ✅ |
-| Upgrade path | Documents need to be reindexed | N/A  | Documents need to be reindexed | Documents need to be reindexed |
+| Upgrade path | Documents need to be reindexed | N/A  | Documents need not be reindexed  | Documents need to be reindexed |
```
jasonbosco, line 123: This has always been the case with Typesense. Users just replace binary and restart the process. Documents need not be sent over again via the API.

learn/what_is_meilisearch/comparison_to_alternatives.md
Kerollmops: However, the number of elements specified in a `sort_by` is limited to 3 max. [Am I right](https://typesense.org/docs/0.24.0/api/search.html#ranking-and-sorting-parameters)?

Kerollmops: We will specify that the `sort_by` on strings consumes a lot of memory, [according to the documentation](https://typesense.org/docs/0.24.0/api/search.html#sorting-on-strings).

jasonbosco: > However, the number of elements specified in a sort_by is limited to 3 max

That's correct. 

> sort_by on strings consumes a lot of memory

This is only true for large string fields like say blog posts, description fields, etc.

learn/what_is_meilisearch/comparison_to_alternatives.md
Kerollmops: Reindexing in our documentation means that the engine processes the documents to compute optimized data structure to search in them efficiently. Therefore, Typesense always reindexes the documents when the binary is updated and even every time the binary is started. Am I right?

jasonbosco: By your definition of reindexing, meaning that the engine is the one that's doing the indexing, that's right - Typesense does reindex automatically on version upgrades, internally.

In our documentation, we refer to reindexing as the process of users sending documents via the API.

learn/what_is_meilisearch/comparison_to_alternatives.md
```diff
@@ -75,31 +75,31 @@ Can't find a client you'd like us to support? [Submit your idea or vote for it](
 |   | Meilisearch | Algolia | Typesense | Elasticsearch |
 |---|:---:|:----:|:---:|:---:|
 | Typo tolerant  | ✅ | ✅ | ✅ | 🔶 <br>Needs to be specified by fuzzy queries |
-| Orderable ranking rules | ✅ | ✅ | 🔶 <br>Tie-breaking order is limited by a unique scoring rule | ❌|
-| Custom rules | ✅ | ✅ | 🔶 <br>Limited to one default sorting rule | 🔶 <br>Function score query
+| Orderable ranking rules | ✅ | ✅ | 🔶 <br>Field priority can be changed, but ranking rules order cannot be changed. | ❌|
+| Custom rules | ✅ | ✅ | ✅ | 🔶 <br>Function score query
```
Kerollmops, line 79: Isn't Typesense limited to 3 `sort_by` rules max?

jasonbosco: That is only for hard-sorting. 

We handle ranking rules using query parameters which are not limited: 

https://typesense.org/docs/0.24.0/api/search.html#ranking-and-sorting-parameters
https://typesense.org/docs/0.24.0/api/search.html#typo-tolerance-parameters

learn/what_is_meilisearch/comparison_to_alternatives.md
Kerollmops: I also read that sometimes, Typesense can be slow to answer as it does sort at query time [there even are workaround to this](https://github.com/typesense/typesense/issues/575#issuecomment-1096591171). We will inform users about this.

> Since Typesense dynamically sorts the results at query time, for stop words like `the` which occurs in a lot of documents, a large number of documents need to be processed to fetch results.
> 
> In the future, we plan to introduce a way to fetch documents (especially these types of single word searches) in a pre-sorted fashion. For now, removing such stop words or pre-caching them with a long TTL is the work around.

jasonbosco: Sure, feel free to mention this, but could you add that this only applies if the number of results fetched for a particular query is in the 10s of millions (count) range. This is not an issue if the number of results fetched are smaller than that.

Kerollmops: That's unfortunately not true. [As an example](https://songs-search.typesense.org/?s%5Bquery%5D=%2A&s%5BrefinementList%5D%5Bprimary_artist_name%5D%5B0%5D=Various%20Artists&s%5Bmenu%5D%5Brelease_decade%5D=2010s), when we search for less than 2 million documents the response time is pretty high (4s).

Capture d’écran 2023-02-16 à 23 47 03

jasonbosco: In your screenshot, it says the number of results returned are 16.9M... So what I said still holds true: 

> this only applies if the number of results fetched for a particular query is in the 10s of millions (count) range

In your example, sorting ~17M records on the fly is indeed not as fast as we'd like it to be. But for smaller datasets below say 5M-10M should still be fast...

Kerollmops: Hum... no, my screenshot shows 1,695,074 documents, so 1,6M. That is not near what your explaining in your previous comment. And my second example shows an even worst case where Typesense answers in more than 2 second to sort half a million documents (660k).

learn/what_is_meilisearch/comparison_to_alternatives.md
ferdi05: I'd like to change `Custom rules` to `Custom ranking rules` as I don't find explicit enough.
wdyt @gmourier ?

learn/what_is_meilisearch/comparison_to_alternatives.md
Kerollmops: Is that also true that, [according to our definition of indexing](https://en.wikipedia.org/wiki/Search_engine_indexing), Typesense does re-index the documents when is started?

learn/what_is_meilisearch/comparison_to_alternatives.md
ferdi05: I'm not even sure it needs to be specified, `✅` would be enough I think

learn/what_is_meilisearch/comparison_to_alternatives.md
ferdi05: I cannot review this, we don't use `Field priority`.
@gmourier is this a common idiom in the search engine industry? Otherwise what would you use so that readers who are more familiar with Meilisearch will understand what is written here?

learn/what_is_meilisearch/comparison_to_alternatives.md
ferdi05: @gmourier I don't remember why we didn't put a simple checkmark there. How about changing it?

learn/what_is_meilisearch/comparison_to_alternatives.md
ferdi05: @jasonbosco would you mind updating the PR with the 3 sort fields limitation.
@Kerollmops I'm not sure we need to specify that `sort_by` consumes a lot of memory, it seems like it only happens if Typesense is misused

learn/what_is_meilisearch/comparison_to_alternatives.md
ferdi05: I indeed think warning readers about this unfortunate slowdown is required.

learn/what_is_meilisearch/comparison_to_alternatives.md
ferdi05: As we're in a Meilisearch website, it makes sense to use Meilisearch definition of reindexing.

Kerollmops: > In our documentation, we refer to reindexing as the process of users sending documents via the API.

So, why did you write that we need that in you comparison table, the user doesn't need to send the documents via the API. Can you remove this sentence, please?

> Version upgrades require re-indexing.

learn/what_is_meilisearch/comparison_to_alternatives.md
jasonbosco: Ah hmm... My bad. I misread the number in your screenshot. 

So yeah fair enough. Feel free to mention that for wild card queries `q=*` (which I believe you call "Placeholder search"), sorting is slow since it's done on the fly at runtime.

learn/what_is_meilisearch/comparison_to_alternatives.md
jasonbosco: > Is that also true that, [according to our definition of indexing](https://en.wikipedia.org/wiki/Search_engine_indexing), Typesense does re-index the documents every time it is started/run as the data-structures are in memory?

That's correct. But Typesense does this automatically internally.

> So, why did you write that we need that in the comparison table of Typesense? Users don't need to send the documents via the API. Can you remove the following sentence, please?
> 
> > Version upgrades require re-indexing.

I just read your docs [here](https://docs.meilisearch.com/learn/advanced/snapshots_vs_dumps.html#dumps) and realized that I've been thinking about your "snapshots" feature all this while. I didn't realize you launched the "dumps" feature which is compatible across version. Good to know, I'll update our table. 

> Could you also update the Elastic Search part about backward compatibility, [they are not](https://discuss.elastic.co/t/is-is-possible-to-run-elasticsearch-6-in-a-backward-compatible-mode/113211/3).

Good to know. I'll update.

> BTW, where did you get the information about the internal Algolia index backward compatibility? There is no information about that on the internet.

Since they operate as a SaaS service, they most likely manage backward/forward compatibility internally and don't expose it to users. Since users don't have to worry about their internal index versions, I marked them as "Fully backward compatible" at least from an end user's perspective.

learn/what_is_meilisearch/comparison_to_alternatives.md
jasonbosco: Looks like the equivalent terminology in Meilisearch is [`Attribute Ranking Order`](https://docs.meilisearch.com/learn/core_concepts/relevancy.html#attribute-ranking-order).

learn/what_is_meilisearch/comparison_to_alternatives.md
jasonbosco: > @jasonbosco would you mind updating the PR with the 3 sort fields limitation.

Done.

learn/what_is_meilisearch/comparison_to_alternatives.md
Kerollmops: Hum... what I understand is that Typesense is slow whenever there is a lot of documents to sort. Therefore, it can happen on a placeholder search, a placeholder search with filters (☝️), or any other requests that contains common words i.e. stop words, URL prefixes, URL domain names.

So, I would rather prefer that we explain that Typesense have performance issues for when more than 700k documents must be sorted.

learn/what_is_meilisearch/comparison_to_alternatives.md
Kerollmops: Thank you very much Jason.

BTW dumps are available since nearly the beginning. Maybe pre-v0.20, so at least Mar 30 2021.

You should probably put _N/A_ for Algolia, as the category specifically talk about a technical aspect that we can't verify. What do you think?

learn/what_is_meilisearch/comparison_to_alternatives.md
jasonbosco: > what I understand is that Typesense is slow whenever there is a lot of documents to sort

You're right. That's the general case.

So to be more precise, I would say "When >700K results match for a given search query, queries might take several seconds, due to run-time sorting".

learn/what_is_meilisearch/comparison_to_alternatives.md
jasonbosco: > You should probably put N/A for Algolia, as the category specifically talk about a technical aspect that we can't verify. What do you think?

I was just thinking the same! I'll add a note that they do this internally automatically.

jasonbosco: Just found this in the [Elasticsearch docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/setup-upgrade.html#upgrade-index-compatibility):

> Elasticsearch has full query and write support for indices created in the previous major version. If you have indices created in 6.x or earlier, you might use the [archive functionality](https://www.elastic.co/guide/en/elasticsearch/reference/current/archive-indices.html) to import them into newer Elasticsearch versions, or you must reindex or delete them before upgrading to 8.6.2

So they are backward compatible up to 1 major version.

learn/what_is_meilisearch/comparison_to_alternatives.md
gmourier: Yes, it seems clearer!

learn/what_is_meilisearch/comparison_to_alternatives.md
gmourier: I think it was when we released the tenant token feature; at the time v0.23 of Typesense was not released yet, and therefore, it was not possible to generate granular rules **for different indexes within a single JWT**.

We could rephrase it to make it clearer, Algolia does not have that capability.

From what I read from [typesense doc](https://typesense.org/docs/0.24.0/api/api-keys.html#generate-scoped-search-key), this is not possible to define different `filter_by` rules per `index A` and `indexB` on a Scoped Search Key. Am I right @jasonbosco? 

That's what we wanted to emphasize by saying, _"Hard filters are not configurable per index for an end-user tenant key"_

Thanks!

learn/what_is_meilisearch/comparison_to_alternatives.md
gmourier: `Field priority` seems ok to me; maybe `Field weight` could be clearer?

learn/what_is_meilisearch/comparison_to_alternatives.md
jasonbosco: @gmourier As of v0.23, you can embed a separate scoped API key inside each search inside a multi index search request, and have different filters for each index.

learn/what_is_meilisearch/comparison_to_alternatives.md
Kerollmops: Can we continue [this conversation, here](https://github.com/meilisearch/documentation/pull/2173#issuecomment-1435000867)?

learn/what_is_meilisearch/comparison_to_alternatives.md
Kerollmops: About the original comment again. Could you please update the sentence to something like: « documents will be automatically reindexed » for Meilisearch, Typesense and ES, please?

learn/what_is_meilisearch/comparison_to_alternatives.md
Kerollmops: After refining a little bit more the results and processing time of Typesense I found out that the engine [takes more than 3 seconds to sort approximately 28,000 documents](https://songs-search.typesense.org/?s%5Bquery%5D=%2A&s%5BrefinementList%5D%5Bprimary_artist_name%5D%5B0%5D=Пётр%20Ильич%20Чайковский). I would rather prefer that you write something like: « When >25K results match for a given search query, queries might take several seconds, due to run-time sorting », please.

learn/what_is_meilisearch/comparison_to_alternatives.md
jasonbosco: Done.

learn/what_is_meilisearch/comparison_to_alternatives.md
jasonbosco: I did a bit more digging and my suspicion is that it's actually not the sorting that's the main cause of the slowness for these examples you shared. 

For eg, here's one example where ~1M documents are retrieved and sorted in 949ms:
Screen Shot 2023-02-18 at 4 09 24 PM

Here's another example where ~600K documents are retrieved and sorted in 478ms:
Screen Shot 2023-02-18 at 4 24 51 PM

So I take back what I said [here](https://github.com/meilisearch/documentation/pull/2174#discussion_r1110171297) - my general statement there does not hold true based on my examples above. 

It's likely a combination of filtering, placeholder search and faceting that's causing majority of the slowdown in the examples you shared. 

In your latest example, 3s to sort 28K documents definitely sounds more like a bug or edge case we probably didn't account for, given the shape of this particular dataset and query. We're going to look deeper into why that particular query is slow.

(Thank you for helping uncover these cases btw!)

In the meantime, I'd still be comfortable if you want to say that placeholder search or 1-3 letter prefix searches with filtering and faceting enabled might take several seconds when results fetched are over say 700K... But based on the examples I shared above, I don't think we can make a general statement about runtime sorting being slow.

Kerollmops: > In your latest example, 3s to sort 28K documents definitely sounds more like a bug or edge case we probably didn't account for, given the shape of this particular dataset and query.

There is not a [single](https://songs-search.typesense.org/?s%5Bquery%5D=%2A&s%5BrefinementList%5D%5Bprimary_artist_name%5D%5B0%5D=Johannes%20Brahms) [example](https://songs-search.typesense.org/?s%5Bquery%5D=%2A&s%5BrefinementList%5D%5Bprimary_artist_name%5D%5B0%5D=Franz%20Schubert) [that](https://songs-search.typesense.org/?s%5Bquery%5D=%2A&s%5BrefinementList%5D%5Bprimary_artist_name%5D%5B0%5D=Pearl%20Jam) [shows](https://songs-search.typesense.org/?s%5Bquery%5D=%2A&s%5BrefinementList%5D%5Bprimary_artist_name%5D%5B0%5D=Fryderyk%20Chopin) [the](https://songs-search.typesense.org/?s%5Bquery%5D=%2A&s%5BrefinementList%5D%5Bprimary_artist_name%5D%5B0%5D=Bob%20Dylan) [issue](https://songs-search.typesense.org/?s%5Bquery%5D=%2A&s%5BrefinementList%5D%5Bprimary_artist_name%5D%5B0%5D=J.%20K.%20Rowling). I am not sure that it is a bug or edge case but more a design issue, here.

> In the meantime, I'd still be comfortable if you want to say that placeholder search or 1-3 letter prefix searches with filtering and faceting enabled might take several seconds when results fetched are over say 700K...

Can we agree that the bound is way lower than 700k but more below 25k?

jasonbosco: Found some time to dig into this in depth. 

Here's what's happening: the search time you see displayed in the UI is actually the SUM of multiple search queries, embedded inside a single multi_search request (all these searches are required to show the facet counts in the RefinementList widgets). 

If you open up the network tab and view the actual request/response to Typesense, you'll notice a single multi_search request, but there are multiple search queries inside that request.

For eg, here's a seemingly bad one: filtering by [`Pearl Jam`](https://songs-search.typesense.org/?s%5Bquery%5D=%2A&s%5BrefinementList%5D%5Bprimary_artist_name%5D%5B0%5D=Pearl%20Jam) says: 29,935 results found - Searched 31,931,260 songs in 3400ms.

Now if you look at the network request, inside the single `multi_search` request, there are two search queries generated:

Screen Shot 2023-02-20 at 12 31 26 AM

Here's the corresponding responses for each search query:

Screen Shot 2023-02-20 at 12 32 40 AM

**Notice how the results for the first query which returns ~30K results is actually just 16ms**. The query that's actually slow is the one that facets on all the fields, across all 32M records, which takes about 3.4s.

This is the common thread I found across all the examples you shared. The actual search request itself is fast, but then there are other facet_by queries that need to be made to populate the facet counts for all the refinement list widgets (since they all need to be recomputed with each filter), and these are the time-consuming requests. 

*Side note:* these refinement list widgets used in the demo come from Instantsearch.js, so you will see the same set of requests generated even when you use these with Meilisearch and your adapter for Instantsearch.js. I wouldn't be surprised if you also see a 1s "search time" if you sum up the processing time of all the searches required to show results from a 28M record dataset, using 5 refinementList widgets. 

---

Here's a separate issue you helped uncover with [filtering](https://songs-search.typesense.org/?s%5Bquery%5D=%2A&s%5BrefinementList%5D%5Bgenres%5D%5B0%5D=Rock&s%5BrefinementList%5D%5Bprimary_artist_name%5D%5B0%5D=Various%20Artists&s%5Bmenu%5D%5Brelease_decade%5D=1970s). 

That example you shared currently does this multi_search request, with 4 different individual searches:

```
curl -s 'https://qtg5aekc2iosjh93p.a1.typesense.net/multi_search?use_cache=true&x-typesense-api-key=8hLCPSQTYcBuK29zY5q6Xhin7ONxHy99' \
  --data-raw '{
    "searches":[
        {"query_by":"primary_artist_name,title,album_name","query_by_weights":"2,2,1","sort_by":"_text_match(buckets: 10):desc","highlight_full_fields":"primary_artist_name,title,album_name","collection":"s","q":"*","facet_by":"genres,primary_artist_name,release_group_types,country,release_decade","filter_by":"genres:=[`Rock`] && primary_artist_name:=[`Various Artists`] && release_decade:=[`1970s`]","max_facet_values":20,"page":1,"per_page":15},
        {"query_by":"primary_artist_name,title,album_name","query_by_weights":"2,2,1","sort_by":"_text_match(buckets: 10):desc","highlight_full_fields":"primary_artist_name,title,album_name","collection":"s","q":"*","facet_by":"genres","filter_by":"primary_artist_name:=[`Various Artists`] && release_decade:=[`1970s`]","max_facet_values":20,"page":1,"per_page":1},
        {"query_by":"primary_artist_name,title,album_name","query_by_weights":"2,2,1","sort_by":"_text_match(buckets: 10):desc","highlight_full_fields":"primary_artist_name,title,album_name","collection":"s","q":"*","facet_by":"primary_artist_name","filter_by":"genres:=[`Rock`] && release_decade:=[`1970s`]","max_facet_values":20,"page":1,"per_page":1},
        {"query_by":"primary_artist_name,title,album_name","query_by_weights":"2,2,1","sort_by":"_text_match(buckets: 10):desc","highlight_full_fields":"primary_artist_name,title,album_name","collection":"s","q":"*","facet_by":"release_decade","filter_by":"genres:=[`Rock`] && primary_artist_name:=[`Various Artists`]","max_facet_values":20,"page":1,"per_page":1}]}' \
  | jq '.results[].search_time_ms'
761
645
358
626
```

By default the UI does an exact filtering `:=` on all the fields. Now notice how the search times change if I change it to `:` filtering:

```
curl -s 'https://qtg5aekc2iosjh93p.a1.typesense.net/multi_search?use_cache=true&x-typesense-api-key=8hLCPSQTYcBuK29zY5q6Xhin7ONxHy99' \
  --data-raw '{
    "searches":[
        {"query_by":"primary_artist_name,title,album_name","query_by_weights":"2,2,1","sort_by":"_text_match(buckets: 10):desc","highlight_full_fields":"primary_artist_name,title,album_name","collection":"s","q":"*","facet_by":"genres,primary_artist_name,release_group_types,country,release_decade","filter_by":"genres:[`Rock`] && primary_artist_name:[`Various Artists`] && release_decade:[`1970s`]","max_facet_values":20,"page":1,"per_page":15},
        {"query_by":"primary_artist_name,title,album_name","query_by_weights":"2,2,1","sort_by":"_text_match(buckets: 10):desc","highlight_full_fields":"primary_artist_name,title,album_name","collection":"s","q":"*","facet_by":"genres","filter_by":"primary_artist_name:[`Various Artists`] && release_decade:[`1970s`]","max_facet_values":20,"page":1,"per_page":1},
        {"query_by":"primary_artist_name,title,album_name","query_by_weights":"2,2,1","sort_by":"_text_match(buckets: 10):desc","highlight_full_fields":"primary_artist_name,title,album_name","collection":"s","q":"*","facet_by":"primary_artist_name","filter_by":"genres:[`Rock`] && release_decade:[`1970s`]","max_facet_values":20,"page":1,"per_page":1},
        {"query_by":"primary_artist_name,title,album_name","query_by_weights":"2,2,1","sort_by":"_text_match(buckets: 10):desc","highlight_full_fields":"primary_artist_name,title,album_name","collection":"s","q":"*","facet_by":"release_decade","filter_by":"genres:[`Rock`] && primary_artist_name:[`Various Artists`]","max_facet_values":20,"page":1,"per_page":1}]}' \
    | jq '.results[].search_time_ms'
273
263
147
250
```

This different in performance is a bug, which we already have a fix for in mind. We'll be pushing this fix out in 0.24.1.rc in the next week or two. 

---

Long story short:

It is not sorting that's slow, but _faceting_ on multiple fields on large datasets. I'd be ok with you saying that faceting on multiple fields on 10s of millions of records can take a couple of seconds. While we're going to look into how to improve this, this is not a bug - so feel free to mention this.

learn/what_is_meilisearch/comparison_to_alternatives.md
dichotommy: `Field weight` seems like the best option, since it returns the most relevant results on Google (and that's where confused readers tend to go, in my experience).

ferdi05: Thanks for your input @gmourier @dichotommy 
@jasonbosco would you mind using `Field weight`? thx

learn/what_is_meilisearch/comparison_to_alternatives.md
ferdi05: I'm actually unsure we want to have the conversation there, as it's a deleted PR now.
Anyway here's where we were

> @Kerollmops
> 
> > I read feedback about a [not-so-great support of Chinese](https://github.com/typesense/typesense/issues/52#issuecomment-1356140139). We work hard at Meilisearch to improve the support of CJK so maybe we can add something about the support of it on the Typesense side, i.e. in progress. What do you think?

>@jasonbosco 
>> 
>> Ah missed seeing that issue. Sure, feel free to mention that it's being improved.
>> 
>> We've been improving support as and when we get feedback from native speakers for different scenarios.

I'm actually fine just leaving check marks to everyone, as @jasonbosco suggested.

learn/what_is_meilisearch/comparison_to_alternatives.md
ferdi05: I'm not even sure what a `Hard filter` is, so I can't help here.

learn/what_is_meilisearch/comparison_to_alternatives.md
jasonbosco: Done.

learn/what_is_meilisearch/comparison_to_alternatives.md
Kerollmops: Thank you very much for the time to dive into this issue. I took a little bit of time on my side too by trying another demo of Typesense, this time with a lot fewer documents in it i.e. 2 million.

> It is not sorting that's slow, but faceting on multiple fields on large datasets. I'd be ok with you saying that faceting on multiple fields on 10s of millions of records can take a couple of seconds. While we're going to look into how to improve this, this is not a bug - so feel free to mention this.

I absolutely agree with what you are saying here. However, it looks like the bounds are not exactly what I would have defined, but, again, I would have decided on a lower one. In [my example below](https://recipe-search.typesense.org/?r%5Bquery%5D=%2A) I ran a placeholder search on 2 millions documents and the facet retrieval took 730ms. Therefore, I would prefer to say that Typesense takes several seconds to answer when is needs to fetch facets on less than 10 million documents.

Capture d’écran 2023-02-20 à 21 33 33
Capture d’écran 2023-02-20 à 21 33 27

What do you think? Could you update the "Faceted Search" category for Typesense to add something like:
> Faceting on a single field with less than 10 million documents can take several seconds.

jasonbosco: If you want to get more precise, it's really the number of unique values in the faceted field that determines how fast/slow facets are computed. So even though there are only 2M docs in this recipes example, the ingredients field is an array field and the number of unique facet values in this field across all 2M docs is in the 10s of millions range. 

```
curl -s 'https://qtg5aekc2iosjh93p.a1.typesense.net/multi_search?x-typesense-api-key=8hLCPSQTYcBuK29zY5q6Xhin7ONxHy99' \
  --data-raw '{"searches":[{"query_by":"title","sort_by":"","highlight_full_fields":"title","collection":"r","q":"*","facet_by":"ingredient_names","filter_by":"","max_facet_values":100,"page":1,"per_page":15}]}' \
  --compressed | jq '.results[0].facet_counts[0].counts[].count' | jq -s 'add'
11390471
```

So I still stand by my 10s of millions statement. If you want to refine it, you can say faceting on over 10s of millions of unique field values can take a couple of seconds.

Kerollmops: Unfortunately, this is different from what I witnessed below. There only is less than 6 million facet values, and Typesense takes more than 2 seconds to process them.
```
curl -s 'https://qtg5aekc2iosjh93p.a1.typesense.net/multi_search?x-typesense-api-key=8hLCPSQTYcBuK29zY5q6Xhin7ONxHy99' --data-raw '{"searches":[{"query_by":"primary_artist_name,title,album_name","query_by_weights":"2,2,1","sort_by":"","highlight_full_fields":"primary_artist_name,title,album_name","collection":"s","q":"*","facet_by":"primary_artist_name","max_facet_values":20,"page":1,"per_page":1}]}' --compressed | jq '.results[0].facet_counts[0].counts[].count' | jq -s 'add'
5933022
```

jasonbosco: `max_facet_values` is set to `20` in your query. So it's only returning the top 20 facets values in the API response, so you're counting a subset of the data. For eg, I you update `max_facet_values: 20000`, you'll still see ~2s response time, and the count is in the 10s of millions:

```
curl -s 'https://qtg5aekc2iosjh93p.a1.typesense.net/multi_search?x-typesense-api-key=8hLCPSQTYcBuK29zY5q6Xhin7ONxHy99' --data-raw '{"searches":[{"query_by":"primary_artist_name,title,album_name","query_by_weights":"2,2,1","sort_by":"","highlight_full_fields":"primary_artist_name,title,album_name","collection":"s","q":"*","facet_by":"primary_artist_name","max_facet_values":20000,"page":1,"per_page":1}]}' --compressed | jq '.results[0].facet_counts[0].counts[].count' | jq -s 'add'
19601674
```

Kerollmops: Thank you! So, what do you think of:
> Faceting can take several seconds when ~10 million facet values must be returned.

learn/what_is_meilisearch/comparison_to_alternatives.md
ferdi05: Hey @jasonbosco I saw your thumbs up. I guess this means you're happy with a green checkmark, so I will edit the comparison accordingly

learn/what_is_meilisearch/comparison_to_alternatives.md
jasonbosco: > Thank you! So, what do you think of:
> 
> > Faceting can take several seconds when ~10 million facet values must be returned.

Yeah, I'd be ok with that.

Kerollmops: Can you update this PR with the sentence above added to Typesense in the « Faceted Search » category, please?

jasonbosco: Done.

learn/what_is_meilisearch/comparison_to_alternatives.md
maryamsulemani97: @gmourier 
Does typesense embedding a separate scoped API key inside each search inside a multi index search request answer your question on different filters for different indexes?

learn/what_is_meilisearch/comparison_to_alternatives.md
jasonbosco: @Kerollmops May I know if Meilisearch estimates facet counts by default or if you return an exhaustive count by default?

Algolia for eg estimates facet counts by taking every third record and extrapolating. Do you do something similar?

In Typesense we return exhaustive facet counts by default, at least as of v0.24 and have the ability to turn on facet count estimation in the upcoming v0.25…

Curious to hear about Meilisearch’s approach.

Kerollmops: > May I know if Meilisearch estimates facet counts by default or if you return an exhaustive count by default?

Meilisearch returns accurate facet counts based on the estimated number of results we computed for a given query.

> Algolia for eg estimates facet counts by taking every third record and extrapolating. Do you do something similar?

What do you mean by _third record_?

learn/what_is_meilisearch/comparison_to_alternatives.md
gmourier: Sorry to be late, @maryamsulemani97 and @jasonbosco

It seems it's not possible to define for one scoped API key a custom rule per accessed index, which makes me think that using Typesense, you are forced to forge `X` scoped API keys per user. `X` is the number of indexes accessible to that user that need a custom rule.

Let's say that I want to define rules for `user 1` when querying `index A` and `index B`

With Meilisearch, I can do it this way when forging a tenant token. You don't need to generate different tenant tokens for each index per user.

```jsonc
{
  "apiKeyUid": ...,
  "exp": ...,
  "searchRules": {
    "indexA": {
      "filter": "user_id = 1 AND is_premium = true"
    },
    "indexB": {
      "filter": "user_id = 1 AND published = true",
    }
  }
}
```

So I would prefer to keep that differentiator in the comparative table. Maybe it's more straightforward to say that competitors require `X` scoped api keys per user if the `X` accessible indexes need a unique rule.

@jasonbosco can you confirm that statement here?

learn/what_is_meilisearch/comparison_to_alternatives.md
Kerollmops: Hey @jasonbosco, can we consider [this performance issue](https://github.com/typesense/typesense/issues/914#issuecomment-1452228304) and change the sentence accordingly?

> If atleast 100000 hits are found for the search query, we will use 10% of the hits to estimate facet counts.

This isn't efficient and doesn't speed up the queries. I am not sure that [the previous sentence we agreed on](https://github.com/meilisearch/documentation/pull/2174/files#r1113828010) is accurate.

jasonbosco: > If atleast 100000 hits are found for the search query, we will use 10% of the hits to estimate facet counts.

This only happens when the user turns facet estimation on using `facet_sample_threshold` and `facet_sample_percent` parameters (added new in 0.25.0.rc). By default, we do not estimate facets or results, so my previous statement still holds true. 

On a side note, in that particular issue, I had run out of RAM which is what resulted in the slowness I saw. I can't replicate the issue anymore.

learn/what_is_meilisearch/comparison_to_alternatives.md
jasonbosco: But that's just an implementation detail. "Tenant tokens" in Typesense are all generated client-side and so it's just one additional method call to generate an additional key for each index. You can then still send all of these keys in a single multi_search request. 

So the end result from a user's perspective is that they can apply "tenant scopes" at an index-level granularity both in Typesense and in Meilisearch. The exact syntax of how you achieve this end goal is what's different.

learn/what_is_meilisearch/comparison_to_alternatives.md
Kerollmops: Hey @jasonbosco,

>  I was able to just simulate direct curl commands similar to what the dynamic widgets would have made, and I was able to replicate the slowness. We'll look into this.

[There are performance issues on this subject](https://github.com/typesense/typesense/issues/914#issuecomment-1454236220). Could you add a line about that, please?

Just after you did your edit ☝️ I think we will be able to merge your PR. Thank you for making those changes. We are happy to describe Typesense better as it evolved all these past months.

learn/what_is_meilisearch/comparison_to_alternatives.md
Kerollmops: So, can a "key generator" generate a key on the client side without calling any HTTP method of Typesense and restricting the documents that the key sees? And is it therefore possible for the "key generator" to do that by just using a Typensense key and derivate those filtered-key from it?

learn/what_is_meilisearch/comparison_to_alternatives.md
jasonbosco: I can't add any other general statements to this PR, because this issue is isolated to this particular user's dataset and it looks like a bug.

For what it's worth, we handle 500M+ searches per month in Typesense Cloud, and I can confidently say that this is not a widespread issue for most users. 

So I'd appreciate not belaboring this topic and merging this PR, since the information that's currently public has factually incorrect statements about Typesense in other places.

learn/what_is_meilisearch/comparison_to_alternatives.md
jasonbosco: > So, can a "key generator" generate a key on the client side without calling any HTTP method of Typesense and restricting the documents that the key sees? 

Yes. But you need to use an existing search API-key as the base, and use that to cryptographically generate the other restricted keys, and these restricted keys don't need HTTP calls to be created, since they are cryptographically generated using the base search API-key as the seed. Users need to generate these restricted API keys (or really any API keys) on the server-side and send it to the client side. 

> And is it, therefore, possible for the "key generator" to do that by just using a Typensense key and derivate those filtered-keys from it?

Yes. But you need to have access to the base search API key to generate additional filtered keys, like I mentioned above. And in a multi-tenant use-case, users don't expose the base search API key to the frontend, instead expose the restricted API keys. 

It feels like we're going off on a tangent here about how these keys work in Typesense (again an implementation detail). To re-iterate the key point we were discussing in this thread previously: Typesense has the ability for users to apply "tenant scopes" at an index-level granularity and so the statement that is currently public is wrong / outdated.

learn/what_is_meilisearch/comparison_to_alternatives.md
Kerollmops: Let's merge that PR the way it is now and see if we need to make changes later.

learn/what_is_meilisearch/comparison_to_alternatives.md
Kerollmops: Yup! Let's merge this PR as it is now.

Kerollmops (approved):
Hey @maryamsulemani97 👋  I think this PR is good now and can be merged as it is. I don't have anything to add. I let you merge it whenever you want 😊

Thank you, @jasonbosco, for the changes!

maryamsulemani97 (approved):
Thank you @jasonbosco! 🙏 

bors merge

ferdi05:
Indeed, thanks @jasonbosco for helping us to make this more accurate!

PR #2175, opened 2023-02-16
Warnings against raising `maxTotalHits` limit

### LINKED ISSUE #2049: Warning about `maxTotalHits` impact on performance
The engine team wants to alert on the potential loss of performance a setting could bring: using a huge number for `maxTotalHits` (1000 or less is required) can impact the performance. Especially, when this change is combined to using the exhaustive pagination (with the `page` parameter), this can drastically increase the search time.

I think we should warn our users of this potential issue. Users should only change `maxTotalHits` if they accept to lose performance at the end (not every user expects Meilisearch to respond in less than 50ms)
- https://docs.meilisearch.com/reference/api/settings.html#pagination
- https://docs.meilisearch.com/learn/configuration/settings.html#pagination

Algolia warn their users as well for their same setting (`paginationLimitedTo`): https://www.algolia.com/doc/api-reference/api-parameters/paginationLimitedTo/

> We recommend that you keep the default value to guarantee excellent performance. Increasing the pagination limit will have a direct impact on the performance of search queries. A too high value will also make it very easy for anyone to retrieve (“scrape”) your entire dataset.

guimachiavelli:
Closes #2049

### DIFF
--- a/learn/advanced/pagination.md
+++ b/learn/advanced/pagination.md
@@ -8,13 +8,13 @@ In this guide, we discuss two different approaches to pagination supported by Me
 
 There are many UI patterns that help your users navigate through search results. One common and efficient solution in Meilisearch is using `offset` and `limit` to create interfaces centered around ["Previous" and "Next" buttons](#previous-and-next-buttons).
 
-Other solutions, such as [creating a page selector](/learn/advanced/pagination.md#numbered-page-selectors) allowing users to jump to any search results page, make use of `hitsPerPage` and `page` to obtain the exhaustive total number of matched documents.
+Other solutions, such as [creating a page selector](/learn/advanced/pagination.md#numbered-page-selectors) allowing users to jump to any search results page, make use of `hitsPerPage` and `page` to obtain the exhaustive total number of matched documents. Numbered page selectors may lead to decreased performance.
 
-Whatever UI you choose, there is a limited maximum number of search results Meilisearch will return for any given query. You can use [the `maxTotalHits` index setting](/reference/api/settings.md#pagination) to configure this, but be aware that higher limits may negatively impact search performance.
+Whatever UI you choose, there is a limited maximum number of search results Meilisearch will return for any given query. You can use [the `maxTotalHits` index setting](/reference/api/settings.md#pagination) to configure this, but be aware that higher limits will negatively impact search performance.
 
 ## "Previous" and "Next" buttons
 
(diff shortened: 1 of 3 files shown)


learn/advanced/pagination.md
```diff
@@ -8,13 +8,13 @@ In this guide, we discuss two different approaches to pagination supported by Me
...
 
-Whatever UI you choose, there is a limited maximum number of search results Meilisearch will return for any given query. You can use [the `maxTotalHits` index setting](/reference/api/settings.md#pagination) to configure this, but be aware that higher limits may negatively impact search performance.
+Whatever UI you choose, there is a limited maximum number of search results Meilisearch will return for any given query. You can use [the `maxTotalHits` index setting](/reference/api/settings.md#pagination) to configure this, but be aware that higher limits will negatively impact search performance.
 
 ## "Previous" and "Next" buttons
 
-Using "Previous" and "Next" buttons for pagination means that users can easily navigate through results, but don't have the ability to jump to an arbitrary results page.
+Using "Previous" and "Next" buttons for pagination means that users can easily navigate through results, but don't have the ability to jump to an arbitrary results page. This is Meilisearch's recommended solution when creating pagination interfaces.
```
guimachiavelli, line 17:
```suggestion
Using "Previous" and "Next" buttons for pagination means that users can easily navigate through results, but don't have the ability to jump to an arbitrary results page. This is Meilisearch's recommended solution when creating paginated interfaces.
```

reference/api/search.md
```diff
@@ -385,6 +385,8 @@ You can use `hitsPerPage` and `page` to [paginate search results](/learn/advance
 
 ::: note
 `hitsPerPage` and `page` take precedence over `offset` and `limit`. If a query contains either `hitsPerPage` or `page`, any values passed to `offset` and `limit` are ignored.
+
+`hitsPerPage` and `page` are resource-intensive options and might negatively impact search performance. This is particularly likely if [`maxTotalHits`](/reference/api/settings.md) is set to a value higher than its default.
```
curquiza, line 389: Can't we physically highlight this, in a warning (instead of notes) or with ⚠️?

There is a lot of text on this page, I would like their eyes to catch what you just added

curquiza (approved):
Thank you gui ❤️

maryamsulemani97 (approved):
🌮 

bors merge

PR #2179, opened 2023-02-20
Remove "documentation team" from contribution docs

dichotommy:
Documentation team is an internal term. "Maintainer" is a more neutral term that is common in open-source communities. Right now, members of Meilisearch's documentation team are the only consistent maintainers of the Meilisearch documentation; however, we can imagine a future where some maintainers are not official employees.


maryamsulemani97 (approved):
🍝 

bors merge

PR #2181, opened 2023-02-21
Add Korean to list of supported languages

### LINKED ISSUE #2172: Add Korean support to the Language page
It could be intended, but the [Language documentation page](https://docs.meilisearch.com/learn/what_is_meilisearch/language.html) doesn't mention that we now support Korean.

guimachiavelli:
Fixes #2172

@ManyTheFish: do I understand correctly that Korean uses the default pipeline for space-separated languages? Or do we use another pipeline?

### DIFF
--- a/learn/what_is_meilisearch/language.md
+++ b/learn/what_is_meilisearch/language.md
@@ -4,8 +4,9 @@ Meilisearch is multilingual, featuring optimized support for:
 
 - Any language that uses whitespace to separate words
 - Chinese
-- Japanese
 - Hebrew
+- Japanese
+- Korean
 - Thai
 
 We aim to provide global language support, and your feedback helps us move closer to that goal. If you notice inconsistencies in your search results or the way your documents are processed, please [open an issue in the Meilisearch repository](https://github.com/meilisearch/meilisearch/issues/new/choose).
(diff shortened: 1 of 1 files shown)


dichotommy:
Note that I think we should remove the "pipeline" language from this page in any case. It's too specific and technical for this context, and prone to go out of date. If it should be documented anywhere, it's in the tokenization page.

ManyTheFish:
Hello!
There is a new dedicated segmenter for Korean but no specialized Normalization, but I agree with @dichotommy about the fact that we shouldn't be too specific. Because we'll never support perfectly a Language, there are always improvements to do in each Language support, for instance, Chinese was supported before v1.0, however, we improved the Chinese Language support by changing the way we normalize words in v1.0.

That's been said, there are a set of Languages on which we have done some efforts to support, the "supported Language", including Korean; and the rest of the Languages that we didn't do anything to support so far.

guimachiavelli:
Thanks for your comment, @ManyTheFish! We have discussed it in the team and decided to keep this PR focussed on informing users about Korean support and remove the pipeline section in another PR (coming shortly after this one).

maryamsulemani97 (approved):
🌯

PR #2226, opened 2023-03-14
Migration Demo

justkahdri:
# Pull Request

> Note: The website that loads each file is agnostic to the filetype, it'll always treat the source as `mdx` although it can be a `.md`. In this case I didn't change the filetypes as it would make the PR less readable but in general it should be changed to `.mdx` in order to have the correct syntax highlighting on the IDE.

Changes are grouped in different commits:
#### `migration:` commits are based on the JSX syntax
- Components need always closing tags (`<br>` → `<br/>`)
- `<!---` html comments not allowed 
- Remove `<>` in no-component things
- [Remove spaces in images paths](https://github.com/meilisearch/documentation/pull/2226/commits/c3044a21c46ae767657c06710e6f37d1200ff9f5)
- JSX styles inline

#### `components:` commits use the new React custom components instead of the old Vue components
- Capsule (tip, note, warning, danger)
- Tabs _←The most notorious change_
- Columns
- Divider 

These components have no change at all in the way they are written:
- CodeSamples
- RouteHighlighter

#### `config:`
Creates the new config JSON files that separates the entries in different categories and enables the folder structure (and file naming) to be more flexible without affecting the actual site. Probably this will change in a near future, since we want to add the possibility to create new sidebar tabs as talked with @dichotommy.

> Disclaimer: The content is probably wrong distributed on each category, 
> is just a demonstration on how it could be written.

#### `content:`
The docs home content now is located in `home.mdx`. _(This can be changed easily if you prefer `homepage.mdx`)_


dichotommy (approved):
LGTM

PR #2229, opened 2023-03-20
v1.1: Update postman

### LINKED ISSUE #2196: v1.1: Update Postman
Update `.vuepress/public/postman/meilisearch-collection.json`

maryamsulemani97:
closes #2196

### DIFF
--- a/.vuepress/public/postman/meilisearch-collection.json
+++ b/.vuepress/public/postman/meilisearch-collection.json
@@ -86,7 +86,7 @@
 							}
 						},
 						"url": {
-							"raw": "{{url}}/indexes/{{indexUID}}/documents",
+							"raw": "{{url}}/indexes/{{indexUID}}/documents?csvDelimiter=;",
 							"host": [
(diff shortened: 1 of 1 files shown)


curquiza (changes requested):
Thanks @maryamsulemani97 😎 

1) I would put the multi-search route into the `Search` folder, and not into another folder.
Or at least, put the `Multi-search` folder right below the `Search` folder, not at the end.

2) Regarding `csvDelimiter`, I would recommend the follow changes:
- Don't activate `csvDelimiter` by default
- Put the non-encoded value directly, I think you can write directly `#` and no `%23`
- Rather use `;` instead of `#`, it's more common if I'm correct (let me know if I'm wrong)

Capture d’écran 2023-03-20 à 18 26 43

3) A mistake coming from me, but let's fix it 😅 
`udpate` -> `update` in the PATCH route

Capture d’écran 2023-03-20 à 18 29 19

maryamsulemani97:
bors merge

PR #2233, opened 2023-03-22
Quick start: use authorization header with Meilisearch cloud

maryamsulemani97:
Updates the quick start to mention that users need the `-H Authorization: Bearer API_KEY` header if they're using Meilisearch Cloud or launched their instance with a master key


maryamsulemani97:
bors merge

PR #2234, opened 2023-03-22
combine installation+supported os pages

### LINKED ISSUE #2148: Combine installation page w/ supported OS page
Just what the title says... as a user, I would expect to find information about which OSes Meilisearch supports on the installation page.

maryamsulemani97:
closes #2148


dichotommy:
Could you make this a draft, and we'll implement the same changes on the new docs site branch once it has reached a more stable state? Thanks

learn/getting_started/installation.md
```diff
@@ -1,8 +1,10 @@
-# Installation
+# Installation and supported operating systems
```
guimachiavelli, line 1: Is it a problem that the installation guide's sidebar link is super long now?

```suggestion
# Installation and OS support
```

This is shorter, but does it make sense?

maryamsulemani97: OS is a well-known acronym, but I think we should use the full form in the title.

maryamsulemani97:
closing for #2274

PR #2243, opened 2023-03-27
Convert `.md` files to mdx +r emove vuepress containers

maryamsulemani97:
- Converts .md files to .mdx
- Removes remaining vuepress containers and replaces them with `<capsule>`


dichotommy (changes requested):
Something weird happened to the commit tree, eh? 😂 No big deal, we can merge it anyway. Just one thing, could you `git rm` `config.js`? We don't want to add that back to the new branch 🙏🏻

PR #2247, opened 2023-03-27
move files to folders+rename files

maryamsulemani97:
- Create `learn`, `reference`, and `resources` folders
  - Move files into appropriate folders
  - Rename files to remove prefixes like `adv_`, `api_`


dichotommy (approved):
LGTM!

PR #2255, opened 2023-03-29
update `sidebarLearn.js` to remove nesting

maryamsulemani97:
update `sidebarLearn.js` to remove nesting


dichotommy (approved):
LGTM! Let's see how it goes 🚀

PR #2257, opened 2023-03-30
New site: Create assets folder

guimachiavelli:
Create an `assets` folder, with three subfolders: 
- `datasets`: sample datasets (json, csv) used in the docs
- `images`: images illustrating guides
- `brand`: images and other assets used in the site design

It's possible we don't need the brand dir, since logos, favicons, etc are almost certainly handled in the code repo. @dichotommy, can you check with basement whether the `brand` dir is necessary?


dichotommy:
I didn't ask Basement, but we can logically presume that the only files we need in this folder are those that we would link to / display in our content. So I would say the branding folder is unnecessary.

Even if we wanted to display the Meili logo on a page, we would probably ask the devs to create a component for us so that it would be well aligned and could be reused.

dichotommy (changes requested):
Remove `brand` folder

PR #2275, opened 2023-04-03
Group config files and add sidebar-footer.json

justkahdri:
DISCLAIMER: This will break staging, since it moves and renames the config files necessary for loading the sidebar and mapping the content. The fix is ready on this [PR](https://github.com/meilisearch/meilisearch.com/pull/86) but it needs these changes merged first.


config/sidebar-footer.json
```diff
@@ -0,0 +1,12 @@
+[
+    {
+      "source": "resources/faq.mdx",
+      "label": "FAQ",
+      "slug": "faq"
+    },
+    {
+      "to": "/",
+      "label": "Back to meilisearch.com",
+      "slug": "landing"
```
dichotommy, line 10: Perhaps an ignorant question: I'm wondering what this `slug` property is actually doing, since there is no route being generated (to my knowledge).

dichotommy (approved):
LGTM 🧑‍🚀 👍🏻 🚀

config/sidebar-footer.json
justkahdri: That's a good question, I answered that on Slack but forgot to mention it here. We would use the `slug` to match each item with its respective icon, and in the case it doesn’t have a `source` (like back to [meilisearch.com](http://meilisearch.com/)) the `to` field is used as the link for it. (It could be an absolute url as `https://blog.meilisearch.com/` too)

justkahdri:
@dichotommy conflicts resolved! All good for merging

PR #2297, opened 2023-04-05
Migration: add samples template file

guimachiavelli:
@bidoubiwa, do we need this file to be in a specific place?


dichotommy (approved):
I think root should be fine.

🏎️

PR #2301, opened 2023-04-06
Remove search sh

dichotommy:
This is just a simple search shell function that ChatGPT created for me, intended for local use, does not need to be stored in git remote


dichotommy:
Sorry about the tag @meilisearch/integration-team , it was an error due to codeowners file.

PR #2303, opened 2023-04-06
Fix homepage links in properties

dichotommy:
When we write links in Markdown e.g. [link content](url), they are transformed at build so that /docs is added at the beginning—this gives us the correct site route.

However, when we write links in HTML/JS—like in component properties—they will not be transformed this way. In those cases, we have to remember the `/docs`


maryamsulemani97 (approved):
🏃

PR #2312, opened 2023-04-06
[FIX]: homepage JSX syntax

justkahdri:
# Pull Request

## What does this PR do?
Fixes homepage JSX syntax on `Featured` component.

_JSX should be always wrapped in only one tag (it can be an empty tag called "fragment"). We can use `()` for grouping it, `{}` means is a JS object so it won't work_


dichotommy (approved):
LGTM! 🙏🏻

PR #2320, opened 2023-04-06
[FIX]: Style syntax on JSX

justkahdri:
## What does this PR do?
- Yet again, fixes JSX syntax erros. CSS should be an object.


dichotommy (approved):
LGTM!

PR #2323, opened 2023-04-06
Migration: fix sidebar depth

maryamsulemani97:
This PR fixed the "On this page sidebar" of some pages to include H3s and H4s so users can easily jump between sections

### DIFF
--- a/learn/advanced/faceted_search.mdx
+++ b/learn/advanced/faceted_search.mdx
@@ -1,7 +1,5 @@
 ---
-
-sidebarDepth: 2
-
+sidebarDepth: 3
 ---
 
 # Faceted search
--- a/learn/advanced/filtering.mdx
+++ b/learn/advanced/filtering.mdx
@@ -1,7 +1,5 @@
 ---
-
-sidebarDepth: 2
-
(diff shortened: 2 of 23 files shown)


dichotommy:
Hey, based on which criteria did you make the decision to extract H3s for these pages?

maryamsulemani97:
>Hey, based on which criteria did you make the decision to extract H3s for these pages?

For API references: only the routes and object (if available) headings should be visible

For the learn section: I used the same sidebar depth as the current site for most pages. For long pages with a lot of text(e.g. master key and tenant tokens), I displayed all heading levels to allow readers to see what the guide offers by looking at the sidebar. 

The only pages that don't show H4s are ones where H4s are called "Example"

dichotommy (approved):
Ok, thanks for explaining. Makes sense to basically follow the current site. I'm not a big fan of extracting H4s, and I think that if it seems necessary then that is a good sign that a page is too long/unfocused. But in any case, I am curious to see whether `sidebarDepth: 4` breaks some things. I don't think we've seen that level of indentation on the new site yet.

PR #2349, opened 2023-04-11
Change SDK page from H3s to H2s

dichotommy:
This page only had H3s, so nothing was showing up in the sidebar. At first I changed the `sidebarDepth`, but I think it makes more sense to follow best practices and switch to H2s.


maryamsulemani97 (approved):
🎨

PR #2370, opened 2023-04-17
Add warning to settings and multi search API reference

### LINKED ISSUE #2326: Add documentation about error handling in multi-search
Please add some documentation about how errors are handled in multi-search.

E.g. when one out of 5 queries in a multi-search is buggy (e.g. uses a non-existing field), that no results are returned, even if the other 4 queries are OK.

Or that only the error for the first query with error will be returned but not the others, e.g. 2 out of 5 queries have bugs, the error message will be only about the first of the two wrong queries, not about the second.

Don't get me wrong: I like the way you implemented it. This is only about improving the docs to be clear about what to expect.

maryamsulemani97:
closes #2326 

This PR adds a warning that if one or more query/setting is badly formatted, or has a typo or error, Meilisearch will thrown an error. In case of multiple errors in the  query/setting, the error only addresses the first erroneous query/setting.

### DIFF
--- a/reference/api/multi_search.mdx
+++ b/reference/api/multi_search.mdx
@@ -66,7 +66,7 @@ Each search result object is composed of the following fields:
 <CodeSamples id="multi_search_1" />
 
 <Capsule intent="warning">
-If one or more sent queries have an error, Meilisearch will not return any results.
+If one or more sent queries have an error, Meilisearch will not return any results. In case of multiple errors, the error message will only address the first erroneous query.
 </Capsule>
 
 #### Response: `200 Ok`
--- a/reference/api/multi_search.mdx
+++ b/reference/api/multi_search.mdx
@@ -65,6 +65,10 @@ Each search result object is composed of the following fields:
 
 <CodeSamples id="multi_search_1" />
 
+<Capsule intent="warning">
+If one or more sent queries have an error, Meilisearch will not return any results.
+</Capsule>
+
 #### Response: `200 Ok`
 
 ```json
--- a/reference/api/multi_search.mdx
+++ b/reference/api/multi_search.mdx
@@ -66,7 +66,7 @@ Each search result object is composed of the following fields:
 <CodeSamples id="multi_search_1" />
 
 <Capsule intent="warning">
-If one or more sent queries have an error, Meilisearch will not return any results. In case of multiple errors, the error message will only address the first erroneous query.
+If one or more sent queries are badly formatted or cause an error, Meilisearch will throw an error. In case of multiple errors in the query, the error message will only address the first erroneous query.
 </Capsule>
 
 #### Response: `200 Ok`
--- a/reference/api/multi_search.mdx
+++ b/reference/api/multi_search.mdx
@@ -66,7 +66,7 @@ Each search result object is composed of the following fields:
 <CodeSamples id="multi_search_1" />
 
 <Capsule intent="warning">
-If one or more sent queries are badly formatted or cause an error, Meilisearch will throw an error. In case of multiple errors in the query, the error message will only address the first erroneous query.
+If Meilisearch encounters an error when processing any of the queries in a multi-search request, it immediately interrupts the process and returns an error message. In case of multiple errors in the query, the error message will only address the first erroneous query.
 </Capsule>
 
 #### Response: `200 Ok`
--- a/reference/api/multi_search.mdx
+++ b/reference/api/multi_search.mdx
@@ -66,7 +66,7 @@ Each search result object is composed of the following fields:
 <CodeSamples id="multi_search_1" />
 
 <Capsule intent="warning">
-If Meilisearch encounters an error when handling any of the queries in a multi-search request, it immediately stops processing the request and returns an error message. If a request contains multiple issues, the returned message will only address the first error.
+If Meilisearch encounters an error when handling any of the queries in a multi-search request, it immediately stops processing the request and returns an error message. The returned message will only address the first error encountered.
 </Capsule>
 
 #### Response: `200 Ok`
--- a/reference/api/multi_search.mdx
+++ b/reference/api/multi_search.mdx
@@ -66,7 +66,7 @@ Each search result object is composed of the following fields:
 <CodeSamples id="multi_search_1" />
 
 <Capsule intent="warning">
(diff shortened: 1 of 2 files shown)


reference/api/settings.mdx
```diff
@@ -152,6 +152,10 @@ If the provided index does not exist, it will be created.
 
 <CodeSamples id="update_settings_1" />
 
+<Capsule intent="warning">
+If one or more settings have an error, Meilisearch will not return any results. In case of multiple errors, the error message will only address the first erroneous setting.
```
dichotommy, line 156: Where is this change coming from? It does not seem to be related to the issue.

reference/api/multi_search.mdx
```diff
@@ -65,6 +65,10 @@ Each search result object is composed of the following fields:
 
 <CodeSamples id="multi_search_1" />
 
+<Capsule intent="warning">
+If one or more sent queries have an error, Meilisearch will not return any results. In case of multiple errors, the error message will only address the first erroneous query.
```
dichotommy, line 69:
```suggestion
If one or more of the multi-search queries cause an error, Meilisearch will not return any results. In case of multiple errors, the error message found in the task object will describe only the first error encountered.
```

WDYT of this change? For me it was unclear that the error message would be found in a task object. I also think that "erroneous query" implies that the query was badly formatted, whereas an error could conceivably come from something else.

dichotommy (changes requested):

dichotommy:
My first thought on seeing this warning was that it should go in the main description of the endpoint, where it will be more visible.

However, this endpoint does not have a description. The section name "`POST` route" is also inconsistent with how we do things on other API reference pages. Could you fix these two things in a separate PR @maryamsulemani97 ?

reference/api/settings.mdx
maryamsulemani97: It's not. But you get the same behavior here, so I thought we should update this as well

dichotommy: Ah, well there are no "results" when updating settings so the first sentence should be changed or removed.

reference/api/multi_search.mdx
dichotommy: That sounds good to me Maryam!

maryamsulemani97: I removed the task object part as Meilisearch only shows the error object:
```
{"message":"Unknown field `inexUid` inside `.queries[2]`: did you mean `indexUid`? expected one of `indexUid`, `q`, `offset`, `limit`, `page`, `hitsPerPage`, `attributesToRetrieve`, `attributesToCrop`, `cropLength`, `attributesToHighlight`, `showMatchesPosition`, `filter`, `sort`, `facets`, `highlightPreTag`, `highlightPostTag`, `cropMarker`, `matchingStrategy`",
"code":"bad_request",
"type":"invalid_request",
"link":"https://docs.meilisearch.com/errors#bad_request"}

**Review by dichotommy — COMMENTED** [review]
I am still not finding these notes to be very clear.

"If one or more sent queries... cause an error, Meilisearch will throw an error." Well, yeah!

What we want users to understand is that the entire request will fail if any part of it causes an error. In other words, the update to settings and the multi search request are not partial: they either succeed completely or fail completely.

@guimachiavelli How would you reword this?

File: reference/api/multi_search.mdx [hunk 3]
```diff
@@ -65,6 +65,10 @@ Each search result object is composed of the following fields:
 
 <CodeSamples id="multi_search_1" />
 
+<Capsule intent="warning">
+If one or more sent queries are badly formatted or cause an error, Meilisearch will throw an error. In case of multiple errors in the query, the error message will only address the first erroneous query.
```
guimachiavelli, line 69:
```suggestion
If Meilisearch encounters an error when processing any of the queries in a multi-search request, it immediately interrupts the process and returns an error message. All queries contained in a multi-search request must be valid, or the whole request fails.
```

reference/api/settings.mdx
```diff
@@ -152,6 +152,10 @@ If the provided index does not exist, it will be created.
 
 <CodeSamples id="update_settings_1" />
 
+<Capsule intent="warning">
+If one or more settings are badly formatted or cause an error, Meilisearch will throw an error. In case of multiple errors in the settings, the error message will only address the first erroneous setting.
```
guimachiavelli, line 156:
```suggestion
If Meilisearch encounters an error when updating any of the settings in a request, it immediately interrupts the process and returns an error message. All settings contained in a request to the update multiple settings endpoint must be valid, or the whole request fails.
```

guimachiavelli (changes requested):
Agreed. The notes do not clearly communicate the points indicated by @dichotommy.

I made a few suggestions that I think express the atomicity of these transactions a bit more clearly. The specific wording could probably be more elegant, though, but I guess that's relatively minor.

guimachiavelli (dismissed):
LGTM

Give @dichotommy a few more days. If he doesn't reply, dismiss his review and merge.

reference/api/multi_search.mdx
```diff
@@ -65,6 +65,10 @@ Each search result object is composed of the following fields:
 
 <CodeSamples id="multi_search_1" />
 
+<Capsule intent="warning">
+If Meilisearch encounters an error when handling any of the queries in a multi-search request, it immediately stops processing the request and returns an error message. If a request contains multiple issues, the returned message will only address the first error.
```
dichotommy, line 69:
```suggestion
If Meilisearch encounters an error when handling any of the queries in a multi-search request, it immediately stops processing the request and returns an error message. The returned message will only address the first error encountered.
```

reference/api/settings.mdx
```diff
@@ -152,6 +152,10 @@ If the provided index does not exist, it will be created.
 
 <CodeSamples id="update_settings_1" />
 
+<Capsule intent="warning">
+If Meilisearch encounters an error when updating any of the settings in a request, it immediately stops processing the request and returns an error message. In this case, the database settings remain unchanged. If a request contains multiple issues, the returned message only addresses the first error.
```
dichotommy, line 156:
```suggestion
If Meilisearch encounters an error when updating any of the settings in a request, it immediately stops processing the request and returns an error message. In this case, the database settings remain unchanged. The returned error message will only address the first error encountered.
```

dichotommy (approved):
I made two small suggestions, but other than that it LGTM. Once merged, please update #2382 moving the multi-search warning into the main endpoint description. Thanks!

guimachiavelli (approved):

PR #2382, opened 2023-04-25
update multi-search API page

maryamsulemani97:
Updates the multi-search API page to:
- Rename "POST route" to "Perform a multi-search"
- Add a description for the route 

Related to: https://github.com/meilisearch/documentation/pull/2370#issuecomment-1521539855

## To do:

- [ ] Move warning added in https://github.com/meilisearch/documentation/pull/2370 to main endpoint description

## Redirects
Added in https://github.com/meilisearch/meilisearch.com/pull/122

### DIFF
--- a/reference/api/multi_search.mdx
+++ b/reference/api/multi_search.mdx
@@ -4,7 +4,7 @@ The `/multi-search` route allows you to perform multiple search queries on one o
 
 ## Search with `multi-search`
 
-Search for documents matching a specific query in one or more indexes.
+Perform multiple searches in a single request.
 
 <RouteHighlighter method="POST" route="/multi-search"/>
 
--- a/reference/api/multi_search.mdx
+++ b/reference/api/multi_search.mdx
@@ -2,7 +2,7 @@
 
 The `/multi-search` route allows you to perform multiple search queries on one or more indexes by bundling them into a single HTTP request. Multi-search is also known as federated search.
 
-## Search with `multi-search`
+## Search with `/multi-search` route
 
 Perform multiple searches in a single request.
 
--- a/reference/api/multi_search.mdx
+++ b/reference/api/multi_search.mdx
@@ -2,9 +2,9 @@
 
 The `/multi-search` route allows you to perform multiple search queries on one or more indexes by bundling them into a single HTTP request. Multi-search is also known as federated search.
 
-## Search with `/multi-search` route
+## Perform a multi-search
 
-Perform multiple searches in a single request.
+Bundle multiple search queries in a single API request.
 
 <RouteHighlighter method="POST" route="/multi-search"/>
 
--- a/reference/api/multi_search.mdx
+++ b/reference/api/multi_search.mdx
@@ -2,15 +2,19 @@
 
 The `/multi-search` route allows you to perform multiple search queries on one or more indexes by bundling them into a single HTTP request. Multi-search is also known as federated search.
 
+## Search with `multi-search`
+
+Search for documents matching a specific query in one or more indexes.
+
 <RouteHighlighter method="POST" route="/multi-search"/>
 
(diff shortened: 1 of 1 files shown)


dichotommy (changes requested):
I don't think that this (removing the main H2 and changing the header level for all the sections on the page) is an acceptable solution to the problem.

If we do this, this API reference page will be differently formatted than all other API reference pages. We will also have to undo it if ever we add a new endpoint.

Consider Health, Version, or Dumps. These API routes each have only a single endpoint, but we still follow the same structure. 

Rather than remove "POST route", it would be better to rename it something like "Perform a multi-search", and add some kind of description. Please look at other API reference pages for inspiration.

reference/api/multi_search.mdx
```diff
@@ -2,7 +2,9 @@
 
 The `/multi-search` route allows you to perform multiple search queries on one or more indexes by bundling them into a single HTTP request. Multi-search is also known as federated search.
 
-## POST route
+## Search with `multi-search`
+
+Search for documents matching a specific query in one or more indexes.
```
maryamsulemani97, line 7:
```suggestion
Perform multiple searches in a single request.
```

reference/api/multi_search.mdx
```diff
@@ -2,7 +2,9 @@
 
 The `/multi-search` route allows you to perform multiple search queries on one or more indexes by bundling them into a single HTTP request. Multi-search is also known as federated search.
 
-## POST route
+## Search with `multi-search`
```
maryamsulemani97, line 5:
```suggestion
## Search with `/multi-search` route
```

guimachiavelli (changes requested):
I think @dichotommy's request still hasn't been fully met, if I understand it correctly.

In most endpoint/route descriptions, we avoid referencing the endpoint directly and focus on what the route actually does. For example, we write "get one document", not "get a document with `/document`".

I recommend focussing on _what_ this route does as opposed to _how_ it does things.

maryamsulemani97:
>I recommend focussing on what this route does as opposed to how it does things.

"Perform multiple searches in a single request" does cover that. The only difference between this and the other search routes is you can bundle all your search requests in one query instead of sending multiple queries.

reference/api/multi_search.mdx
```diff
@@ -2,7 +2,9 @@
 
 The `/multi-search` route allows you to perform multiple search queries on one or more indexes by bundling them into a single HTTP request. Multi-search is also known as federated search.
 
-## POST route
+## Perform a multi-search
+
+Bundle multiple search queries in a single API request.
```
dichotommy, line 7:
```suggestion
Bundle multiple search queries in a single API request. Multi-search queries do not need to belong to the same index. You can search through multiple indexes at once using this endpoint.
```

WDYT @guimachiavelli ?

dichotommy (changes requested):
The warning added by https://github.com/meilisearch/documentation/pull/2370 should be moved from the "Example" section to the "Perform a multi-search" section.

reference/api/multi_search.mdx
guimachiavelli: I find it a bit weird to speak of queries _belonging_ to an index, to be honest. Perhaps we can just skip that sentence altogether?

```suggestion
Bundle multiple search queries in a single API request. Using this endpoint to search through multiple indexes at once.
```

guimachiavelli (changes requested):
I think @dichotommy's last couple of suggestions are good. Let's follow them, and then merge this PR, as it's been open for too long and we want this info online.

For anyone reading this, in case any further concerns related to the multi-search API reference pop up, please open a separate issue.

guimachiavelli (approved):
🤹

PR #2392, opened 2023-04-27
Improve site organization + visibility: Delete the "Philosophy" page

### LINKED ISSUE #2386: Improve site organization + visibility: Delete the “Philosophy” page
As part of https://github.com/meilisearch/documentation/issues/2380, we'd like to remove the [Philosophy](https://www.meilisearch.com/docs/learn/what_is_meilisearch/philosophy) page.

### Suggestion
Instead of deleting the page, we could move some of the content to the [overview page under the "Philosophy" section](https://www.meilisearch.com/docs/learn/what_is_meilisearch/overview#philosophy).

maryamsulemani97:
closes #2386 

Deletes Philosophy page and moves content to `/learn/what_is_meilisearch/overview`

Redirects added in https://github.com/meilisearch/meilisearch.com/pull/124


guimachiavelli (changes requested):
I think copypasting the Philosophy page verbatim is not a very good solution, as it's just a whole lot of content in a page that already has a sort of unclear role in the docs.

It's weird, for instance, to have "Anti-pattern" as subheading of "Philosophy". The second paragraph of "Front-facing search" is also extremely obscure, but that might be because I'm lazy and dumb.


In my opinion, we should condense the whole Philosophy section into a couple of paragraphs describing our current approach to product design and development. 

I guess instead of moving content, I think we need to first establish what is the goal of this section, then achieve it with the fewest words possible.

guimachiavelli (changes requested):
I thinks this is a step in the right direction, but still needs more work.

The focus is still on what users can do ("you can configure search parameters", "you can delete ranking rules"), instead of guiding principles. What is behind the fact users can configure search parameters and ranking rules? Why does Meilisearch focusses on this?

dichotommy:
Hey @maryamsulemani97 , I think it's important to explain and justify your decisions when it comes to keeping/ignoring information from a deleted page.

For this, I always recommend that you start by identifying the key pieces of information and make an argument for why they are important. Usually key pieces of information are things that are not present in the rest of the site, so you want to preserve them. However they could also be important for external reasons related to Meilisearch's business plan.

Here is my analysis of the philosophy page's key information:

- "For end-users, Meilisearch aims to provide an intuitive search-as-you-type experience, with a response time under 50 milliseconds."
  - This is important because we want to emphasize that Meilisearch targets end users (not log search or other use cases).
  - This is also important because understanding that Meilisearch is designed for search-as-you-type, and for that purpose must return results in under 50ms, **helps users understand our design decisions.** There are cases where Meilisearch sacrifices exactness or thoroughness of search for speed, and users will be better able to accept this if they have a correct understanding of the engine's philosophy/goals.

- "Meilisearch is fully committed to the philosophy of prefix search."
  - Like the last one, this is important for users because it explains why our engine is designed the way it is.
  - Meilisearch does not try to match queries exactly—it tries to find the best matches for what the user might be trying to type based on what they have already typed.
  - This explains some built-in behavior that cannot be turned off. It gives users a better sense of what Meilisearch is built for and whether it's the right product for them. (The Wikipedia link is not the best, though)
- "Meilisearch should not be your main data store."
  - **Meilisearch is a search engine, not a database.** It will never compete with databases when it comes to the speed of adding, removing, or editing documents.
  - Meilisearch works best if you use an external database for that stuff, and only reindex your data in Meilisearch when necessary.
- "Meilisearch queries should be sent directly from the front-end. The more proxies there are between Meilisearch and the end-user, the slower the queries and search experience will be."
  - This is an important tip. However, it stands out from the others I've highlighted because it says nothing about Meilisearch's use-case and does not seem helpful for comparing Meilisearch with other tools.
  - I would argue that this does not belong in a section titled philosophy. The information should be preserved, but moved elsewhere.

I would recommend restarting your work with this in mind, and in the future starting with this step. That way you can verify your assumptions about what is important to keep and what you can get rid of before even opening a PR.

Hope this helps!

learn/what_is_meilisearch/overview.mdx
```diff
@@ -32,9 +32,15 @@ _Meilisearch helps the Rust community find crates on [crates.meilisearch.com](ht
 
 ## Philosophy
 
-Meilisearch is committed to providing **open source**, **easy to use**, **customizable** instant search that **focuses on the needs of end-users**.
+We always aim for a simple and intuitive experience for both developers and end-users. Ease of use was our primary goal starting out, and continues to drive the development of Meilisearch today. Meilisearch requires very little configuration to get up and running. It works out-of-the-box with default settings that meet the needs of most projects. However, searching is still highly customizable.
```
guimachiavelli, line 35:
```suggestion
Our goal is to provide a simple and intuitive experience for both developers and end-users. Ease of use was the primary focus of Meilisearch from its first release, and it continues to drive its development today.
```

learn/what_is_meilisearch/overview.mdx
```diff
@@ -32,9 +32,15 @@ _Meilisearch helps the Rust community find crates on [crates.meilisearch.com](ht
...
 
-Read more about [the philosophy of our search engine](/learn/what_is_meilisearch/philosophy).
+The returned results are **sorted according to a set of consecutive rules called [ranking rules](/learn/core_concepts/relevancy#ranking-rules)**. Our default ranking rules work for most use cases as we developed them by working directly with our users. You can delete existing rules, add new ones, or even change the order in which they are executed.
+
+You can also **configure the [search parameters](/reference/api/search)** to refine your search even further. We support [filters](/learn/advanced/filtering) and [faceting](/learn/advanced/faceted_search#faceted-search).
```
guimachiavelli, line 39:
```suggestion
Meilisearch's ease-of-use goes hand-in-hand with ultra relevant search results. Meilisearch **sorts results according to a set of [ranking rules](/learn/core_concepts/relevancy#ranking-rules)**. Our default ranking rules work for most use cases as we developed them by working directly with our users. You can also **configure the [search parameters](/reference/api/search)** to refine your search even further.
```

learn/what_is_meilisearch/overview.mdx
```diff
@@ -32,9 +32,15 @@ _Meilisearch helps the Rust community find crates on [crates.meilisearch.com](ht
...
+
+Meilisearch should **not be your main data store**. It is a search engine, not a database. Meilisearch should contain only the data you want your users to search through. If you must add data that is irrelevant to search, be sure to [make those fields non-searchable](/learn/configuration/displayed_searchable_attributes#searchable-fields) to improve relevancy and response time.
+
+For end-users, Meilisearch aims to provide an intuitive search-as-you-type experience, with a response time under 50 milliseconds. This helps users spend less time tinkering with search queries, and more time looking at results. Meilisearch is designed for front-facing search, search for users of your site, app, or product. As a result, we are fully committed to the philosophy of [prefix search](/learn/advanced/prefix).
```
guimachiavelli, line 43:
```suggestion
Meilisearch provides an intuitive search-as-you-type experience with response times under 50 milliseconds, no matter whether you are developing a site or an app. This helps end-users find what they are looking for quickly and efficiently. To make that happen, we are fully committed to the philosophy of [prefix search](/learn/advanced/prefix).
```

guimachiavelli (changes requested):
I think the latest changes are pretty good. So good, in fact, they highlight some of text we are moving is not great: sentences don't flow neatly into one another, some of them repeat information already given, etc. You could argue it's out of scope for the PR, but I couldn't resist tidying it up a bit.

guimachiavelli (approved):

PR #2405, opened 2023-05-09
Clarify storage across indexes

### LINKED ISSUE #878: Clarify storage across indexes
Would be great if the [docs](https://docs.meilisearch.com/reference/under_the_hood/storage.html) said something about whether identical documents or parts of the documents are shared across indexes.
E.g. if I add the exact same document A to index A as well as B, will it use twice the amount of RAM? Or is it going to be shared? Does it depend on whether the documents are 100% identical or are the fields shared?

That would be great, thanks :) 

By the way, happy to provide a PR if someone is able to reply to the questions but I think when answering here, one might as well create a PR right away ;) 

--- _Edit by @guimachiavelli (2023-05-04)_ ---

## Tasks
Update the following pages with a short description of Meilisearch's behaviour when the same document is added to multiple indexes in a single instance.

- [ ] Update `/learn/core_concepts/documents.mdx`
- [ ] Update `/learn/advanced/storage.mdx`

maryamsulemani97:
closes #878 

I didn't update the storage page as it was mostly on LMDB

### DIFF
--- a/learn/core_concepts/documents.mdx
+++ b/learn/core_concepts/documents.mdx
@@ -4,7 +4,9 @@ sidebarDepth: 3
 
 # Documents
 
-A document is an object composed of one or more fields. Each field consists of an **attribute** and its associated **value**. Documents function as containers for organizing data, and are the basic building blocks of a Meilisearch database. To search for a document, you must first add it to an [index](/learn/core_concepts/indexes).
+A document is an object composed of one or more fields. Each field consists of an **attribute** and its associated **value**. Documents function as containers for organizing data and are the basic building blocks of a Meilisearch database. To search for a document, you must first add it to an [index](/learn/core_concepts/indexes).
+
+If two indexes contain the exact same document, they will be treated as different documents and nothing will be shared between the indexes. Both documents might have different sizes depending on the [index's settings](/reference/api/settings). For example, documents in an index with `filterableAttributes` will occupy more space.
 
 ## Structure
 
--- a/learn/core_concepts/documents.mdx
+++ b/learn/core_concepts/documents.mdx
@@ -6,7 +6,7 @@ sidebarDepth: 3
 
 A document is an object composed of one or more fields. Each field consists of an **attribute** and its associated **value**. Documents function as containers for organizing data and are the basic building blocks of a Meilisearch database. To search for a document, you must first add it to an [index](/learn/core_concepts/indexes).
 
-Nothing will be shared between two indexes if they contain the exact same document. Instead, both documents will be treated as different documents. Depending on the [index's settings](/reference/api/settings), the documents might have different sizes. For example, documents in an index with `filterableAttributes` will occupy more space.
+Nothing will be shared between two indexes if they contain the exact same document. Instead, both documents will be treated as different documents. Depending on the [index's settings](/reference/api/settings), the documents might have different sizes.
 
 ## Structure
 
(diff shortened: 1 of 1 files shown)


learn/core_concepts/documents.mdx
```diff
@@ -4,7 +4,9 @@ sidebarDepth: 3
 
 # Documents
 
-A document is an object composed of one or more fields. Each field consists of an **attribute** and its associated **value**. Documents function as containers for organizing data, and are the basic building blocks of a Meilisearch database. To search for a document, you must first add it to an [index](/learn/core_concepts/indexes).
+A document is an object composed of one or more fields. Each field consists of an **attribute** and its associated **value**. Documents function as containers for organizing data and are the basic building blocks of a Meilisearch database. To search for a document, you must first add it to an [index](/learn/core_concepts/indexes).
+
+If two indexes contain the exact same document, they will be treated as different documents and nothing will be shared between the indexes. Both documents might have different sizes depending on the [index's settings](/reference/api/settings). For example, documents in an index with `filterableAttributes` will occupy more space.
```
guimachiavelli, line 9: I'm wondering if it's clear the `they` in "they will be treated as different documents" is correct. I know it's referring to the document copies, but technically the only plural thing that comes before it is indexes. Perhaps edit this slightly to avoid misunderstandings and make it 100% grammatical?

guimachiavelli (changes requested):

learn/core_concepts/documents.mdx
maryamsulemani97: But it does make sense when you read the whole sentence: "they will be treated as different documents and nothing will be shared between the indexes"
I'll try to rephrase it to something clearer

maryamsulemani97, line 9:
```suggestion
Nothing will be shared between two indexes if they contain the exact same document. Instead, both documents will be treated as different documents. Depending on the [index's settings](/reference/api/settings), the documents might have different sizes. For example, documents in an index with `filterableAttributes` will occupy more space.
```
I don't like this(too many documents), but it's clearer.

learn/core_concepts/documents.mdx
```diff
@@ -4,7 +4,9 @@ sidebarDepth: 3
 
 # Documents
 
-A document is an object composed of one or more fields. Each field consists of an **attribute** and its associated **value**. Documents function as containers for organizing data, and are the basic building blocks of a Meilisearch database. To search for a document, you must first add it to an [index](/learn/core_concepts/indexes).
+A document is an object composed of one or more fields. Each field consists of an **attribute** and its associated **value**. Documents function as containers for organizing data and are the basic building blocks of a Meilisearch database. To search for a document, you must first add it to an [index](/learn/core_concepts/indexes).
+
+Nothing will be shared between two indexes if they contain the exact same document. Instead, both documents will be treated as different documents. Depending on the [index's settings](/reference/api/settings), the documents might have different sizes. For example, documents in an index with `filterableAttributes` will occupy more space.
```
maryamsulemani97, line 9: Too many documents, but it's clearer

guimachiavelli, line 9: I suggest removing the last sentence. The explanation works fine without it, in my opinion.

```suggestion
Nothing will be shared between two indexes if they contain the exact same document. Instead, both documents will be treated as different documents. Depending on the [index's settings](/reference/api/settings), the documents might have different sizes.
```

guimachiavelli (dismissed):
One minor suggestion and we're good to go.

Issue #1591, opened 2022-04-13
Change `document id` to `document ID`


dichotommy:
`ID` might not be grammatically correct in this situation, but it is more readable and easier to transform into plural/title-case.

Note for later: `reference/api/documents.md` line 235 `id's` should be `IDs`

maryamsulemani97:
We use `id` in the [primary key guide](https://www.meilisearch.com/docs/learn/core_concepts/primary_key#document-id).

maryamsulemani97:
fixed in https://github.com/meilisearch/documentation/pull/2421

PR #2421, opened 2023-05-11
v1.2: Delete and get documents by filter

### LINKED ISSUE #2411: v1.2: Delete and get documents by filter
With v1.2, users will be able to get and delete documents by filter.

## To do
- [ ] Update [documents API page](https://www.meilisearch.com/docs/reference/api/documents) to add the new routes:
   - [ ] POST `indexes/:index_uid/documents/fetch`
   - [ ] POST `indexes/:index_uid/documents/delete` 
   - [ ] Update links to use the new endpoints

## References
- Issue on engine repo: https://github.com/meilisearch/meilisearch/issues/3477
- Specs:
   - Spec on get documents by filter: https://github.com/meilisearch/specifications/pull/234
   - Spec on delete documents by filter: https://github.com/meilisearch/specifications/pull/236
 - SME: @dureuill, @irevoire  @macraig

maryamsulemani97:
closes #2411 and #1591

## For reviewers
To avoid confusion over which route to use GET `/indexes/{index_uid}/documents` vs POST `/indexes/{index_uid}/documents/fetch` and delete by batch vs delete by filter, I added a warning for GET `/indexes/{index_uid}/documents` and delete by batch saying they will be deprecated in the near future.


reference/api/documents.mdx
```diff
@@ -4,31 +4,95 @@ The `/documents` route allows you to create, manage, and delete documents.
...
-| **`offset`**    | `0`           | Number of documents to skip                                   |
-| **`limit`**     | `20`          | Number of documents to return                                 |
-| **`fields`**    | `*`           | Document attributes to show (case-sensitive, comma-separated) |
+| Query Parameter | Default Value | Description                                                                                              |
+| :-------------- | :------------ | :------------------------------------------------------------------------------------------------------- |
+| **`offset`**    | `0`           | Number of documents to skip                                                                              |
+| **`limit`**     | `20`          | Number of documents to return                                                                            |
+| **`fields`**    | `*`           | Document attributes to show (case-sensitive, comma-separated)                                            |
+| **`filter`**    | N/A           | An array of array of strings for refining results based on attributes in the `filterableAttributes` list |
```
maryamsulemani97, line 32:
```suggestion
| **`offset`**    | `0`           | Number of documents to skip                                                                              |
| **`limit`**     | `20`          | Number of documents to return                                                                            |
| **`filter`**    | N/A           | An array of array of strings for refining results based on attributes in the `filterableAttributes` list |
```

.code-samples.meilisearch.yaml
```diff
@@ -1065,3 +1065,21 @@ filtering_update_settings_1: |-
       "genres",
       "director"
     ]'
+fetch_documents_1: |-
```
bidoubiwa, line 1068: Following the same logic as the naming for get and post search, I think it is more understandable if you name this sample `get_documents_post_1`. `fetch_documents` is not very intuitive in what it does.

reference/api/documents.mdx
```diff
@@ -301,21 +384,67 @@ Delete one document based on its unique id.
 
 You can use this `taskUid` to get more details on [the status of the task](/reference/api/tasks#get-one-task).
 
+## Delete documents by filter
```
bidoubiwa, line 387: Why not just call it `Delete documents` ? It does not exist yet and I suppose we want to promote that route as the default route to delete documents

.code-samples.meilisearch.yaml
```diff
@@ -1065,3 +1065,21 @@ filtering_update_settings_1: |-
...
+  curl \
+    -X POST http://localhost:7700/indexes/books/documents/fetch \
+    -H 'Content-Type: application/json' \
+    --data-binary '{ 
+      "filter": "(rating > 3 AND (genres = Adventure OR genres = Fiction)) AND language = English",
+      "fields": ["title", "genres", "rating", "language"],
+      "limit": 3
+    }'
+delete_documents_by_filter_1: |-
```
bidoubiwa, line 1077:
```suggestion
delete_documents_1: |-
```

See comment below

sample-template.yaml
```diff
@@ -162,3 +162,5 @@ faceted_search_update_settings_1: |-
 search_parameter_guide_facet_stats_1: |-
 faceted_search_1: |-
 faceted_search_2: |-
+fetch_documents_1: |-
+delete_documents_by_filter_1: |-
```
bidoubiwa, line 166: should be updated if you agree with my other comments

reference/api/documents.mdx
```diff
@@ -4,31 +4,105 @@ The `/documents` route allows you to create, manage, and delete documents.
 
 [Learn more about documents.](/learn/core_concepts/documents)
 
-## Get documents
+## Get documents with POST
```
bidoubiwa, line 7: Following the same logic as with the search references, to not confuse the user we might just call it `Get Documents` and not add POST.

reference/api/documents.mdx
maryamsulemani97: We have an endpoint called "Delete all documents". It might be confusing for the users if we have "Delete documents" and "Delete all documents". "Delete documents by filter" is more specific so sounded like a better option

.code-samples.meilisearch.yaml
maryamsulemani97: Thanks for pointing that out! I'll update it!

reference/api/documents.mdx
```diff
@@ -301,21 +384,67 @@ Delete one document based on its unique id.
...
+
+### Path parameters
+
+| Name              | Type   | Description                                                            |
+| :---------------- | :----- | :--------------------------------------------------------------------- |
+| **`index_uid`** * | String | [`uid`](/learn/core_concepts/indexes#index-uid) of the requested index |
+
+### Body
+
+A filter expression written as a string or array of array of strings for the documents to be deleted. `filter` will not work without first explicitly adding attributes to the [`filterableAttributes` list](/reference/api/settings#update-filterable-attributes). [Learn more about filters in our dedicated guide.](/learn/advanced/filtering#filter-basics)
```
irevoire, line 401: it would be nice if we could put a big emphasis on the fact that **you absolutely need** to make your fields filterable.
Like a warning panel or something.
One user already missed the info, and @brunoocasali was also afraid of it, so I guess it could be good to ensure no one misses it.

reference/api/documents.mdx
maryamsulemani97: There are two endpoints: 
- Get documents with GET
- Get documents with POST

Since both allow users to do the same thing(almost): get all documents or get some documents based on a filter, we decided to go with the same heading names: Get documents with GET/POST.

reference/api/search.mdx
```diff
@@ -20,7 +20,7 @@ Search for documents matching a specific query in the given index.
 This is the preferred endpoint to perform search when an API key is required, as it allows for [preflight requests](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request) to be cached. Caching preflight requests **considerably improves search speed**.
 
 <Capsule intent="note">
-By default, [this endpoint returns a maximum of 1000 results](/learn/advanced/known_limitations#maximum-number-of-results-per-search). If you want to scrape your database, use the [get documents endpoint](/reference/api/documents#get-documents) instead.
+By default, [this endpoint returns a maximum of 1000 results](/learn/advanced/known_limitations#maximum-number-of-results-per-search). If you want to scrape your database, use the [get documents endpoint](/reference/api/documents#get-documents-with-post) instead.
```
irevoire, line 23: Oh, I didn't know this was documented, but it's wrong!
Since the get/fetch documents routes are exposed for the administrator of the instance, there is no limitation on it.
You can query as many documents as you want.

reference/api/search.mdx
```diff
@@ -136,7 +136,7 @@ This endpoint only accepts [string filter expressions](/learn/advanced/filtering
 This endpoint should only be used when no API key is required. If an API key is required, use the [POST](/reference/api/search#search-in-an-index-with-post-route) route instead.
 
 <Capsule intent="note">
-By default, [this endpoint returns a maximum of 1000 results](/learn/advanced/known_limitations#maximum-number-of-results-per-search). If you want to scrape your database, use the [get documents endpoint](/reference/api/documents#get-documents) instead.
+By default, [this endpoint returns a maximum of 1000 results](/learn/advanced/known_limitations#maximum-number-of-results-per-search). If you want to scrape your database, use the [get documents endpoint](/reference/api/documents#get-documents-with-post) instead.
```
irevoire, line 139: same

irevoire (changes requested):

reference/api/documents.mdx
bidoubiwa: For me `Delete Documents` and `Delete all documents` are clearly distinguishable from each other. But it's your call :)

reference/api/documents.mdx
bidoubiwa: In the search references, there is `search` and `search with GET`.

Since we want to promote Get Documents with POST as the default value, adding `POST` may be confusing for a user that does not understand HTTP methods.

reference/api/documents.mdx
maryamsulemani97: We could put it in yellow warning boxes 🤔

reference/api/documents.mdx
maryamsulemani97: >In the search references, there is search and search with GET.

It's search with GET and search with POST, same as here.

>Since we want to promote get Documents with POST as the default value, adding POST may be confusing for a user that does not understand HTTP methods.

About promoting POST: I added it to the top and added a warning under the GET route, saying it will be deprecated in the future so use the POST rooute

Not knowing about HTTP verbs: I'm assuming users would have some idea. If they don't, we have the warning that says we recommend using POST instead of GET

reference/api/documents.mdx
maryamsulemani97: Sorry, I forgot to add this before. We have four delete documents routes:
- Delete all documents
- Delete one document
- Delete documents by filter
- Delete documents by batch

If we didn't have the delete by batch route, "Delete documents" would have worked.

reference/api/documents.mdx
bidoubiwa: My bad, I did not see the search with POST one!

reference/api/search.mdx
irevoire: Yes, it applies to the search routes

reference/api/search.mdx
guimachiavelli, line 139: Do we need to set up redirects for `/api/documents#get-documents`?

learn/core_concepts/documents.mdx
```diff
@@ -6,6 +6,10 @@ sidebarDepth: 3
 
 A document is an object composed of one or more fields. Each field consists of an **attribute** and its associated **value**. Documents function as containers for organizing data, and are the basic building blocks of a Meilisearch database. To search for a document, you must first add it to an [index](/learn/core_concepts/indexes).
 
+<Capsule intent="note">
+Meilisearch orders documents in the order they were inserted into the database.
+</Capsule>
+
```
guimachiavelli, line 12: This feels a bit random. This page defines a document and gives pointers on its constituent parts and etc. Why mention this here? I'm not convinced this is an improvement over leaving it in the API reference and am afraid some might confuse this ordering with relevancy ranking, or even sorting.

How can people use this information? As in, when is it relevant for a user to know the internal document order? I think answering these questions is very important when deciding the best location for this tidbit of data.

reference/api/documents.mdx
```diff
@@ -4,31 +4,113 @@ The `/documents` route allows you to create, manage, and delete documents.
...
 [Learn more about documents.](/learn/core_concepts/documents)
 
-## Get documents
+## Get documents with POST
+
+<RouteHighlighter method="POST" route="/indexes/{index_uid}/documents/fetch"/>
+
+Get a set of documents.
+
+Using `offset` and `limit`, you can browse through all your documents.
```
guimachiavelli, line 13: Perhaps we can leave "through all your", but imperative is better here.

```suggestion
Use `offset` and `limit` to browse through documents.
```

reference/api/documents.mdx
```diff
@@ -4,31 +4,113 @@ The `/documents` route allows you to create, manage, and delete documents.
...
+  "offset":0,
+  "limit":3,
+  "total":5
+}
+```
+
+## Get documents with GET
+
+<Capsule intent="danger">
+We recommend using POST `/indexes/{index_uid}/documents/fetch` as this endpoint will be deprecated in the near future.
```
guimachiavelli, line 87: Let's reword this so we don't use "we".

reference/api/documents.mdx
```diff
@@ -301,29 +392,79 @@ Delete one document based on its unique id.
 
 You can use this `taskUid` to get more details on [the status of the task](/reference/api/tasks#get-one-task).
 
+## Delete documents by filter
+
+<RouteHighlighter method="POST" route="/indexes/{index_uid}/documents/delete"/>
+
+Delete a selection of documents based on a filter.
```
guimachiavelli, line 399: Not a fan of "selection of documents". I'm struggling to justify it, but it feels off here.

```suggestion
Delete a set of documents based on a filter.
```

reference/api/documents.mdx
```diff
@@ -301,29 +392,79 @@ Delete one document based on its unique id.
...
+
+### Path parameters
+
+| Name              | Type   | Description                                                            |
+| :---------------- | :----- | :--------------------------------------------------------------------- |
+| **`index_uid`** * | String | [`uid`](/learn/core_concepts/indexes#index-uid) of the requested index |
+
+### Body
+
+A filter expression written as a string or array of array of strings for the documents to be deleted.
```
guimachiavelli, line 409: "string or array of array of strings"

is there a repeated "array of" here?

reference/api/documents.mdx
```diff
@@ -301,29 +392,79 @@ Delete one document based on its unique id.
...
+    "enqueuedAt": "2023-05-15T08:38:48.024551Z"
+}
+```
+
+You can use this `taskUid` to get more details on [the status of the task](/reference/api/tasks#get-one-task).
+
 ## Delete documents by batch
 
+<Capsule intent="danger">
+We recommend using POST `/indexes/{index_uid}/documents/delete` as this endpoint will be deprecated in the near future.
```
guimachiavelli, line 444: Same as the other comment, rewrite without first person.

reference/api/documents.mdx
```diff
@@ -301,29 +392,79 @@ Delete one document based on its unique id.
...
+</Capsule>
+
 <RouteHighlighter method="POST" route="/indexes/{index_uid}/documents/delete-batch"/>
 
-Delete a selection of documents based on an array of document id's.
+Delete a selection of documents based on an array of document ids.
```
guimachiavelli, line 449: Only accept this if you accept the other one too.

```suggestion
Delete a set of documents based on an array of document ids.
```

guimachiavelli (changes requested):

reference/api/documents.mdx
maryamsulemani97: >is there a repeated "array of" here?

Yes. It's `string` or `array of array of strings`

reference/api/search.mdx
maryamsulemani97: Good catch. I always forget the redirects 😞

reference/api/documents.mdx
maryamsulemani97: Is there a reason for not using it? We use it in [warning boxes](https://www.meilisearch.com/docs/reference/api/settings#:~:text=the%20returned%20documents.-,WARNING,-Updating%20distinct%20attributes) in the settings API reference

learn/core_concepts/documents.mdx
maryamsulemani97: I think it fits under the "add or update documents" and "add or replace document" endpoints

reference/api/documents.mdx
guimachiavelli: I thought we had agreed to slowly reduce usage of first person (singular and plural) across our docs. Other companies, like Microsoft, advise [avoiding whenever possible](https://learn.microsoft.com/en-us/style-guide/grammar/person#avoid-plural-first-person-we-us).

It's also good practice, as rephrasing "we recommend [doing X]" often results in the text being more to the point:


> We recommend that you change your password
vs
> Change your password 

> We recommend that you use the default schedule setting to keep your software up to date.
vs
> Keep your software up to date with the default schedule setting.

> We recommend using POST `/indexes/{index_uid}/documents/fetch` as this endpoint will be deprecated in the near future.
vs
> This endpoint will be deprecated in the near future. Consider using `/indexes/{index_uid}/documents/fetch` instead.
or even
> Meilisearch will deprecate this endpoint in the near future. Use `/indexes/{index_uid}/documents/fetch` instead.

reference/api/documents.mdx
```diff
@@ -47,25 +129,34 @@ Documents are ordered by Meilisearch depending on the hash of their id.
...
+      "release_date":1215475200
     }
   ],
-  "offset": 0,
-  "limit": 2,
-  "total": 500134
+  "offset":0,
+  "limit":2,
+  "total":5403
 }
```
guimachiavelli, line 160: Why are we removing the space between `:` and value? Have we decided on this recently?

Taking a quick look at other JSON blocks in other pages (geosearch, filtering, distinct attribute), it seems we  always use that space.

reference/api/documents.mdx
```diff
@@ -180,6 +275,10 @@ If you send an already existing document (same [document id](/learn/core_concept
 
 To completely overwrite a document, check out the [add or replace documents route](/reference/api/documents#add-or-replace-documents).
 
+<Capsule intent="note">
+Meilisearch orders documents in the order they were inserted into the database.
```
guimachiavelli, line 279: Perhaps add a short sentence describing why this is important?

guimachiavelli (changes requested):
Two minor additional comments on top of the open discussions.

More importantly, there have been changes to the spec concerning error messages when using the new endpoints. @maryamsulemani97, can you update this PR to reflect those changes?

[Issue 3774 on the engine repo](https://github.com/meilisearch/meilisearch/issues/3774) seems to summarise everything.

reference/api/documents.mdx
maryamsulemani97: Good catch! I'll fix it

reference/api/documents.mdx
maryamsulemani97: @irevoire 
Do you have anything on this?

irevoire: Huum, no, it's not important, I just noticed a sentence was wrong in the spec and fixed it, but actually, it doesn't really have anything to do with the initial PR :sweat: 
I don't think we should document it

reference/errors/error_codes.mdx
```diff
@@ -142,6 +142,10 @@ The provided [document identifier](/learn/core_concepts/primary_key#document-id)
 
 The [`fields`](/reference/api/documents#query-parameters) parameter is invalid. It should be a string.
 
+## `invalid_document_filter`
+
+The [`filter`](/reference/api/documents#query-parameters) parameter is invalid. It should be a string, array of strings, or array of array of strings.
```
irevoire, line 147: This error code can also be returned if the filter can't be parsed.
Also in the case the get route was used it can only be a strings :grimacing: 
tbh I don't really know how to document it, maybe we should just drop a link to this page; https://www.meilisearch.com/docs/learn/advanced/filtering#configuring-filters

irevoire (changes requested):
Should we document these two new usage of the filters at the beginning of this page: https://www.meilisearch.com/docs/learn/advanced/filtering

> Filters have several use-cases, such as refining search results and creating faceted search interfaces. Faceted search interfaces are particularly efficient in helping users navigate a great number of results across many broad categories.

reference/api/documents.mdx
maryamsulemani97: I'll remove it then. Thanks!

reference/errors/error_codes.mdx
maryamsulemani97: I'll add more details! 

I didn't get this:
>Also in the case the get route was used it can only be a strings  

Do you mean we can only use strings with `filter` for GET `/indexes/{index_uid}/document`?
The following query worked for me:
```
 curl \
    -X GET 'http://localhost:7700/indexes/books/documents?limit=2&filter=id=2'
```
`id` is an integer for this dataset

irevoire: Hey, yes, almost :grin: 
I'm saying that the whole filter can only be a string.
In Meilisearch the filters can be either a string, an array of string,s or an array of array of strings.
But in the case of the get request, it can only be a string.
Our array syntax (with the comma) that we use for the other route like `/tasks?ids=1,2,3` doesn't works for these filter, i.e.: `../documents?filter=id=2,id=3` is invalid. You must write `../documents?filter=id=2 OR id=3` (and you must encode the space correctly).

maryamsulemani97: >Our array syntax (with the comma) that we use for the other route like /tasks?ids=1,2,3 doesn't works for these filter, i.e.: ../documents?filter=id=2,id=3 is invalid. You must write ../documents?filter=id=2 OR id=3 (and you must encode the space correctly).

Thanks for clarifying!

maryamsulemani97: - Updated the error code to add more details: lines 145-154 in `reference/errors/error_codes.mdx` 
- Updated the GET endpoint to add more details on `filter`:  line 94 in `reference/api/documents.mdx`

reference/api/settings.mdx
```diff
@@ -198,7 +198,7 @@ You can use this `taskUid` to get more details on [the status of the task](/refe
 
 ## Displayed attributes
 
-The attributes added to the `displayedAttributes` list appear in search results. `displayedAttributes` only affects the search endpoints. It has no impact on the [GET documents endpoint](/reference/api/documents#get-documents).
+The attributes added to the `displayedAttributes` list appear in search results. `displayedAttributes` only affects the search endpoints. It has no impact on the [get documents with POST](/reference/api/documents#get-documents-with-post) and [get documents with GET](/reference/api/documents#get-documents-with-get) endpoints.
```
macraig, line 201: @maryamsulemani97 same question as above. Should we just replace the endpoint link to favor its usage or do we want to specify both endpoints?

learn/inner_workings/known_limitations.mdx
```diff
@@ -105,7 +105,7 @@ user = 1 OR user = 2 […] OR user = 1500 OR user = 1501 […] OR user = 2000 OR
 
 **Limitation:** By default, Meilisearch returns up to 1000 documents per search.
 
-**Explanation:** Meilisearch limits the maximum amount of returned search results to protect your database from malicious scraping. You may change this by using the `maxTotalHits` property of the [pagination index settings](/reference/api/settings#pagination-object). `maxTotalHits` only applies to the [search route](/reference/api/search) and has no effect on the [get documents endpoint](/reference/api/documents#get-documents).
+**Explanation:** Meilisearch limits the maximum amount of returned search results to protect your database from malicious scraping. You may change this by using the `maxTotalHits` property of the [pagination index settings](/reference/api/settings#pagination-object). `maxTotalHits` only applies to the [search route](/reference/api/search) and has no effect on the [get documents with POST](/reference/api/documents#get-documents-with-post) and [get documents with GET](/reference/api/documents#get-documents-with-get) endpoints.
```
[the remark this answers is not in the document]
maryamsulemani97: Since we're explaining a limitation, it would be better to mention both endpoints. If we specifically mention the POST route, it would mean the limitation doesn't apply to GET, which isn't true.

I only used the POST route when we're telling users how to retrieve documents("use the get documents endpoint"), as we want them to use POST instead of GET. E.g. the notes in `search.mdx`.

reference/api/settings.mdx
maryamsulemani97: Same as above. Omitting the GET route might imply `displayedAttributes` impacts it.

.code-samples.meilisearch.yaml
```diff
@@ -1065,3 +1065,19 @@ filtering_update_settings_1: |-
       "genres",
       "director"
     ]'
+get_documents_post_1: |-
+  curl \
+    -X POST http://localhost:7700/indexes/books/documents/fetch \
+    -H 'Content-Type: application/json' \
+    --data-binary '{ 
+      "filter": "(rating > 3 AND (genres = Adventure OR genres = Fiction)) AND language = English",
```
bidoubiwa, line 1073: In the known facet values of the movies dataset, `Fiction` does not exist. Did you mean `Science Fiction` ?

bidoubiwa (changes requested):

.code-samples.meilisearch.yaml
maryamsulemani97: I should have used the same index for one page. I'll keep that in mind next time 🙏

learn/inner_workings/known_limitations.mdx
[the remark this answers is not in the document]
macraig: Got it, thanks!

macraig (approved):
🚀

.code-samples.meilisearch.yaml
```diff
@@ -1065,3 +1065,19 @@ filtering_update_settings_1: |-
...
+      "fields": ["title", "genres", "rating", "language"],
+      "limit": 3
+    }'
+delete_documents_by_filter_1: |-
+  curl \
+    -X POST http://localhost:7700/indexes/movies/documents/delete \
+    -H 'Content-Type: application/json' \
+    --data-binary '{
+      "filter": "genres = action AND genres = adventure"
```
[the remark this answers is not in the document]
brunoocasali: Yeah, I'm aware of that 😅, but it triggered me after I saw the conditional. If is not too much trouble I think the change is for the best!

irevoire (approved):

bidoubiwa (approved):

guimachiavelli (approved):

PR #2425, opened 2023-05-15
v1.2: New filter operators

### LINKED ISSUE #2415: v1.2: Add new `IS NULL` and `IS EMPTY` filter operators
v1.2 introduces two new filter operators, `IS NULL` and `IS EMPTY`.

`IS NULL` matches fields that exist but have an explicit `null` value:

```json
{
  "name": null
}
```

`IS EMPTY` matches fields that exist but have an empty value, including strings, objects, and arrays:

```json
{
  "name": "",
  "reviews": [],
  "locations": {}
}
```

Both operators can be used together with `NOT`: `IS NOT EMPTY`, `NOT [attribute] IS EMPTY`, `IS NOT NULL`, `NOT [attribute] IS NULL`.

## Tasks

- [ ] Update `/learn/advanced/filtering.mdx`

## References
- spec: https://github.com/meilisearch/specifications/pull/232
- issues: https://github.com/meilisearch/meilisearch/issues/3615, https://github.com/meilisearch/meilisearch/issues/3614
- SME: @Kerollmops

guimachiavelli:
Adds description of new `IS EMPTY`, `IS NULL` operators.

Closes #2415

### DIFF
--- a/learn/advanced/filtering.mdx
+++ b/learn/advanced/filtering.mdx
@@ -168,10 +168,10 @@ Both forms are equivalent.
 
 #### `IS EMPTY`
 
-The `IS EMPTY` operator selects documents with fields that exist, but contain empty values. The following expression only returns documents with an empty synopsis field:
+The `IS EMPTY` operator selects documents in which the specified attribute exists but contains empty values. The following expression only returns documents with an empty `overview` field:
 
 ```
-synopsis IS EMPTY
+overview IS EMPTY
 ```
 
 `IS EMPTY` matches the following JSON values:
@@ -193,10 +193,10 @@ Both forms are equivalent.
 
 #### `IS NULL`
 
-The `IS NULL` operator selects documents with fields that exist, but contain `null` values. The following expression only returns documents with a `null` synopsis field:
+The `IS NULL` operator selects documents in which the specified attribute exists but contains a `null` value. The following expression only returns documents with a `null` `overview` field:
 
 ```
-synopsis IS NULL
+overview IS NULL
 ```
 
 Use `NOT` to build the negated form of `IS NULL`:
--- a/learn/advanced/filtering.mdx
+++ b/learn/advanced/filtering.mdx
@@ -92,6 +92,8 @@ Meilisearch supports the following filter operators:
 - [Inequality `!=`](#inequality)
 - [Comparison `> < <= >=`](#comparison)
 - [`EXISTS`](#exists)
+- [`IS EMPTY`](#is-empty)
+- [`IS NULL`](#is-null)
 - [`IN`](#in)
 - [`NOT`](#not)
 - [`AND`](#and)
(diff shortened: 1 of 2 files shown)


learn/advanced/filtering.mdx
Kerollmops, line 1: Can we add that both operators interact with the `EXISTS` operator too? In the sense that `IS NOT NULL` and `IS NOT EMPTY` can return `NOT EXISTS` documents.

Kerollmops (changes requested):
I like what you've done, thank you 🥨

learn/advanced/filtering.mdx
```diff
@@ -92,6 +92,8 @@ Meilisearch supports the following filter operators:
 - [Inequality `!=`](#inequality)
 - [Comparison `> < <= >=`](#comparison)
 - [`EXISTS`](#exists)
+- [`IS EMPTY`](#is-empty)
+- [`IS NULL`](#is-null)
 - [`IN`](#in)
 - [`NOT`](#not)
 - [`AND`](#and)
```
maryamsulemani97, line 99: I think we can remove this list, WDYT? It was a good idea when we didn't have the "on this page sidebar" allowing users to jump between operators without scrolling

learn/advanced/filtering.mdx
```diff
@@ -164,6 +166,48 @@ NOT release_date EXISTS
...
+- `[]`
+- `{}`
+
+Meilisearch does not treat `null` values as empty. To match `null` fields, use the [`IS NULL`](#is-null) operator.
+
+Use `NOT` to build the negated form of `IS EMPTY`:
+
+```
+attribute IS NOT EMPTY
+NOT attribute IS EMPTY
```
maryamsulemani97, line 189: For consistency, can we use a[ real attribute](https://www.meilisearch.com/docs/learn/advanced/filtering#configuring-filters:~:text=the%20following%20fields%3A-,JSON,-%5B%0A%20%20%7B%0A%20%20%20%20%22id) instead of "attribute"?

learn/advanced/filtering.mdx
```diff
@@ -164,6 +166,48 @@ NOT release_date EXISTS
...
+
+```
+overview IS NULL
+```
+
+Use `NOT` to build the negated form of `IS NULL`:
+
+```
+attribute IS NOT NULL
+NOT attribute IS NULL
```
maryamsulemani97, line 206: Same as previous comment

learn/advanced/filtering.mdx
guimachiavelli: Duh, I initially used "overview", but then changed it when I was reviewing because I was comparing it with the spec.

maryamsulemani97 (approved):
🐠

Kerollmops (approved):
👌

PR #2426, opened 2023-05-15
v1.2: update postman

### LINKED ISSUE #2408: v1.2: Update Postman
Update `.vuepress/public/postman/meilisearch-collection.json`

maryamsulemani97:
closes #2408

### DIFF
--- a/assets/misc/meilisearch-collection-postman.json
+++ b/assets/misc/meilisearch-collection-postman.json
@@ -1,7 +1,7 @@
 {
 	"info": {
-		"_postman_id": "0807ba0a-e13a-49f7-970f-2f5b0eec892b",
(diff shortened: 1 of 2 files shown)


curquiza (changes requested):
Thank you Maryam

- ⚠️ The delete by filter route does not work, URL is missing.
- I recommend using `genre` instead of `genres` otherwise the examples you put does not work with the other basic examples of the collection

curquiza (approved):
Thank you Maryam!
LGTM 👌

guimachiavelli (approved):
LGTM

Also, to build on what @curquiza said—it's actually a good idea to use postman daily, when we need to set up a clean instance quickly or test whether an API route or etc works as we think it does.

PR #2449, opened 2023-06-01
remove SQL from filter expressions

maryamsulemani97:
This PR removes "```SQL" from filter expressions that aren't using SQL.


guimachiavelli (approved):
🧋 

In case you're curious why we have those: in the previous website, code blocks didn't show the language we specified, so we tagged some of the more complex filter expressions with `SQL` so they were easier to parse visually even if we knew they weren't SQL.

Issue #2434, opened 2023-05-17
Breakdown “Advanced topics”: Add Tasks and async workflow section


maryamsulemani97:
_The issue will be updated as we work on it_
- Create a new "Tasks and async workflow” section 
  - Move the tasks and async operations guide here and break it into multiple guides:
    - [x] Filtering and paginating tasks
    - [x] Separate “understanding tasks” section to serve as an intro guide
 - Add redirects

PR #2455, opened 2023-06-06
Breakdown Advanced topics: Add Tasks and async workflow section

maryamsulemani97:
closes https://github.com/meilisearch/documentation/issues/2434

This PR:
- Adds a new section called "Tasks and asynchronous workflow" after the "Core concepts" section
- Adds two new pages to this section:
  -  Asynchronous operations
  - Filtering and paginating tasks
- Moves both pages to a new `async` folder under `learn`

Redirects added in: https://github.com/meilisearch/meilisearch.com/pull/130/files

### DIFF
--- a/config/sidebar-learn.json
+++ b/config/sidebar-learn.json
@@ -109,7 +109,7 @@
         "routes": [
             {
                 "source": "learn/async/asynchronous_operations.mdx",
-                "label": "Task and asynchronous operations",
+                "label": "Tasks and asynchronous operations",
                 "slug": "asynchronous_operations"
             },
             {
--- a/config/sidebar-learn.json
+++ b/config/sidebar-learn.json
@@ -104,6 +104,22 @@
       ]
   },
   {
+        "title": "Tasks and asynchronous workflow",
+        "slug": "async",
+        "routes": [
+            {
+                "source": "learn/async/asynchronous_operations.mdx",
+                "label": "Asynchronous operations",
+                "slug": "asynchronous_operations"
+            },
+            {
+                "source": "learn/async/filtering_and_paginating_tasks.mdx",
+                "label": "Filtering and paginating tasks",
+                "slug": "filtering_and_paginating_tasks"
+            }
+    ]
+  }, 
+  {
       "title": "Configuration",
       "slug": "configuration",
       "routes": [
@@ -191,11 +207,6 @@
 			"title": "Advanced topics",
 			"slug": "advanced",
 			"routes": [
-					{
-							"source": "learn/advanced/asynchronous_operations.mdx",
-							"label": "Tasks and asynchronous operations",
-							"slug": "asynchronous_operations"
-					},
 					{
 							"source": "learn/advanced/filtering.mdx",
 							"label": "Filtering",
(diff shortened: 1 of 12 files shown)


dichotommy (commented):
Hey, this looks pretty good to me! I think we can go forward with it if @guimachiavelli also approves.

I think that the asynchronous operations page is still quite broad and could be split further, but before doing so I would like to have a better understanding of what information users are looking for about tasks, and what their goals are.

One small suggestion would be renaming the "filtering and paginating tasks" page to something like "Managing the task database" (and of course changing the slug). I think a title like this would have more longevity and flexibility if we wanted to add additional topics, such as deleting tasks, or even the "task queue" + canceling tasks.

WDYT @guimachiavelli @maryamsulemani97 ?

guimachiavelli:
I think @dichotommy's proposed change can work. It's a bit of a shame that "manage" is a lot more vague than "filter" and "paginate", but I agree the very fact it's more ambiguous will make it easier to update it in the future.

Other than that, my only comment is whether it's a good idea to prefer "asynchronous operations" over "tasks". As far as I know, the preferred term throughout the docs is "tasks"—we speak about the "task database", for instance, not the "asynchronous operations database". 

An example scenario to illustrate my concern: a user adds a document and receives a (summarised) task object. Will they know the page they need is "asynchronous operations"?

maryamsulemani97:
>Will they know the page they need is "asynchronous operations"?

The section is called "Tasks and asynchronous workflow". Should we change the page name to "Tasks and asynchronous operations"(the current name)?

maryamsulemani97:
We use the term "asynchronous operations" a few times in the docs, so let's go with "Tasks and asynchronous operations."

dichotommy:
Is there a PR adding redirects in `meilisearch/meilisearch.com` associated with this one?

config/sidebar-learn.json
```diff
@@ -103,6 +103,22 @@
 					}
       ]
   },
+  {
+        "title": "Tasks and asynchronous workflow",
+        "slug": "async",
+        "routes": [
+            {
+                "source": "learn/async/asynchronous_operations.mdx",
+                "label": "Task and asynchronous operations",
```
guimachiavelli, line 112: Any reason for the singular here? Plural sounds better to me.

```suggestion
                "label": "Tasks and asynchronous operations",
```

learn/async/asynchronous_operations.mdx
```diff
@@ -1,7 +1,7 @@
 ---
 sidebarDepth: 3
 ---
-# Tasks and asynchronous operations
+# Task and asynchronous operations
```
guimachiavelli, line 4: Same as my previous comment. Since we're talking about a general type of thing, plural seems more adequate. 

At the very least, it should be consistent: either both plural or both singular, no?

```suggestion
# Tasks and asynchronous operations
```

guimachiavelli (changes requested):

config/sidebar-learn.json
maryamsulemani97: My bad. It was a typo 🙏

learn/async/asynchronous_operations.mdx
maryamsulemani97: I copied the label from sidebar-learn.json 🤡

guimachiavelli (approved):

PR #2456, opened 2023-06-08
Update `relevancy.mdx` to remove h3s from containers+add missing tab

maryamsulemani97:
1. This PR removes the following headings from containers:
Screenshot 2023-06-07 at 16 35 38
These headings appear in the sidebar when they shouldn't

2. Updates `Tabs.Container labels` to add "Exactness". The "Exactness" tab is currently missing from the container:
Screenshot 2023-06-08 at 12 18 04

### DIFF
--- a/learn/core_concepts/relevancy.mdx
+++ b/learn/core_concepts/relevancy.mdx
@@ -83,7 +83,7 @@ Results are sorted by **the similarity of the matched words with the query words
 
 ![Demonstrating the typo ranking rule by searching for 'vogli'](https://raw.githubusercontent.com/meilisearch/documentation/main/assets/images/ranking-rules/vogli3.png)
 
-**Typo**
+##### Typo
 
 - `vogli`: 0 typo
 - `volli`: 1 typo
@@ -95,7 +95,7 @@ The `typo` rule sorts the results by increasing number of typos on matched query
 <Tabs.Content label="Proximity">
 ![Demonstrating the proximity ranking rule by searching for 'new road'](https://raw.githubusercontent.com/meilisearch/documentation/main/assets/images/ranking-rules/new_road.png)
 
-**Proximity**
+##### Proximity
 
 The reason why `Creature` is listed before `Mississippi Grind` is because of the `proximity` rule. The smallest **distance** between the matching words in `creature` is smaller than the smallest **distance** between the matching words in `Mississippi Grind`.
 
@@ -105,7 +105,7 @@ The `proximity` rule sorts the results by increasing distance between matched qu
 <Tabs.Content label="Attribute">
 ![Demonstrating the attribute ranking rule by searching for 'belgium'](https://raw.githubusercontent.com/meilisearch/documentation/main/assets/images/ranking-rules/belgium.png)
 
-**Attribute**
+##### Attribute
 
 `If It's Tuesday, This must be Belgium` is the first document because the matched word `Belgium` is found in the `title` attribute and not the `overview`.
 
@@ -116,7 +116,7 @@ The `attribute` rule sorts the results by [attribute importance](/learn/core_con
 <Tabs.Content label="Exactness">
 ![Demonstrating the exactness ranking rule by searching for 'Knight'](https://raw.githubusercontent.com/meilisearch/documentation/main/assets/images/ranking-rules/knight.png?raw=true)
 
-**Exactness**
+##### Exactness
 
 `Knight Moves` is displayed before `Knights of Badassdom`. `Knight` is exactly the same as the search query `Knight` whereas there is a letter of difference between `Knights` and the search query `Knight`.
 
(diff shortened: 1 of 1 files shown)


guimachiavelli (changes requested):
Hmmmm…isn't the problem here that we're using `h3`s to tag headings that should be `h5`s?

Like, the way I read that section is:

```markdown
### Built-in rules
#### Words
#### Typo
[…]
#### Examples
##### Typo
##### Words
[etc]
```

- We have a section for "Built-in Rules" (h3)
- This section contains subsections for ranking rules (h4) and a final subsection for "Examples" (also h4)
- "Examples" itself contains subsections for each individual example  (h5)

maryamsulemani97:
>"Examples" itself contains subsections for each individual example (h5)

"Examples" is a container with 4 tabs. We have an h3 for each of these tabs.

I replaced the h3s with h5s

dichotommy:
I'm going to contradict @guimachiavelli 's comment and say that you had it right the first time Maryam: these should not be headers, they should just be bold text. 

The problem is that headers and tabs do not play well together. Let's take `Words` as an example. If you try to navigate to `#words` anchor, either from the right-hand sidebar or by entering the link, it will not work. Either you will see the top of the page because the anchor was not found, or you will see the correct part of the page but the wrong tab (because `typo` tab is open by default).

maryamsulemani97:
>The problem is that headers and tabs do not play well together. Let's take Words as an example. If you try to navigate to #words anchor, either from the right-hand sidebar or by entering the link, it will not work. Either you will see the top of the page because the anchor was not found, or you will see the correct part of the page but the wrong tab (because typo tab is open by default).

Reverted the H5s to bold

guimachiavelli:
I still disagree with @dichotommy. Headings are not necessarily links—they are semantic elements indicating relationship and information hierarchy. `h[number]` generating an anchor link is a UX thing we actively program for.

Using bold might solve the immediate visual issue, but at the cost of weakening the connection between e.g. "typo" and the content that comes after it.

As far as I know, `h5`s won't appear on the right-hand menu, so while users might still click on the heading and get the anchor link, the impact will be relatively limited. Because of that, I believe `h5`s are better and lead to sturdier data on the long run.

The ideal solution, in my opinion, would be to use `h5`s and commission basement to find a proper solution—such as not adding anchor links to headings inside tabs, or, even better, somehow detecting when a URI fragment is pointing to something in a hidden tab and display it.

learn/core_concepts/relevancy.mdx
```diff
@@ -77,13 +77,13 @@ Results are sorted by **the similarity of the matched words with the query words
...
 
-<Tabs.Container labels={["Typo", "Proximity", "Attribute"]}>
+<Tabs.Container labels={["Typo", "Proximity", "Attribute", "Exactness"]}>
 
 <Tabs.Content label="Typo">
 
 ![Demonstrating the typo ranking rule by searching for 'vogli'](https://raw.githubusercontent.com/meilisearch/documentation/main/assets/images/ranking-rules/vogli3.png)
 
-### Typo
+**Typo**
```
maryamsulemani97, line 86:
```suggestion
##### Typo
```

maryamsulemani97:
Visually there isn't any difference between bold and h5:
Screenshot 2023-06-21 at 11 59 47

Screenshot 2023-06-21 at 11 51 10

>Let's take Words as an example. If you try to navigate to #words anchor, either from the right-hand sidebar or by entering the link, it will not work. Either you will see the top of the page because the anchor was not found, or you will see the correct part of the page but the wrong tab (because typo tab is open by default).

You don't get anchor links by clicking on H5s in a container. Since users won't get anchor links, they can't use them so this won't be a problem. 

Something I didn't consider, but Gui brought up, is the scraper. Bold text is just like ordinary text (content) to it.

learn/core_concepts/relevancy.mdx
```diff
@@ -95,7 +95,7 @@ The `typo` rule sorts the results by increasing number of typos on matched query
 <Tabs.Content label="Proximity">
 ![Demonstrating the proximity ranking rule by searching for 'new road'](https://raw.githubusercontent.com/meilisearch/documentation/main/assets/images/ranking-rules/new_road.png)
 
-### Proximity
+**Proximity**
```
maryamsulemani97, line 98:
```suggestion
##### Proximity
```

learn/core_concepts/relevancy.mdx
```diff
@@ -105,7 +105,7 @@ The `proximity` rule sorts the results by increasing distance between matched qu
 <Tabs.Content label="Attribute">
 ![Demonstrating the attribute ranking rule by searching for 'belgium'](https://raw.githubusercontent.com/meilisearch/documentation/main/assets/images/ranking-rules/belgium.png)
 
-### Attribute
+**Attribute**
```
maryamsulemani97, line 108:
```suggestion
##### Attribute
```

learn/core_concepts/relevancy.mdx
```diff
@@ -116,7 +116,7 @@ The `attribute` rule sorts the results by [attribute importance](/learn/core_con
 <Tabs.Content label="Exactness">
 ![Demonstrating the exactness ranking rule by searching for 'Knight'](https://raw.githubusercontent.com/meilisearch/documentation/main/assets/images/ranking-rules/knight.png?raw=true)
 
-### Exactness
+**Exactness**
```
maryamsulemani97, line 119:
```suggestion
##### Exactness
```

guimachiavelli (approved):
🥼

PR #2465, opened 2023-06-15
change "async" site route to "tasks"

dichotommy:
Pages in our new "tasks and asynchronous operations" section are 404ing. We're not sure why, since the sidebar file appears to be well formatted. On the off chance that the error is being caused by our use of the word "async" in the URL, we will try changing it to "tasks"

### DIFF
--- a/config/sidebar-learn.json
+++ b/config/sidebar-learn.json
@@ -100,7 +100,7 @@
   },
   {
         "title": "Tasks and asynchronous workflow",
-        "slug": "async",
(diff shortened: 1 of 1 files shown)


guimachiavelli (changes requested):
I think we also need to fix all links pointing to these pages?

Either a cmd+f or looking at what @maryamsulemani97 changed in #2455

dichotommy:
The 404 was caused by a formatting error in docs scraper file, so I'm closing this

Issue #2377, opened 2023-04-24
Improve readability


dichotommy:
One of our planned projects for Q2 2023 is a **readability review of our documentation**.

~Using aggregated readability metrics, we will calculate a score for each of our pages. We will then identify "low-hanging fruit" to increase the scores across the board (e.g. lowering the site's sentence-length limit) and take action.~

~For pages with unusually high readability scores, we may take special action, especially if the page is involved in new user onboarding.~

_Edit by Maryam (June 12, 2023)_

## Phase one: visual effectiveness

- [ ] Reduce prose chunks
  - This can be done by:
    - Adding new headers to the prose
    - Converting the text to bullets or tables 
    - Add more code samples or replace existing text with code samples
- [x] Reduce the use of containers (tips, warnings, danger boxes) 
   - Related issue: https://github.com/meilisearch/documentation/issues/2474
   - Related PR: https://github.com/meilisearch/documentation/pull/2466
- [ ] Separate understanding-oriented sections from task-oriented sections
   - Users should be able to get a task done without understanding how the engine works

See [Notion page](https://www.notion.so/meilisearch/Readability-pages-acc3dd3918574bbd90c03374691c6c53)

guimachiavelli:
Due to the reorg and change of priorities, we do not have enough resources to allocate to assess readability metric.

PR #2466, opened 2023-06-15
Reduce containers

### LINKED ISSUE #2474: Reduce containers
As part of #2377, we'd like to reduce containers(warnings, danger, notes, and tips) in the docs.

When to use what container?

- Tips for any additional information users don’t need to understand the guide/concept but might be helpful
- Danger when doing or missing something will result in an error, unexpected behavior, or mess up the database
- Warnings are like beware signs
- Notes for “important” information. Anything we want to highlight or feel might get lost if added as regular text

We'll remove any containers that don't follow the above guidelines.

maryamsulemani97:
Part of https://github.com/meilisearch/documentation/issues/2377

This PR reduces containers from the `learn` section. It moves any content from the container that isn't important and adds it as regular text.

Closes #2474

### DIFF
--- a/learn/fine_tuning_results/faceted_search.mdx
+++ b/learn/fine_tuning_results/faceted_search.mdx
@@ -116,21 +116,15 @@ The following response shows the facet distribution when searching for `classics
 }
 ```
 
-`facetDistribution` contains an object for every attribute passed to the `facets` parameter. Each object contains the different values for that attribute and the count of matching documents with that value. Meilisearch does not return empty facets: if there are no results for the Arabic language, it will not be present in `facetDistribution`.
+`facetDistribution` contains an object for every attribute passed to the `facets` parameter. Each object contains different values for that attribute and the count of matching documents with that value. Meilisearch does not return empty facets: if there are no results for the Arabic language, it will not be present in `facetDistribution`.
 
-<Capsule intent="note">
 By default, `facets` returns a maximum of 100 facet values for each faceted field. You can change this value using the `maxValuesPerFacet` property of the [`faceting` index settings](/reference/api/settings#faceting).
-</Capsule>
 
 ### Facet stats
 
-When using the `facets` parameter, Meilisearch results include a `facetStats` object. `facetStats` contains the lowest (`min`) and highest (`max`) numerical values across all documents in each facet.
+When using the `facets` parameter, Meilisearch results include a `facetStats` object. `facetStats` contains the lowest (`min`) and highest (`max`) numerical values across all documents in each facet. Meilisearch ignores numeric strings like `"21"` when computing `facetStats`.
 
 `facetStats` is useful when creating UI components such as range sliders. These allow users to refine their search by selecting from a range of facet values.
-
-<Capsule intent="note">
-Meilisearch ignores numeric strings like `"21"` when computing `facetStats`.
-</Capsule>
 
 The following response shows the lowest and highest book ratings when searching for `"classic"`:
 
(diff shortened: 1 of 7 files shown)


guimachiavelli (changes requested):
Couple minor suggestions and we should be good to go.

learn/fine_tuning_results/faceted_search.mdx
```diff
@@ -116,22 +116,16 @@ The following response shows the facet distribution when searching for `classics
 }
 ```
 
-`facetDistribution` contains an object for every attribute passed to the `facets` parameter. Each object contains the different values for that attribute and the count of matching documents with that value. Meilisearch does not return empty facets: if there are no results for the Arabic language, it will not be present in `facetDistribution`.
+`facetDistribution` contains an object for every attribute passed to the `facets` parameter. Each object contains the returned values for that attribute and the count of matching documents with that value. Meilisearch does not return empty facets: if there are no results for the Arabic language, it will not be present in `facetDistribution`.
 
-<Capsule intent="note">
```
dichotommy, line 121: I would keep this container and make it a "tip". It's extra information that could be helpful.

learn/fine_tuning_results/faceted_search.mdx
```diff
@@ -116,22 +116,16 @@ The following response shows the facet distribution when searching for `classics
...
 By default, `facets` returns a maximum of 100 facet values for each faceted field. You can change this value using the `maxValuesPerFacet` property of the [`faceting` index settings](/reference/api/settings#faceting).
-</Capsule>
 
 ### Facet stats
 
-When using the `facets` parameter, Meilisearch results include a `facetStats` object. `facetStats` contains the lowest (`min`) and highest (`max`) numerical values across all documents in each facet.
+When using the `facets` parameter, Meilisearch results include a `facetStats` object. `facetStats` contains the lowest (`min`) and highest (`max`) numerical values across all documents in each facet. Meilisearch ignores numeric strings like `"21"` when computing `facetStats`.
```
dichotommy, line 125:
```suggestion
When using the `facets` parameter, search results include a `facetStats` object. `facetStats` contains the lowest (`min`) and highest (`max`) numerical values across all documents in each facet. Only numeric values are considered; string values are ignored, even if the string contains a number.
```

learn/index_settings/synonyms.mdx
```diff
@@ -98,17 +98,12 @@ To create a mutual association between four words, this is the JSON syntax that
 
 ## Multi-word synonyms
 
-Meilisearch treats multi-word synonyms as [phrases](/reference/api/search#phrase-search-2).
+Meilisearch treats multi-word synonyms as [phrases](/reference/api/search#phrase-search-2). Multi-word synonyms are limited to a maximum of **three words**. For example, although you could make `League of Legends` and `LOL` into synonyms, you could not do the same for `The Lord of the Rings` and `LOTR`.
 
 Take note that **multi-word phrases are treated differently** than associations between individual words.
 
 When a multi-word phrase is considered the synonym of another word or phrase, the **exact search query will always take precedence over its synonym(s)**.
```
dichotommy, line 105:
```suggestion
Multi-word synonyms follow a different set of rules than single-word synonyms:

- When a multi-word phrase is set as a synonym, **the exact search query will always take precedence over its multi-word synonym(s)**
- Meilisearch treats multi-word synonyms as [phrases](/reference/api/search#phrase-search-2)
- Multi-word synonyms are limited to a maximum of three words

```

dichotommy (changes requested):

guimachiavelli (changes requested):
Do take a look at @dichotommy's requested changes and re-request my review once they're done.

guimachiavelli (approved):

PR #2470, opened 2023-06-19
Update `TO` operator syntax

### LINKED ISSUE #2463: Fix TO filter operator example in Learn > Advanced > filtering
The example for the TO operator in "[Learn > Advances > Filtering](https://www.meilisearch.com/docs/learn/advanced/filtering#to)" is wrong:

Wrong (current):
```
rating.users 80 TO rating.users 89
```

Right (should be):
```
rating.users 80 TO  89
```

maryamsulemani97:
closes #2463 

Updates the syntax for the `TO` operator in `filtering.mdx`


guimachiavelli (approved):
Preemptively approving this under the assumption you have tested both the old and the new examples via the command-line or postman.

If you haven't done that, please do, just so we can be sure 1) we're actually something that is wrong and, more importantly, 2) the new example works.

PR #2475, opened 2023-06-20
update aws

maryamsulemani97:


guimachiavelli (changes requested):
I've made a few comments at the issues that motivated this PR. Let me know if you disagree, otherwise please implement them here and request another review ✨ 

Also, in the future, try to make PR titles clearer. For example, in this case describe what these updates actually are or what you intend to achieve with them.

guimachiavelli (approved):

Issue #2403, opened 2023-05-03
Misleading sentence on the search API


irevoire:
Hey,

I was looking for some info on how the filter works in meilisearch and noticed we had this table summarizing all the types of available search parameters.
[This one](https://github.com/meilisearch/documentation/blob/a0ad491bc474bedc25ffaee509812c551a5e6168/reference/api/search.mdx?plain=1#L41) seems wrong, though. A filter can be **either** a `string`, an `array of strings`, or an `array of array of strings`.

I don't know if it's a good idea to show all these types, but I would be in favor of either;
- showing the three options
- only saying `string`

dichotommy:
Perhaps we could change it to `string`* and link to the [section of the filtering guide about filter expressions](https://www.meilisearch.com/docs/learn/advanced/filtering#filter-expressions).

maryamsulemani97:
@irevoire

Does this apply to `facets` parameter too?

PR #2493, opened 2023-06-26
update filter data type

maryamsulemani97:
closes https://github.com/meilisearch/documentation/issues/2403

Updates the `filter` search parameter type to String* and links it to the [filtering guide section on filter expressions](https://www.meilisearch.com/docs/learn/advanced/filtering#filter-expressions)


guimachiavelli:
Do we use `String*` anywhere else in the docs? Is there precedence in our website or elsewhere? 

I don't think I've seen it before and personally I'm not confident I would have understood it without knowing what was there before. Then again, I might be dumb.

maryamsulemani97:
>Do we use String* anywhere else in the docs? Is there precedent in our website or elsewhere?

We don't. It was either `String*` or `string, an array of strings, or an array of array of strings`. @dichotommy suggested using `string*` and linking to the [section of the filtering guide about filter expressions](https://www.meilisearch.com/docs/learn/advanced/filtering#filter-expressions).

guimachiavelli:
Ok, then, thanks for the info. Unless we can find a significant number of other docs sites using `[String*]`, I strongly believe we shouldn't use it. There's no precedent I'm aware of, it does not mirror common programming language conventions (e.g. `String[]`), and it sort of looks like regex, but follows a different logic.

I think it's possible to come up with alternatives. A few that come to mind, in no particular order:

1. Only use "strings" on the table and describe other options in the linked section
2. Use something like  "string, other" (we'd probably need to find something instead of `other`)
3. "Consult dedicated section for more information"

guimachiavelli (approved):

PR #2494, opened 2023-06-27
[MEI-18]: Move redirects from meilisearch.com repo

justkahdri:
# Pull Request

## Related issue
Fixes [MEI-18](https://linear.app/basement/issue/MEI-18/move-redirects-inside-documentation-repo)

## What does this PR do?
- Moves redirects already present in `next.config.js` to the new `redirects.json` inside documentation.


justkahdri:
@guimachiavelli here are all the redirects previously added to the [meilisearch.com](https://github.com/meilisearch/meilisearch.com) repo. 

Had to remove some comments since JSON syntax don't allow them, maybe could be a good idea to consider yaml for config files since they allow comments, but I'm not a fan of the array syntax inside YAML files. I'll let you and your team decide since you'll be the ones writing it more.

config/redirects.json
```diff
@@ -0,0 +1,132 @@
...
+  },
+  {
+    "source": "/docs/learn/advanced/asynchronous_operations#filter-by-date",
+    "destination": "/docs/learn/async/managing_tasks#filter-by-date"
+  },
+  {
+    "source": "/docs/learn/advanced/asynchronous_operations#combine-filters",
+    "destination": "/docs/learn/async/managing_tasks#combine-filters"
+  }  
+]
```
guimachiavelli, line 132:
```suggestion
]

```

guimachiavelli:
thanks for this, @justkahdri!

I also prefer JSON in general and dislike YAML arrays. 

I'll discuss this with the team, though, just to be sure we all agree.

guimachiavelli:
@justkahdri, this is possibly a terrible idea, but what would be the downsides to adding extra fields with "metadata" to the redirects file? For example:

```json
{
  "source": "[old URL]",
  "destination": "[new URL]",
  "permanent": true,
  "comment": "moved [subsection] into new page"
}
```

Would the `comment` field be ignored, or would it trigger an error somewhere along the process?

justkahdri:
Btw from my side, we're ready to release this feature, so when you're ready to merge this let me know and I'll push the change on the other repo. 👍🏼

guimachiavelli (approved):
LGTM

PR #2507, opened 2023-07-10
[MEI-23]: Broken links checker

justkahdri:
# Pull Request
This PR adds a new action to check if any broken links are introduced before merging to `main`.

## Related issue
[MEI-23](https://linear.app/basement/issue/MEI-23/broken-links-checker)


curquiza (commented):
Thanks for the PR!

Not sure to understand where this new check comes from (the one in red)

Capture d’écran 2023-07-20 à 16 57 27

justkahdri:
> Thanks for the PR!
> 
> Not sure to understand where this new check comes from (the one in red)

@curquiza is created inside the action, I'm not sure why there are two of them, but the one called `Docs Link Validation` is the important one. Once the errors described [here](https://github.com/meilisearch/documentation/pull/2507#issuecomment-1640841215) are fixed, the action should pass.

curquiza:
Hum ok, I understand I guess, the line is about the github action itself and the second line is to block the PR for the merge because a comment has been generated

.github/actions/validate-docs-links/README.MD
```diff
@@ -0,0 +1,17 @@
+# GitHub Action for Validating Documentation Links
```
curquiza, line 1: I was wondering: what are the advantages of creating our own checker (and maintain it in this repo) compared to the current solutions we could already find on the GitHub action marketplace?
Like this one: https://github.com/marketplace/actions/markdown-link-check
(I took the first link I found, I did not do any investigation)

justkahdri: From the ones I checked, they don't match our necessity. 

Taking in consideration how the documentation is structured we're not just checking `mdx` files to one another, we also need to retrieve the slugs from the config JSON files and check those too. 

Also keeping the action "in house" would facilitate us any upgrade needed in the future. (e.g. skip broken links handled by the `redirects.json`)

curquiza: > Taking in consideration how the documentation is structured, we're not just checking mdx files to one another, we also need to retrieve the slugs from the config JSON files and check those too.

I guessed the reason would be something like this, thank you for taking the time to explain 😊

curquiza (commented):
Ok for me 👌
I don't approve yet because I think we should fix the dead link before continuing.
I would like to see everything works when there are no dead links.
Do we plan to do another PR to fix the dead links or will it be done in this PR?

justkahdri:
I think we can create a new PR to fix the links, but I don't have the context to know where they should point to CC @guimachiavelli

guimachiavelli:
yeap, it's in my list, just haven't gotten around to it yet as I'm focussing on getting everything ready for v1.3. I should have some time on thursday, if all goes according to plan.

guimachiavelli:
Hmmm…I just merged a fix into this branch, but link validation failed again. 

It seems none of the broken links in the new report had been previously detected (I made a copy of the original report on #2521).

curquiza:
Hey, looks like a [broken link has been seen by a user](https://github.com/meilisearch/documentation/issues/2526) despite @guimachiavelli fixed all the links reported by the checker in https://github.com/meilisearch/documentation/pull/2524

curquiza:
Oh yes you are right, my bad, I'm tired 😅 The best would have been to merge it on `main` and then rebase this PR.

Knowing this, you are right, let's do this. It will avoid the broken links in documentation

I will temporarily change the settings of the repo to merge despite the failure. I will inform gui about this so that he's aware of it when coming back from holiday

PR #2533, opened 2023-08-11
[MEI-23]: Review broken links checker

justkahdri:
# Pull Request

## Related issue
Fixes #<issue_number>

## What does this PR do?
- ...

## PR checklist
- [ ] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [ ] Have you read the contributing guidelines?
- [ ] Have you made sure that the title is accurate and descriptive of the changes?

Thank you so much for contributing to Meilisearch!


justkahdri:
Hey @curquiza @guimachiavelli, on this PR I added the possibility of running the links checker locally. To do it the steps would be:

1. Install `node.js` in the case you don't have it already. https://nodejs.org/en
2. Install `yarn` in the case you don't have it already. https://classic.yarnpkg.com/lang/en/docs/install/#mac-stable
3. Run `yarn links-check` and wait for the result

About the broken links that wasn't caught before, I'm still trying to replicate the bug and fix it, but most probably it skipped those links because of the quantity. I'm confident it won't happen that a PR breaks a link and the action passes, giving a false positive.

guimachiavelli:
Just fixed the reported broken links and it seems nothing new has come up.

Regarding the cli tool: works great. Only comment I have is that the `type` column is always empty. Not a deal breaker, though.

Final question: how complex would it be to add line numbers?

justkahdri:
>Final question: how complex would it be to add line numbers?

The line number for the file the broken link comes from? It is somewhat complex indeed, because the links are retrieved from the processed mdx tree and not from the plain text itself. I could research and scope it for next week if you want, but I can't add that feature to this PR today unfortunately.

guimachiavelli (approved):

PR #2538, opened 2023-08-24
Add link to document pinning blog posts

### LINKED ISSUE #2517: How-to guide on document pinning
@meilisearch/product-team and @meilisearch/devrel-team are currently working on communicating workarounds to cover functionality Meilisearch does not yet offer. One of these is document pinning: in short, ensuring a document is always the top search result.

The @meilisearch/devrel-team has written a blog teaching users how to pin documents with React and InstantSearch. They have requested to feature this content on the docs since the documentation has higher visibility than the blog. 

As previously stated, the docs currently avoids publishing guides and tutorials with third-party dependencies. At the same time, @Strift is correct in saying docs users who search for document pinning will not find anything at the moment.

Based on our discussion so far, I see three ways of doing this:

1. Publishing the blog post under "Cookbooks" with minor editorial changes bringing it inline with the rest of the docs;
2. Finding a spot in the docs where we can explicitly mention "document pinning" and link this mention to the blog post;
3. Reworking the blog post in a dependency-agnostic version using e.g. vanilla javascript (similar to what we do for [search result pagination](https://www.meilisearch.com/docs/learn/front_end/pagination)).

My personal preference is #3, since this might make the resource uniformly useful to all front-end devs, instead of only those who use React. On the other hand, depending on implementation complexity a vanilla JS how-to guide might end up being counter-productive. Do we have any data regarding front-end frameworks preferred by our users?

guimachiavelli:
Fixes #2517

### DIFF
--- a/learn/core_concepts/relevancy.mdx
+++ b/learn/core_concepts/relevancy.mdx
@@ -1,5 +1,4 @@
 ---
-title: Meilisearch Documentation — Relevancy
 sidebarDepth: 4
 ---
 
--- a/learn/core_concepts/relevancy.mdx
+++ b/learn/core_concepts/relevancy.mdx
@@ -178,7 +178,7 @@ In general, `sort` will be most useful when you want to allow users to define wh
 Custom ranking rules, instead, are always active once configured and will be useful when you want to promote certain types of results. A good use-case for custom ranking rules is ensuring discounted products in a webshop always feature among the top results.
 
 <Capsule intent="tip" title="Promoting search results and document pinning">
-If you want to promote or pin specific documents so they are displayed more prominently than other search results, consult these Meilisearch blog articles on [implementing promoted search results with React InstantSearch](https://blog.meilisearch.com/promoted-search-results-with-react-instantsearch) and [document boosting](https://blog.meilisearch.com/document-boosting).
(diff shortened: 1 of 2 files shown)


Strift (commented):
Two piece of feedback on my end:
- It's best if we can make it so searching for "Document boosting" brings results, otherwise I fear user might not find it
- I tend to see both guides as a workaround for the same issue; one backend implementation, and one frontend implementation. Both have downsides/limitations. I'm not sure this is conveyed by the current wording.

guimachiavelli:
Hi, @Strift!

Responding:

> It's best if we can make it so searching for "Document boosting" brings results, otherwise I fear user might not find it

Agreed, which is why I'm using both document boosting and pinning in both blocks. I can invert their usage order so "boosting" is on the `<Capsule>` title if you prefer, but this will have no impact on search results in our search bar. I still think both "pinning" and "boosting" should be present to cover both ways of referring to this feature/technique.

FYI, the docs scraper runs on the `main` branch, so the search bar will only reflect changes in this PR after it has been merged.

> I tend to see both guides as a workaround for the same issue; one backend implementation, and one frontend implementation. Both have downsides/limitations. I'm not sure this is conveyed by the current wording.

Done. If that's the angle we're going for, I'd suggest making it explicit in the blog posts' intro too.

Issue #2542, opened 2023-08-31
Meilisync guides


guimachiavelli:
Meilisearch Cloud has recently deployed a new feature to simplify keeping Meilisearch synchronized with a user's primary database. We need to write a few guides to support this feature and also cover a common user need. Content outlines and instructions have already been primarily developed by @alallema.

Writing responsibility will be shared between @meilisearch/docs-team  and @meilisearch/devrel-team, with docs supplying the first guide and devrel creating the remainder using that as a template.

## Tasks
- [x] Syncing with MySQL and Meilisync
- [x] Syncing with Postgres and Meilisync
- [ ] Syncing with MongoDB and Meilisync
- [ ] Create new "Syncing with a primary database" section under "Learn"

## References
- SME: @alallema 
- [Outline and instructions (private link)](https://www.notion.so/meilisearch/Primary-DBs-with-Meilisearch-cf98503b6e6a40308f4e273ee828820e)
- [Devrel kickoff document (private link)](https://www.notion.so/meilisearch/Syncing-Meilisearch-with-SQL-databases-23414f58997843bf97b3b5a5bbe86444)

PR #2549, opened 2023-09-07
Meilisync MySQL guide

guimachiavelli:
Partially addresses #2542

Disclaimer for reviewers: I was not able to successfully install `meilisync` on my machine. I suspect it's related to some python misconfiguration on my end, but it became a huge time drain and I'm working on a short deadline for this. The guide might be rougher than usual because of that.

### DIFF
--- a/learn/primary_db_sync/meilisync_mysql.mdx
+++ b/learn/primary_db_sync/meilisync_mysql.mdx
@@ -1,8 +1,8 @@
 # Syncing databases with MySQL and `meilisync`
 
-Though Meilisearch is a database, it is optimized for search and underperforms when used as a primary data store. It is highly recommended you use an external database to store your data and periodically synchronise it with Meilisearch to ensure comprehensive and accurate search results.
+Though Meilisearch is a database, it is not recommended you use it as your primary data store. Instead, you should use an external database to store your data and periodically synchronise it with Meilisearch.
 
-This guide uses [`meilisync`](https://github.com/long2ice/meilisync) to keep Meilisearch up to date with a MySQL database.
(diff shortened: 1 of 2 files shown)


alallema:
@guimachiavelli,
It's not good news at all if it's complicated to install. In that case, do you want me to provide the code sample for the `</Tabs.Content>`?

learn/primary_db_sync/meilisync_mysql.mdx
```diff
@@ -0,0 +1,170 @@
...
+
+If your project already has a `docker-compose.yml`, add the `meilisync` settings to the `services` field.
+
+When you are done, open your console, move to your Docker project directory, and run the following command:
+
+```sh
+docker-compose up
+```
+
+After a few seconds, Docker will inform you it has successfully installed `meilisync`.
```
alallema, line 41: It will not install it but directly run it. This is why you need a configuration file beforehand, as it is referred to in this line of the Dockerfile:
```
- ./config.yml:/meilisync/config.yml
```
It is the equivalent to this line and maybe should be moved there:
```
meilisync start
```

alallema (commented):
Congrats @guimachiavelli  🎉
I just put a comment but the rest is great for me

Strift (commented):
I've left a few minor comments. overall it looks good to me

guimachiavelli:
Thanks for the feedback, @alallema and @Strift.

I have addressed all of it, I believe, including the most important bit regarding `docker-compose up` not only building a docker image, but also running it.

Let me know if you spot anything else, otherwise please mark it as approved and I'll deploy it by the end of the day today.

Strift (approved):
Looks perfect to me!

PR #2553, opened 2023-09-12
v1.4: Separator and non-separator tokens

### LINKED ISSUE #2548: v1.4: Customizable text separators
In many cases, Meilisearch relies on specific characters to delimit words. For example, two terms separated by a period are less likely to be closely related than the same two terms separated by a single space: `[…] white. Shirt […]` vs `[…] white shirt […]`.

v1.4 allows users to customize which characters count as separators. This can be useful in specific datasets where characters that usually indicate semantic distance have a different meaning. For example, while Meilisearch considers `#` a separator by default, it has a different function in a dataset containing musical notation.

Users can indicate a character is not a separator by adding it to the `nonSeparatorTokens` list:

```bash
curl \
  -X PUT 'http://localhost:7700/indexes/articles/settings/non-separator-tokens' \
  -H 'Content-Type: application/json'  \
  --data-binary '["@", "#", "&"]'
```

Users can also indicate a string is separator by adding it to the `separatorTokensList`:

```bash
curl \
  -X PUT 'http://localhost:7700/indexes/articles/settings/separator-tokens' \
  -H 'Content-Type: application/json'  \
  --data-binary '["&sect;", "&sep"]'
```

Internally, separators are either hard or soft. New separators are always soft separators. There is no way of creating new hard separators. Hard separators can be removed, but cannot be converted into soft separators.

As described by @ManyTheFish, the pseudocode logic is:

```
IF token is in the customizable separator list
    IF token is in the engine hard separator list
        it's a hard separator
    ELSE
        it's a soft separator
ELSE
  it's not a separator
```

For usability reasons, the distinction between soft and hard separators is a purely internal one. Docs should avoid mentioning it.

## Tasks
- [ ] Update `reference/api/settings.mdx`
- [ ] Update `learn/advanced/datatypes.mdx` with link to [full list of separators](https://docs.rs/charabia/0.8.3/src/charabia/separators.rs.html#16-62)
- [ ] Update telemetry table

## References
- spec: https://github.com/meilisearch/specifications/pull/254/files
- SME: @ManyTheFish 
- issue in engine repo: https://github.com/meilisearch/meilisearch/issues/3610
- slack thread clarifying custom separators (private link): https://meilisearch.slack.com/archives/C05AAHLV6F3/p1693923610311409

guimachiavelli:
Closes #2548


curquiza (approved):
Code samples are good to me

learn/inner_workings/datatypes.mdx
```diff
@@ -12,14 +12,18 @@ String tokenization is the process of **splitting a string into a list of indivi
 
 A string is passed to a tokenizer and is then broken into separate string tokens. A token is a **word**.
 
-- For Latin-based languages, the words are separated by **space**.
-- For Kanji characters, the words are separated by **character**.
+- For Latin-based languages, the words are separated by **space**
```
ManyTheFish, line 15: It's not only for Latin-based characters but for everything, Some Languages don't use any separator but Meilisearch doesn't make any exceptions.
Let's say we find a space in the middle of a Kanji string, Meilisearch will split on it as it would make for any other script.
Moreover, Latin-based languages are not the only ones using spaces, for example, the Cyrillic Script uses some of the separators used in Latin Languages.

learn/inner_workings/datatypes.mdx
```diff
@@ -12,14 +12,18 @@ String tokenization is the process of **splitting a string into a list of indivi
...
+- For Kanji characters, the words are separated by **character**
```
ManyTheFish, line 16: It's wrong, Kanjis are no more split by character, but the words are split based on Language dictionaries depending on how well the Language is supported in Meilisearch.

learn/inner_workings/datatypes.mdx
```diff
@@ -12,14 +12,18 @@ String tokenization is the process of **splitting a string into a list of indivi
...
 
-For Latin-based languages, there are two kinds of **space separators**:
+For Latin-based languages, there are two kinds of **space separators**: soft and hard. Hard separators indicate significant context switch such as a new sentence or paragraph. Soft separators only delimit one word from another.
```
ManyTheFish, line 18: It's not specific to Latin
```suggestion
There are two kinds of **space separators**: soft and hard. Hard separators indicate significant context switch such as a new sentence or paragraph. Soft separators only delimit one word from another.
```

ManyTheFish (changes requested):

ManyTheFish (approved):

PR #2554, opened 2023-09-13
Remove sample-template.yaml because unused

curquiza:
This file is not used and make docs team to duplicate information for nothing


guimachiavelli (approved):
🐶

PR #2558, opened 2023-09-13
v1.4: Docker version bump

### LINKED ISSUE #2555: v1.4: Docker version bump
With every new release, we need to update links to our docker images so we are always referencing the most up-to-date version of Meilisearch.

guimachiavelli:
Closes #2555

### DIFF
--- a/learn/cookbooks/docker.mdx
+++ b/learn/cookbooks/docker.mdx
@@ -9,7 +9,7 @@ Docker is a tool that bundles applications into containers. Docker containers en
 Docker containers are distributed in images. To use Meilisearch, use the `docker pull` command to download a Meilisearch image:
 
 ```sh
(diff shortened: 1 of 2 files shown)


curquiza (approved):
We should create an automation for this 😂

guimachiavelli:
We probably should, but I'm not exactly sure how. I looked into creating global variables with next.js a few weeks ago, but wasn't able to get string interpolation to work on mdx files.

curquiza:
I think I know how to do it, with a CI we would run, I have one to update the Meilisearch version in the Meilisearch repo: https://github.com/meilisearch/meilisearch/blob/main/.github/workflows/update-cargo-toml-version.yml
The issue is taking the time to do it 😂

PR #2600, opened 2023-10-26
Remove vscode file

curquiza:
We should not merge this kind of file. This kind of settings should be individually managed depending on the text editor of the contributor.

Let me know if I'm wrong and I miss something @guimachiavelli


guimachiavelli:
Apologies, myself or someone else must have gotten carried away with a `git add .`. What do you thing about adding `.vscode` to `.gitignore`?

curquiza:
> Apologies, myself or someone else must have gotten carried away with a git add .. What do you thing about adding .vscode to .gitignore?

Definitely a good point! I will add it

PR #2602, opened 2023-10-26
Update dependabot for GHA only

curquiza:
Dependabot is not used anymore for npm since the website migration

I updated dependabot file to take care of github action dependencies only
We could also completely remove the file if you prefer @guimachiavelli

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,10 +1,11 @@
+# Set update schedule for GitHub Actions only
+
 version: 2
 updates:
-- package-ecosystem: npm
-  directory: "/"
-  schedule:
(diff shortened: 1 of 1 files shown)


guimachiavelli:
I'd be fine with removing it, but perhaps it's best to wait until the linters are in place?

curquiza:
I was thinking, the more we add automation, the best it is to have the dependabot updates to avoid issues with outdated GitHub actions dependencies in the future, especially because we run on `ubuntu-latest` (which is what we want for this kind of repo!)
So I would go to keep it

curquiza:
I'll let you merge if you are confortable with my point of view

PR #2628, opened 2023-11-09
Cloud analytics guide

guimachiavelli:
Adds a Meilisearch Cloud-specific guide teaching users how to activate, configure, and deactivate analytics functionality. It also includes a glossary explaining some of the terms and components available in the analytics UI.

## Notes for reviewers
This is the first round of reviews. Focus on:
1. Is the content accurate? Are there any mistakes or misrepresentations regarding usage and configuration of the feature?
2. In your opinion, does the content address all major points of the feature? Are there things missing? Is there information you think we can remove?

Additionally, a few notes:
- screenshots are not final and will be updated in the next draft


learn/cloud/analytics.mdx
```diff
@@ -0,0 +1,197 @@
...
+
+To track conversion rates, [configure your application to send `conversion` events to Meilisearch Cloud](#configuring-conversion-rate).
+
+### Click-through rate and average click position
+
+Click-through rate is the number of times users have clicked on a search result divided by the total number of searches.
+
+Average click position indicates the average position of the search results users clicked on, with 0 indicating the first search result. 
+
+In both metrics, lower values indicate searches are returning relevant results.
```
gmourier, line 160: A higher CTR is better than a lower one because it shows users engaging with the result. A low CTR could mean the results aren't relevant enough.

learn/cloud/analytics.mdx
```diff
@@ -0,0 +1,197 @@
...
+
+## Understanding the analytics interface
+
+### Total users and user ids
+
+Indicates the total number of user ids. User ids are set with the `X-MS-USER-ID` HTTP header. If `X-MS-USER-ID` is not present, user ids are generated based on `User-Agent` and `CF-Connecting-IP`.
+
+### Total searches and search IDs
+
+The total number of generated search ids. Meilisearch Cloud generates a new search id for each [user id](#total-users-and-user-ids) after 3 seconds of user inactivity. Multiple queries may be associated with a single search id.
```
gmourier, line 134: The old info I have on it is `5` seconds. Need double confirmation 🤓

learn/cloud/analytics.mdx
```diff
@@ -0,0 +1,197 @@
...
+- `eventName`: an arbitrary string describing the event
+- `indexUid`: the document's index
+- `objectID`: the document's primary key
+- `position`: the document's position in the list of search results
+
+## Understanding the analytics interface
+
+### Total users and user ids
+
+Indicates the total number of user ids. User ids are set with the `X-MS-USER-ID` HTTP header. If `X-MS-USER-ID` is not present, user ids are generated based on `User-Agent` and `CF-Connecting-IP`.
```
gmourier, line 130: In case you want to add a ref to the `CF-Connecting-IP` mention @guimachiavelli

https://developers.cloudflare.com/fundamentals/reference/http-request-headers/#cf-connecting-ip

learn/cloud/analytics.mdx
```diff
@@ -0,0 +1,197 @@
...
+  }'
+```
+
+It is not possible to associate multiple `conversion` events with the same search.
+
+#### The `conversion` event object
+
+The `conversion` event must deliver an object with the following fields:
+
+```js
```
gmourier, line 110:
```suggestion
```json
```

learn/cloud/analytics.mdx
```diff
@@ -0,0 +1,197 @@
...
+
+Conversion rate indicates the number of searches that resulted in a `conversion` event divided by the total number of searches.
+
+To track conversion rates, [configure your application to send `conversion` events to Meilisearch Cloud](#configuring-conversion-rate).
+
+### Click-through rate and average click position
+
+Click-through rate is the number of times users have clicked on a search result divided by the total number of searches.
+
+Average click position indicates the average position of the search results users clicked on, with 0 indicating the first search result.
```
gmourier, line 158: Is it `0` or `1` for the first position? Cross-linking @macraig message [here](https://meilisearch.slack.com/archives/C060MLU0X44/p1700050365021069?thread_ts=1700048907.819179&cid=C060MLU0X44)

learn/cloud/analytics.mdx
```diff
@@ -0,0 +1,197 @@
...
+sidebarDepth: 3
+---
+
+# Analytics
+
+Enable Meilisearch Cloud analytics to gain insight on what your users are searching for, how often they find it, and how much they have to wait between typing a query and receiving results.
+
+## Requirements
+
+You must have a Meilisearch Cloud account to have access to search analytics.
```
Strift, line 13: The requirements seem like the appropriate section to create an account if you don't have one.

I would add a link to `https://meilisearch.com/cloud?utm_campaign=oss&utm_source=docs&utm_medium=analytics` here.

learn/cloud/analytics.mdx
```diff
@@ -0,0 +1,197 @@
...
+
+## Activating analytics
+
+### Enable analytics in the project overview
+
+Log into your Meilisearch Cloud account and navigate to your project's overview. Find the "Analytics" section and click on the "Enable analytics" button:
```
Strift, line 19: I generally use italics when quoting the UI copy, and I thought that was the practice. Just happy to know more about this / agree on something so we can edit the editing guidelines

learn/cloud/analytics.mdx
```diff
@@ -0,0 +1,197 @@
...
+
+![The analytics section of the project overview. It shows one button, "Enable analytics", and a short explanation of the feature.](https://raw.githubusercontent.com/meilisearch/documentation/cloud-analytics-guide/assets/images/cloud-analytics/01-analytics-enable.png)
+
+Meilisearch Cloud will begin processing your request. The "Analytics" section will update when the feature is enabled.
+
+### Update URLs in your application
+
+Enabling analytics generates a new URL for your project. Meilisearch Cloud is only able to track metrics for queries sent to this URL.
```
Strift, line 27: Qogita has reported that "a new URL for your project" can be confusing. Should we say that it's a new host for the meilisearch instance or something like this?

learn/cloud/analytics.mdx
```diff
@@ -0,0 +1,197 @@
...
+curl \
+  -X POST 'http://edge.meilisearch.com/indexes/products/search' \
+  -H 'Content-Type: application/json' \
+  --data-binary '{ "q": "green socks" }'
+```
+
+Your old project URL will remain functional, but API calls using the old URL will not be reflected in the analytics interface.
+
+<Capsule intent="note" title="Analytics and latency">
+To capture and process search analytics data, Meilisearch Cloud must route API requests through a proxy. This might increase latency in some applications.
```
Strift, line 41: This currently gives me more of a "warning, you might not want to do that" feel, where in reality it's not that big a deal for most users. Do you think we could be reassuring/informative here by giving numbers?

afaik the latency only increase by 10 to 25 ms (@brunoocasali @qdequele please correct me if I'm wrong)

learn/cloud/analytics.mdx
```diff
@@ -0,0 +1,197 @@
...
+</Capsule>
+
+### Configuring click-through rate and average click position
+
+To track metrics like clickthrough rate and average click position, Meilisearch Cloud needs to know when users click on search results.
```
Strift, line 46: clickthrough is spelled differently here than in the title above.

"click-through" seems to be the spelling Grammarly prefers

learn/cloud/analytics.mdx
```diff
@@ -0,0 +1,197 @@
...
+
+Every time a user clicks on a search result, your application must send a `click` event to the `POST` endpoint of Meilisearch Cloud analytics route:
+
+```sh
+curl \
+  -X POST 'http://edge.meilisearch.com' \
+  -H 'Content-Type: application/json' \
+  -H 'Authorization: Bearer Meilisearch_API_Key'
```
Strift, line 54: i didn't mention it in my temp docs but i think we should make it clear which type of token we're using, ie. "MEILISEARCH_READ_ONLY_TOKEN" or "DEFAULT_SEARCH_API_KEY" 

i find that reassuring/clarifying but might just be me

learn/cloud/analytics.mdx
```diff
@@ -0,0 +1,197 @@
...
+  "eventType": "click",
+  "eventName": "search result click",
+  "indexUid": "products",
+  "objectID": "0",
+  "position": 0
+}
+```
+
+- `eventType`: a string indicating this is a `click` event
+- `eventName`: a string describing the event
```
Strift, line 79: for a proper analytics pipeline it's recommended to be very strict in the way you name events.

we might want to give users some notice here (especially since we don't parse/validate any data 😂)

the best practice in the industry (amplitude, segment, algolia) seems to be to title-case it and use the past tense ie. "Search Result Clicked"

learn/cloud/analytics.mdx
```diff
@@ -0,0 +1,197 @@
...
+- `indexUid`: a string indicating the clicked document's index
+- `objectID`: a string indicating the clicked document's primary key
+- `position`: an integer indicating the clicked document's position in the list of search results
+
+### Configuring conversion rate
+
+To track conversion rate, first identify what should count as a conversion for your application. For example, in a web shop, a conversion might be a user finalizing the checkout process.
```
Strift, line 86: there can be multiple conversion events. using the webshop example we can give two examples: 'add to card' and 'checkout'

learn/cloud/analytics.mdx
```diff
@@ -0,0 +1,197 @@
...
+Once you have established what counts as a conversion in your application, configure it to send a `conversion` event to the `POST` endpoint of Meilisearch Cloud analytics route:
+
+```sh
+curl \
+  -X POST 'http://edge.meilisearch.com' \
+  -H 'Content-Type: application/json' \
+  -H 'Authorization: Bearer Meilisearch_API_Key'
+  --data-binary '{ 
+    "eventType": "conversion",
+    "eventName": "product added to cart",
```
Strift, line 97: use the same convention as above for event names, if you agree

learn/cloud/analytics.mdx
```diff
@@ -0,0 +1,197 @@
...
+  "eventType": "conversion",
+  "eventName": "product added to cart",
+  "indexUid": "products",
+  "objectID": "0",
+  "position": 0
+}
+```
+
+- `eventType`: indicates this is a `conversion` event
+- `eventName`: an arbitrary string describing the event
```
Strift, line 121: i would not use "arbitrary" to, on the contrary, encourage best practices for naming events & analyzing them

learn/cloud/analytics.mdx
```diff
@@ -0,0 +1,197 @@
...
+- `indexUid`: the document's index
+- `objectID`: the document's primary key
+- `position`: the document's position in the list of search results
+
+## Understanding the analytics interface
```
Strift, line 126: can we opt for something more optimized for in-docs search and google rankings?

top of my head I have "Search analytics metrics", but there might be other viable alternatives

learn/cloud/analytics.mdx
Strift: We changed the lowest value to 1 in [this pr](https://github.com/meilisearch/meilisearch-cloud/pull/3852) but as of Nov 16 12PM CET it's not deployed yet

Strift (commented):
Love the guide :) It's very complete and insightful. I made suggestions regarding some minor things.

Side note: in the future, we might want to start separating feature listing (ie. the list of metrics) from implementation guide. but i don't think it would fit the current docs, and it would be bad if we don't generalize this practice.
I'm saying this because I think it can be better for both in-app search and SEO. But it's a topic for later :)

learn/cloud/analytics.mdx
```diff
@@ -0,0 +1,197 @@
+---
+title: Analytics
+description: Meilisearch Cloud offers in-depth search analytics to help you understand how users search in your application.
+sidebarDepth: 3
+---
+
+# Analytics
+
+Enable Meilisearch Cloud analytics to gain insight on what your users are searching for, how often they find it, and how much they have to wait between typing a query and receiving results.
```
brunoocasali, line 9: > how much they have to wait between typing a query and receiving results

I don't think we give that info 🤔

learn/cloud/analytics.mdx
brunoocasali: If we are very strict, the naming URL is correct, but I can understand why they misunderstood. What about something like this:

Enabling analytics requires an update in your Meilisearch integration's implementation. Meilisearch Cloud is only able to track metrics for queries sent to this URL. Instead of using the old URL like `ms-xxxxxx-123.sfo.meilisearch.com`, for example, you'll have to switch to `edge.meilisearch.com`.

learn/cloud/analytics.mdx
```diff
@@ -0,0 +1,197 @@
...
+Update your application so all API requests point to the new URL:
+
+```sh
+curl \
+  -X POST 'http://edge.meilisearch.com/indexes/products/search' \
+  -H 'Content-Type: application/json' \
+  --data-binary '{ "q": "green socks" }'
+```
+
+Your old project URL will remain functional, but API calls using the old URL will not be reflected in the analytics interface.
```
brunoocasali, line 38: Maybe add that as an alert!?

learn/cloud/analytics.mdx
brunoocasali: This does feel like the right thing to do to me @Strift. Because we usually mention type-as-you-search, search responses under 50ms and other topics like this one. So it may not seem that much (for you and me), but since we have that kind of statement, we need to be clear in my POV that this has some drawbacks (but also, a lot of benefits).

learn/cloud/analytics.mdx
brunoocasali: I would add when and where (since the position of the search result matters in the final number)

learn/cloud/analytics.mdx
brunoocasali: Just another info on the same topic, any token is acceptable, and it should work.

learn/cloud/analytics.mdx
brunoocasali: Still 5 seconds https://github.com/meilisearch/meilisearch-edge/blob/main/src/search.js#L84C47-L84C51

learn/cloud/analytics.mdx
```diff
@@ -0,0 +1,197 @@
...
+The `click` event must deliver an object with the following fields:
+
+```json
+{
+  "eventType": "click",
+  "eventName": "search result click",
+  "indexUid": "products",
+  "objectID": "0",
+  "position": 0
+}
```
brunoocasali, line 75: @qdequele why don't we also ask for the search query used? Do we already have the connection between which search had a conversion/click?

learn/cloud/analytics.mdx
```diff
@@ -0,0 +1,197 @@
...
+
+### Disable analytics in the project overview
+
+Log into your Meilisearch Cloud account and navigate to your project's overview. Find the "Analytics" section and press the "Disable analytics" button:
+
+![The analytics section of the project overview. It shows one button, "Disable analytics", and a short explanation of the feature.](https://raw.githubusercontent.com/meilisearch/documentation/cloud-analytics-guide/assets/images/cloud-analytics/02-analytics-disable.png)
+
+### Update URLs in your application
+
+Disabling analytics changes your project URL to its previous value. Update your application so all API requests point to the correct URL:
```
brunoocasali, line 186: FYI, the same thing happens when they enable analytics, the edge.meilisearch will still work even after disabling, the disabling only affects the front-end, so the user has no disruption in their service. I don't know if you want to mention that, but it is just to let you know.

The only thing that will never work, is the user trying the edge without enabling the analytics at least once!

learn/cloud/analytics.mdx
qdequele: Yes, we have! But it would be nice to allow the developer to send it through the headers, like the user ID.

qdequele (commented):
Hello, I'm not a reviewer, but here are my two cents on this. I have the impression that this page is a melting pot of various things. It starts with a very brief overview, then moves on to guides on specific features, and then mixes high-level documentation with internal information.

I would at least try to avoid complicating the documentation by removing information about how the users' IDs and search IDs are generated. This is complex, subject to change, and not of interest to a large portion of the customers. However, we could have a dedicated section for those who want to delve deeper into this topic.

Moreover, highlighting that the analytics API may result in slower requests is counterproductive. The latency is not guaranteed, and the additional latency depends on the user's geographical location and distance from the Meilisearch project. If we want users to test it, I would avoid mentioning anything negative upfront. We can address those concerns elsewhere.

learn/cloud/analytics.mdx
Strift: To clarify, I'm not against giving a warning. But I'd love it if we could make it a bit more reassuring by giving estimated latency numbers. This way, the customer has the full picture.

learn/cloud/analytics.mdx
guimachiavelli: This would be a change for a future analytics version, correct?

learn/cloud/analytics.mdx
guimachiavelli: @Strift: thanks for spotting it. "Click-through" is indeed the only correct spelling at the moment. Most compound nouns tend to become a single word with time, but we're not there yet 😞

learn/cloud/analytics.mdx
guimachiavelli: Do we support multiple conversion events, though? How does Meilisearch parse multiple conversion events? @brunoocasali

learn/cloud/analytics.mdx
guimachiavelli: Great, thanks for the info. I'll edit the samples to follow that pattern.

learn/cloud/analytics.mdx
guimachiavelli: I'm thinking this whole section might be better off as a separate document. I think it was in large part responsible for @qdequele's impression the guide is unfocussed and meandering.

guimachiavelli:
Thanks for your feedback, folks. I will have a second draft ready for Tuesday morning.

@qdequele, to address the issues you raised: the focus of this first draft was verifying all information I had was correct and expressed in a technically accurate way, which is why I only requested reviews from one product designer and one engineer. The second draft will focus on sorting out structure and text flow. That said, thanks for your input, as it confirmed my feelings this document is not the place for an overview of the analytics interface.

qdequele:
Thanks for your answer @guimachiavelli! 👍

learn/cloud/analytics.mdx
Strift: Just for the record, we got [another support request](https://app.crisp.chat/website/38a71dd2-729c-4970-b061-3e1db1c1eb20/inbox/session_a22624a3-1f38-4856-93c0-48e3e2459212/) asking

> I've noticed that the hostname on the dashboard is currently set as edge.meilisearch.com. Should I consider changing it from the previous format, which was similar to ms-2918e39a832a-2300.sgp.meilisearch.io? Is there a specific advantage or requirement for switching to the new format?

learn/cloud/analytics.mdx
Strift: I think it would make sense too. My initial idea was to have maybe an "analytics" section in the docs with 2 pages, one being dedicated to the "sending events" kinda?

guimachiavelli:
I have made a few final changes to improve clarity and integrate reviewer suggestions.

Based on the data we have from the support front line via @qdequele and @Strift, I'm expediting the review process for the analytics guide. Technical reviews indicate the content is accurate enough for public release. Going through a second review round would not add as much value as getting this content to users earlier than planned.

I will merge this PR as soon as I get an approval.

learn/cloud/analytics.mdx
```diff
@@ -0,0 +1,145 @@
...
+Every time a user clicks on a search result, your application must send a `click` event to the `POST` endpoint of Meilisearch Cloud analytics route:
+
+```sh
+curl \
+  -X POST 'http://edge.meilisearch.com' \
+  -H 'Content-Type: application/json' \
+  -H 'Authorization: Bearer DEFAULT_SEARCH_API_KEY'
+  --data-binary '{ 
+    "eventType": "click",
+    "eventName": "Search result clicked",
```
Strift, line 53:
```suggestion
    "eventName": "Search Result Clicked",
```

learn/cloud/analytics.mdx
```diff
@@ -0,0 +1,145 @@
...
+Once you have established what counts as a conversion in your application, configure it to send a `conversion` event to the `POST` endpoint of Meilisearch Cloud analytics route:
+
+```sh
+curl \
+  -X POST 'http://edge.meilisearch.com' \
+  -H 'Content-Type: application/json' \
+  -H 'Authorization: Bearer Meilisearch_API_Key'
+  --data-binary '{ 
+    "eventType": "conversion",
+    "eventName": "Product Added to Cart",
```
Strift, line 93:
```suggestion
    "eventName": "Product Added To Cart",
```

Strift: that's not exactly "title case" sorry, but that's the convention i was referring to

learn/cloud/analytics.mdx
```diff
@@ -0,0 +1,145 @@
...
+    "indexUid": "products",
+    "objectID": "0",
+    "position": 0
+  }'
+```
+
+It is not possible to associate multiple `conversion` events with the same search.
```
Strift, line 100: that's weird. the way I see it, a given search could have "added to card" and then later "purchased". Or is it a misunderstanding on my end? 🤔

that being said if Bruno / Quentin agreed on the technical correctness of this, it's outside of the scope of this PR

learn/cloud/analytics.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+#### The `conversion` event object
+
+The `conversion` event must deliver an object with the following fields:
+
+```json
+{
+  "eventType": "conversion",
+  "eventName": "Product added to cart",
```
Strift, line 109:
```suggestion
  "eventName": "Product Added To Cart",
```

Strift (commented):
I added a small comment about something that could be outside the scope of the PR.

I suggested the casing changes to event names as it seemed easy to integrate. Feel free to postpone them if it messes up your workflow.

Seems good for me any way, we can patch update the rest within the day. At least we'll have a resource to send customers 👌

learn/cloud/analytics.mdx
Strift: Ok, I verified with Quentin and here's what's happening

- you can send multiple conversion events (with different event names), and the database will store them
- the current UI of the Cloud, indeed, is not able to display the per event stats, so it would be confusing

so from my understanding the info is not exactly "correct", but it's probably the right advise to give our users

Strift (approved):
LGTM ✨

PR #2632, opened 2023-11-21
Fix analytics guide asset links

guimachiavelli:
Fixes broken image links on Meilisearch Cloud analytics guide.


Strift (commented):
I do see it in the vercel preview, but then again I was seeing it too earlier sooo 😅

Strift (approved):

PR #2638, opened 2023-11-23
Add information about reserved attributes

### LINKED ISSUE #2498: Warn users about attributes prefixed with `_`
When creating new features, Meilisearch sometimes must reserve document fields. For example, when we introduced geosearch, `_geo` became a reserved attribute.

This might create problems for users whose datasets include attributes prefixed with `_`. Through no fault of their own, a user might have to delay upgrading Meilisearch and/or dedicating time and effort to adjust their dataset if Meilisearch reserves an attribute present in their datset..

To minimise the chances of this happening, we should mention Meilisearch periodically introduces new reserved attributes, and that these attributes are always prefixed with `_`.

## Tasks
- [ ] Update `/learn/core_concepts/documents.mdx`

guimachiavelli:
Closes #2498

### DIFF
--- a/learn/core_concepts/documents.mdx
+++ b/learn/core_concepts/documents.mdx
@@ -26,13 +26,21 @@ Nothing will be shared between two indexes if they contain the exact same docume
 
 A field is a set of two data items linked together: an attribute and a value. Documents are made up of fields.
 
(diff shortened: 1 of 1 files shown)


irevoire (approved):
lgtm

PR #2680, opened 2024-01-11
Meilisearch Cloud: Include documentation for custom user bindings

fabio-nettis:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/4309

## What does this PR do?

This PR includes documentation for the `X-MS-USER-ID` header. This header allows to manually set the user id associated with a search and subsequent requests to the `/event` endpoint. Based on the [feedback of this issue](https://github.com/meilisearch/meilisearch/issues/4302#issuecomment-1885274716).

## PR checklist
Please check if your PR fulfills the following requirements:
- [x] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [x] Have you read the contributing guidelines?
- [x] Have you made sure that the title is accurate and descriptive of the changes?

Thank you so much for contributing to Meilisearch!

PR #2689, opened 2024-01-17
Fix faulty #2680 merge

guimachiavelli:
For some reason #2680 did not reflect my suggestions, despite the user seeming to have accepted them. This PR fixes that by pushing the final text.

### DIFF
--- a/learn/cloud/analytics.mdx
+++ b/learn/cloud/analytics.mdx
@@ -38,9 +38,7 @@ curl \
 **The previous API URL will remain functional**, but requests targeting it will not send any data to the analytics interface. If you created any custom API keys using the previous URL, you will need to replace them.
 
 <Capsule intent="danger" title="Analytics and multi-search queries">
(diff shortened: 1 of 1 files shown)


curquiza:
@guimachiavelli 
You can see on the PR no commit has been pushed following your suggestions -> it means the suggestions were not applied

curquiza (approved):
tutto bene

guimachiavelli:
huh, you're right. I guess I just assumed the contributor had accepted my suggestions because they had marked my suggestions as resolved without any comment?

curquiza:
> huh, you're right. I guess I just assumed the contributor had accepted my suggestions because they had marked my suggestions as resolved without any comment?

Most of the contributors don't know how to use commit suggestions. I think they thought the suggestions would be applied if they click on "Resolve conversation"
When you do commit suggestions like this, don't wait for their approval -> you are the maintainer, you can apply them directly and then, merge the PR

PR #2716, opened 2024-02-14
Update security guide

guimachiavelli:


learn/security/managing_api_keys.mdx
```diff
@@ -0,0 +1,165 @@
...
+  "createdAt": "2022-01-01T10:00:00Z",
+  "updatedAt": "2022-01-01T10:00:00Z"
+}
+```
+
+To update an API key, you must use the [update API key endpoint](/reference/api/keys#update-a-key), which can only be accessed with the master key or an API key with the `keys.update` action.
+
+Meilisearch supports partial updates with the `PATCH` route. This means your payload only needs to contain the data you want to update—in this case, `description`.
+
+## Creating an API key
```
macraig, line 44: Is there a reason for putting the Update section before the Creation section?

learn/security/managing_api_keys.mdx
macraig, line 1: There's several mentions of the different kinds of key actions like keys.get, keys.create, etc as they're an essential part of how keys work. We could consider adding a reference link for users that aren't familiar with how key permissions work. Alternatively, we assume users reading this guide have technical understanding on the subject, in which case you can completely ignore this comment :)

learn/security/basic_security.mdx
```diff
@@ -0,0 +1,160 @@
...
+```sh
+set MEILI_MASTER_KEY="MASTER_KEY"
+./meilisearch
+```
+
+</Tabs.Content>
+</Tabs.Container>
+
+<Capsule intent="tip" title="Tools for generating a master key">
+The master key must be at least 16-bytes-long and composed of valid UTF-8 characters. Use one of the following tools to generate a secure master key:
```
CaroFG, line 55: We could mention that when launching a self-hosted instance of Meilisearch without a master key, Meilisearch suggests an autogenerated one:

```
We generated a new secure master key for you (you can safely use this token):

>> --master-key qOauLmgeXv7AyYVXQwinXc0riFBJ8XRzN_ATndt6XdE <<
```

learn/security/managing_api_keys.mdx
```diff
@@ -0,0 +1,165 @@
...
+  ],
+  "expiresAt": "2023-01-01T00:00:00Z",
+  "createdAt": "2022-01-01T10:00:00Z",
+  "updatedAt": "2022-01-01T10:00:00Z"
+}
+```
+
+It is good practice to always set an expiry date when creating a new API key. If you are sure this is not necessary in your application, you can create an API key with no expiry date by explicitly passing a `null` value to `expiresAt`.
+
+## Listing API keys
```
CaroFG, line 74: Shall we remind that there are 2 default API keys?

learn/security/managing_api_keys.mdx
CaroFG: I also find the order a little bit disturbing, I'd start with listing, then creating, updating and deleting

config/sidebar-learn.json
```diff
@@ -202,9 +202,29 @@
...
+        "label": "Managing API keys",
+        "slug": "managing_api_keys"
+      },
+      {
+        "source": "learn/security/resetting_master_key.mdx",
+        "label": "Resetting the master key",
+        "slug": "resetting_master_key"
+      },
+      {
+        "source": "learn/security/differences_master_api_keys.mdx",
```
CaroFG, line 220: I'd move this to be the second element on the list menu

learn/security/managing_api_keys.mdx
guimachiavelli: I'm not sure what was the reasoning behind the order here. It might have been because we already start with two API keys, so creation is not necessarily the first thing a user will do? In any case, I was trying to avoid opening the can of worms that is this page (my controversial take is we need to delete this page and write a new guide).

Since we have two votes for changing the order, though, I'll follow the order suggested by @CaroFG. It's also the order we have in the reference.

learn/security/basic_security.mdx
guimachiavelli: I considered doing that, but since the idea is that users will follow each step from beginning to end, it felt weird to tell them something that only happens if they deviate from the tutorial.

learn/security/managing_api_keys.mdx
guimachiavelli: Good idea, added a few links to the actions table.

learn/security/basic_security.mdx
CaroFG: But it would make people's lives so much easier

macraig (approved):

PR #2744, opened 2024-02-28
v1.7: Updates to OpenAI embedder

### LINKED ISSUE #2734: v1.7: Updates to OpenAI embedder
v1.7 introduces two minor changes to Meilisearch's OpenAI embedder.

1. Two new supported models: `text-embedding-3-small` and `text-embedding-3-large`
2. Users can set the `dimensions` parameter for the OpenAI embedder (`dimensions` must be >0 and lower than the model size)

## Questions
- Why would a user manually set the dimension? What is the benefit of setting 100 dimensions when the model was built with 1500?

## Tasks
- [ ] Update OpenAI instructions in `/experimental/vector_search`

## References
- SME: @dureuill 
- GH discussion: https://github.com/orgs/meilisearch/discussions/677#discussioncomment-8306384
- engine issue: https://github.com/meilisearch/meilisearch/issues/4394

guimachiavelli:
Closes #2734

### DIFF
--- a/learn/experimental/vector_search.mdx
+++ b/learn/experimental/vector_search.mdx
@@ -65,7 +65,7 @@ It is mandatory to pass an OpenAI API key through the `OPENAI_API_KEY` environme
 
 `documentTemplate` is an optional field. Use it to customize the data you send to the embedder. It is highly recommended you configure a custom template for your documents.
 
-`dimensions` is an optional field. It must be bigger than zero and lower or equal to the chosen model's dimensions. This option is not compatible with `openai-text-embedding-ada-002`. Fewer dimensions may improve performance, but lead to lower search result accuracy.
+`dimensions` is an optional field. It must be bigger than zero and lower or equal to the chosen model's dimensions. This option is not compatible with `text-embedding-ada-002`. Fewer dimensions may improve performance, but lead to lower search result accuracy.
 
 <Capsule intent="tip" title="`documentTemplate` usage">
 `documentTemplate` must be a [Liquid template](https://shopify.github.io/liquid/). Use `{{ doc.attribute }}` to access the `attribute` field value of your documents. Any field you refer to in this way must exist in all documents or an error will be raised at indexing time.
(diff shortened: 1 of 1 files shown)


learn/experimental/vector_search.mdx
```diff
@@ -51,16 +51,21 @@ curl \
...
 
-`documentTemplate` is an optional field you can use to customize the data you send to the embedder. It is highly recommended you configure a custom template for your documents.
+`model` is a mandatory field indicating a compatible OpenAI model. Meilisearch supports the following [OpenAI models](https://platform.openai.com/docs/models/embeddings): `openai-text-embedding-ada-002`, `text-embedding-3-small`, and `text-embedding-3-large`. 
+
+`documentTemplate` is an optional field. Use it to customize the data you send to the embedder. It is highly recommended you configure a custom template for your documents.
+
+`dimensions` is an optional field. It must be bigger than zero and lower or equal to the chosen model's dimensions. This option is not compatible with `openai-text-embedding-ada-002`. Fewer dimensions may improve performance, but lead to lower search result accuracy.
```
dureuill, line 68: > This option is not compatible with `openai-text-embedding-ada-002`.

Technically, if you use it with this model, Meilisearch will do the right thing as long as you pass the exact natural dimensions of this model (1536).

It is OK to simplify like in the text.

Also, typo on model name:
```suggestion
`dimensions` is an optional field. It must be bigger than zero and lower or equal to the chosen model's dimensions. This option is not compatible with `text-embedding-ada-002`. Fewer dimensions may improve performance, but lead to lower search result accuracy.
```

dureuill (approved):
Approved but see inline comment

learn/experimental/vector_search.mdx
guimachiavelli: Yes, I was aware of the fact `dimensions` would work with `ada-002` if the user were to specify the exact model size. I chose to simplify it to prevent giving users a false choice: "yes, you can specify this setting, but it won't work unless you use the default value."

Thanks for spotting the typo!

PR #2746, opened 2024-02-28
v1.7: Add Hugging Face vector search updates

### LINKED ISSUE #2736: v1.7: Huggingface embedder updates
v1.7 allows users to generate embeddings with NVidia GPUs.

## Questions
- Is this being exposed on Meilisearch Cloud? If so, will it require any user input or will it be sorted out on our end?

## Tasks
- [ ] Update `learn/experimental/vector_search`

## References
- SME: @dureuill 
- Initial documentation: https://github.com/meilisearch/meilisearch/issues/4306#issuecomment-1893849620
- Engine issue: https://github.com/meilisearch/meilisearch/issues/4306

guimachiavelli:
Closes #2736 

@Strift and @CaroFG, notifying you this PR includes a short guide aimed at enabling GPU usage with the Hugging Face embedding. It's aimed at advanced self-hosted users, but thought you might want to be aware of it.

### DIFF
--- a/learn/cookbooks/computing_hugging_face_embeddings_gpu.mdx
+++ b/learn/cookbooks/computing_hugging_face_embeddings_gpu.mdx
@@ -1,0 +1,72 @@
+---
+title: Computing Hugging Face embeddings with the GPU — Meilisearch documentation
+description: This guide for experienced users shows you how to compile a Meilisearch binary that generates Hugging Face embeddings with an Nvidia GPU.
+---
+
+This guide for experienced users shows you how to compile a Meilisearch binary that generates Hugging Face embeddings with an Nvidia GPU.
+
+## Prerequisites
+
+- A [CUDA-compatible Linux distribution](https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#id12)
+- Nvidia GPU with CUDA support
+- A modern Rust compiler
+
+## Install CUDA
+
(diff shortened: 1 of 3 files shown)


learn/cookbooks/computing_hugging_face_embeddings_gpu.mdx
```diff
@@ -0,0 +1,90 @@
...
+After you have installed CUDA in your machine, run the following command in your command-line terminal:
+
+```sh
+nvcc --version | head -1
+```
+
+If CUDA is working correctly, you will see the following response:
+
+```
+nvcc: NVIDIA (R) Cuda compiler driver
```
dureuill, line 29: I think this one might be cut short, was it not?

dureuill (approved):
:+1:

Strift:
@guimachiavelli the rest of docs isn't enough for you anymore you have to write in our cookbooks now 😒

(thanks for the ping)

learn/cookbooks/computing_hugging_face_embeddings_gpu.mdx
guimachiavelli: The code sample pipes the result into `head -1` so we only get the first line. As far as I understand, we only care about receiving a positive reply here stating cuda was correctly installed, so I thought using `head` would be simpler and more to the point.

I can change it if you think it's better to have the full output, though.

dureuill: OK I didn't see the head -1, that makes sense then

It doesn't need to be changed

PR #2749, opened 2024-02-29
v1.7: Ranking score details

### LINKED ISSUE #2735: v1.7: Stabilize `scoreDetails`
v1.7 stabilizes `scoreRankingDetails`. The feature is no longer gated behind an experimental feature. To activate it, users must set the search parameter `showRankingScoreDetails` to `true`.

## Questions
- what happens if a user sets `showRankingScore` to `false` and `showRankingScoreDetails` to `true`?

## Tasks
- [ ] Remove `/learn/experimental/ranking_rule_score_details`
- [ ] Set up redirect for removed page pointing to search parameters page
- [ ] Update `/reference/api/search`

## References
- SME: @dureuill 
- engine issue: https://github.com/meilisearch/meilisearch/issues/4359

guimachiavelli:
Closes #2735


dureuill (approved):
Perfect 👌

.code-samples.meilisearch.yaml
```diff
@@ -1064,7 +1064,7 @@ update_experimental_features_1: |-
     -X PATCH 'http://localhost:7700/experimental-features/' \
     -H 'Content-Type: application/json'  \
     --data-binary '{
-      "scoreDetails": true
+      "exportPuffinReports": true
```
curquiza, line 1067: Could we use `vectorStore` instead which is a way more interesting feature 😇

guimachiavelli: As I mentioned in the postman PR, I sort of wanted to avoid emphasizing the connection of vector search with experimental features. My reasoning is that we often mention experimental features are unstable and recommend against using them in production environments, which is at odds with the rest of our communication being very much recommending people to use vector search as much as possible.

Happy to change this if you don't think it's a problem.

curquiza: Use `metrics` instead in this case. `exportPuffinReports` has almost zero feature interest for a classic user. They will use it when engine team will ask to use it in issue reports

guimachiavelli, line 1067:
```suggestion
      "metrics": true
```

curquiza (approved):

PR #2750, opened 2024-02-29
v1.7: Update Postman collection

### LINKED ISSUE #2739: v1.7: Update Postman collection
As title.

guimachiavelli:
Closes #2739

### DIFF
--- a/assets/misc/meilisearch-collection-postman.json
+++ b/assets/misc/meilisearch-collection-postman.json
@@ -2624,7 +2624,7 @@
 						"header": [],
 						"body": {
 							"mode": "raw",
-							"raw": "{\n    \"exportPuffinReports\": true\n}",
+							"raw": "{\n    \"metrics\": true\n}",
 							"options": {
 								"raw": {
 									"language": "json"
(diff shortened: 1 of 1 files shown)


curquiza (changes requested):
cf my comment [here](https://github.com/meilisearch/documentation/issues/2739#issuecomment-1974876411) 🤓

assets/misc/meilisearch-collection-postman.json
```diff
@@ -2618,7 +2624,7 @@
 						"header": [],
 						"body": {
 							"mode": "raw",
-							"raw": "{\n    \"scoreDetails\": true\n}",
+							"raw": "{\n    \"exportPuffinReports\": true\n}",
```
curquiza, line 2627: Let's use a more shiny feature like AI instead 😇

guimachiavelli: I sort of didn't want to have vector search here to avoid continuously associating it with experimental/unstable features.

curquiza: In this case, use `metrics` instead. It's way more useful for the users, which is not the case of `exportPuffinReports`

PR #2755, opened 2024-03-07
v1.7: Vector search Cloud waitlist

guimachiavelli:
As requested by @macraig, this PR adds a warning notifying Cloud users must enter the waitlist before activating vector search on Meilisearch Cloud.

### DIFF
--- a/learn/experimental/vector_search.mdx
+++ b/learn/experimental/vector_search.mdx
@@ -19,11 +19,7 @@ If using Meilisearch Cloud, navigate to your project overview and find "Experime
 
 ![A section of the project overview interface titled "Experimental features". There are two options: "Score details" and "Vector store". "Vector store" is turned on.](https://raw.githubusercontent.com/meilisearch/documentation/main/assets/images/vector-search/01-cloud-vector-store.png)
 
-<Capsule intent="note" title="Waiting list">
-To ensure proper onboarding and scaling of Meilisearch Cloud's latest AI-powered search offering, you must enter the waitlist before activating vector search.
-</Capsule>
-
-Alternatively, use [the `/experimental` route](/reference/api/experimental_features) to activate vector search during runtime:
+Alternatively, use [the `/experimental-features` route](/reference/api/experimental_features) to activate vector search during runtime:
 
 ```sh
 curl \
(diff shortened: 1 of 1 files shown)


learn/experimental/vector_search.mdx
```diff
@@ -19,6 +19,10 @@ If using Meilisearch Cloud, navigate to your project overview and find "Experime
 
 ![A section of the project overview interface titled "Experimental features". There are two options: "Score details" and "Vector store". "Vector store" is turned on.](https://raw.githubusercontent.com/meilisearch/documentation/main/assets/images/vector-search/01-cloud-vector-store.png)
 
+<Capsule intent="note" title="Waiting list">
+To ensure proper onboarding and scaling of Meilisearch Cloud's latest AI-powered search offering, you must enter the waitlist before activating vector search.
```
macraig, line 23: This is only for v1.7 onwards, previous versions from v1.3 to v1.6 don't have any embedder settings so you can just enable the experimental feature to use the vector store

learn/experimental/vector_search.mdx
```diff
@@ -19,6 +19,10 @@ If using Meilisearch Cloud, navigate to your project overview and find "Experime
...
+</Capsule>
+
 Alternatively, use [the `/experimental` route](/reference/api/experimental_features) to activate vector search during runtime:
```
macraig, line 26: This applies up to v1.6 included

macraig (changes requested):
Added a few comments regarding versioning, hopefully it's not too confusing

learn/experimental/vector_search.mdx
guimachiavelli: Does that mean the experimental features route will be unavailable for v1.7?

learn/experimental/vector_search.mdx
guimachiavelli: If I understand your comment correctly, I don't think there's much to change here. We only document the latest Meilisearch release, so describing the behaviour of previous releases is sort of out of scope.

Also is this tied to the engine version or to the Meilisearch Cloud version? We added embedder options on v1.6.

guimachiavelli: Discussed this over slack and decided to keep it as is.

macraig (approved):

PR #2757, opened 2024-03-13
Add LLM synonyms

Strift:
Add LLM synonyms in docs scraper config


guimachiavelli (approved):

CaroFG (changes requested):
If we want "llm" to return "large language model" I think we should create a mutual association and not just a one-way one

PR #2810, opened 2024-04-18
v1.8: Search request queue size

### LINKED ISSUE #2797: v1.8: Limit number of concurrent search requests
v1.8 introduces a  limit of 1000 concurrent search requests. The main goal is to protect instances from DoS-style attacks.

Although the limit itself is active by default in all installs, it is possible to configure it with an experimental instance option, `--experimental-search-queue-size`.

## Tasks
- [ ] Create new experimental page
- [ ] Update experimental overview with new feature
- [ ] Update "Known limitations" page

## References
- engine issue: https://github.com/meilisearch/meilisearch/issues/4489
- PRD: https://meilisearch.notion.site/API-usage-of-the-search-limiter-c1edc092f3c746d6899c5ba7bfa9a6a8
- SME: @irevoire

guimachiavelli:
Closes #2797

### DIFF
--- a/config/sidebar-learn.json
+++ b/config/sidebar-learn.json
@@ -537,6 +537,11 @@
         "source": "learn/experimental/replication_parameters.mdx",
         "label": "Replication parameters",
         "slug": "replication_parameters"
+      },
+      {
+        "source": "learn/experimental/search_queue_size.mdx",
+        "label": "Search queue size",
+        "slug": "search_queue_size"
       }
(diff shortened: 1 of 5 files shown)


ManyTheFish (commented):
We may explain that if the search queue is full, then a random search in the queue will be dropped, and will return too_many_search_requests, returned with an HTTP 503 error, and the Retry-After header set to 10s

PR #2824, opened 2024-05-16
Vector search reference

guimachiavelli:
This PR adds an initial base reference for:

- `hybrid` and `vector` search parameters
- `embedders` index setting

### DIFF
--- a/reference/api/search.mdx
+++ b/reference/api/search.mdx
@@ -1077,7 +1077,7 @@ Configures Meilisearch to return search results based on a query's meaning and c
 
 `hybrid` must be an object. It accepts two fields: `embedder` and `semanticRatio`.
 
-`embedder` must be a string indicating an embedder configured with the `/settings` endpoint. Defaults to `"default"`.
+`embedder` must be a string indicating an embedder configured with the `/settings` endpoint. If you don't specify an embedder and your index contains a single embedder, Meilisearch uses it by default. If an index contains multiple embedders, Meilisearch will use the embedder named `default`.
 
 `semanticRatio` must be a number between `0.0` and `1.0` indicating the proportion between keyword and semantic search results. `0.0` causes Meilisearch to only return keyword results. `1.0` causes Meilisearch to only return meaning-based results. Defaults to `0.5`.
 
--- a/reference/api/search.mdx
+++ b/reference/api/search.mdx
@@ -54,6 +54,8 @@ By default, [this endpoint returns a maximum of 1000 results](/learn/advanced/kn
 | **[`matchingStrategy`](#matching-strategy)**            | String           | `last`        | Strategy used to match query terms within documents |
 | **[`showRankingScore`](#ranking-score)**                | Boolean          | `false`       | Display the global ranking score of a document      |
 | **[`attributesToSearchOn`](#customize-attributes-to-search-on-at-search-time)**            | Array of strings           | `["*"]`        | Restrict search to the specified attributes |
+| **[`hybrid`](#hybrid-search)**                | Object          | `null`       | Return results based on query keywords and meaning      |
+| **[`vector`](#vector)**                | Array of numbers          | `null`       | Search using a custom query vector      |
 
 [Learn more about how to use each search parameter](#search-parameters).
 
@@ -162,6 +164,8 @@ By default, [this endpoint returns a maximum of 1000 results](/learn/advanced/kn
 | **[`matchingStrategy`](#matching-strategy)**            | String           | `last`        | Strategy used to match query terms within documents |
 | **[`showRankingScore`](#ranking-score)**                | Boolean          | `false`       | Display the global ranking score of a document      |
 | **[`attributesToSearchOn`](#customize-attributes-to-search-on-at-search-time)**            | Array of strings           | `["*"]`        | Restrict search to the specified attributes |
+| **[`hybrid`](#hybrid-search)**                | Object          | `null`       | Return results based on query keywords and meaning      |
+| **[`vector`](#vector)**                | Array of numbers          | `null`       | Search using a custom query vector      |
 
 [Learn more about how to use each search parameter](#search-parameters).
 
@@ -248,6 +252,8 @@ This is not necessary when using the `POST` route or one of our [SDKs](/learn/wh
 | **[`matchingStrategy`](#matching-strategy)**            | String           | `last`        | Strategy used to match query terms within documents |
 | **[`showRankingScore`](#ranking-score)**                | Boolean          | `false`       | Display the global ranking score of a document      |
 | **[`attributesToSearchOn`](#customize-attributes-to-search-on-at-search-time)**            | Array of strings           | `["*"]`        | Restrict search to the specified attributes |
+| **[`hybrid`](#hybrid-search)**                | Object          | `null`       | Return results based on query keywords and meaning      |
+| **[`vector`](#vector)**                | Array of numbers          | `null`       | Search using a custom query vector      |
 
 ### Query (q)
 
@@ -1060,3 +1066,51 @@ The following query returns documents whose `overview` includes `"adventure"`:
 <CodeSamples id="search_parameter_guide_attributes_to_search_on_1" />
 
 Results would not include documents containing `"adventure"` in other fields such as `title` or `genre`, even if these fields were present in the `searchableAttributes` list.
+
+### Hybrid search (experimental)
+
+**Parameter**: `hybrid`<br />
+**Expected value**: An object with two fields: `embedder` and `semanticRatio`<br />
+**Default value**: `null`
+
+Configures Meilisearch to return search results based on a query's meaning and context.
+
+`hybrid` must be an object. It accepts two fields: `embedder` and `semanticRatio`.
+
+`embedder` must be a string indicating an embedder configured with the `/settings` endpoint. Defaults to `"default"`.
+
+`semanticRatio` must be a number between `0.0` and `1.0` indicating the proportion between keyword and semantic search results. `0.0` causes Meilisearch to only return keyword results. `1.0` causes Meilisearch to only return meaning-based results. Defaults to `0.5`.
+
+<Capsule intent="warning">
+Meilisearch will return an error if you use `hybrid` before activating your instance's `vectorStore` and [configuring an embedder](/reference/api/settings#embedders-experimental).
+</Capsule>
+
+#### Example
+
+```sh
+curl -X POST 'localhost:7700/indexes/products/search' \
+   -H 'content-type: application/json' \
+  --data-binary '{ 
+    "q": "kitchen utensils",
+    "hybrid": {
+      "semanticRatio": 0.9,
+      "embedder": "default"
+    }
+  }'
+```
+
+### Vector (experimental)
+
+Use a custom-generated vector to perform a search query. Must be an array of numbers corresponding to the dimensions of the custom vector.
+
+#### Example
+
+```sh
+curl -X POST 'localhost:7700/indexes/products/search' \
+  -H 'content-type: application/json' \
+  --data-binary '{ "vector": [0, 1, 2] }'
+```
+
+<Capsule intent="warning">
+Meilisearch will return an error if you use `vector` before activating your instance's `vectorStore` and [configuring a custom embedder](/reference/api/settings#embedders-experimental).
+</Capsule>
--- a/reference/api/search.mdx
+++ b/reference/api/search.mdx
@@ -1087,29 +1087,19 @@ Meilisearch will return an error if you use `hybrid` before activating your inst
 
 #### Example
 
-```sh
-curl -X POST 'localhost:7700/indexes/products/search' \
-   -H 'content-type: application/json' \
-  --data-binary '{ 
-    "q": "kitchen utensils",
-    "hybrid": {
-      "semanticRatio": 0.9,
-      "embedder": "default"
-    }
-  }'
-```
+<CodeSamples id="search_parameter_guide_hybrid_1" />
 
 ### Vector (experimental)
 
-Use a custom-generated vector to perform a search query. Must be an array of numbers corresponding to the dimensions of the custom vector.
+Use a custom vector to perform a search query. Must be an array of numbers corresponding to the dimensions of the custom vector.
+
+`vector` is mandatory when performing searches with `userProvided` embedders. You may also use `vector` to override an embedder's automatic vector generation.
+
+`vector` dimensions must match the dimensions of the embedder.
 
 #### Example
 
-```sh
-curl -X POST 'localhost:7700/indexes/products/search' \
-  -H 'content-type: application/json' \
-  --data-binary '{ "vector": [0, 1, 2] }'
-```
+<CodeSamples id="search_parameter_guide_vector_1" />
 
 <Capsule intent="warning">
 Meilisearch will return an error if you use `vector` before activating your instance's `vectorStore` and [configuring a custom embedder](/reference/api/settings#embedders-experimental).
--- a/reference/api/search.mdx
+++ b/reference/api/search.mdx
@@ -54,8 +54,8 @@ By default, [this endpoint returns a maximum of 1000 results](/learn/advanced/kn
 | **[`matchingStrategy`](#matching-strategy)**            | String           | `last`        | Strategy used to match query terms within documents |
 | **[`showRankingScore`](#ranking-score)**                | Boolean          | `false`       | Display the global ranking score of a document      |
 | **[`attributesToSearchOn`](#customize-attributes-to-search-on-at-search-time)**            | Array of strings           | `["*"]`        | Restrict search to the specified attributes |
-| **[`hybrid`](#hybrid-search)**                | Object          | `null`       | Return results based on query keywords and meaning      |
-| **[`vector`](#vector)**                | Array of numbers          | `null`       | Search using a custom query vector      |
+| **[`hybrid`](#hybrid-search-experimental)**                | Object          | `null`       | Return results based on query keywords and meaning      |
+| **[`vector`](#vector-experimental)**                | Array of numbers          | `null`       | Search using a custom query vector      |
 
 [Learn more about how to use each search parameter](#search-parameters).
 
@@ -164,8 +164,8 @@ By default, [this endpoint returns a maximum of 1000 results](/learn/advanced/kn
 | **[`matchingStrategy`](#matching-strategy)**            | String           | `last`        | Strategy used to match query terms within documents |
 | **[`showRankingScore`](#ranking-score)**                | Boolean          | `false`       | Display the global ranking score of a document      |
 | **[`attributesToSearchOn`](#customize-attributes-to-search-on-at-search-time)**            | Array of strings           | `["*"]`        | Restrict search to the specified attributes |
-| **[`hybrid`](#hybrid-search)**                | Object          | `null`       | Return results based on query keywords and meaning      |
-| **[`vector`](#vector)**                | Array of numbers          | `null`       | Search using a custom query vector      |
+| **[`hybrid`](#hybrid-search-experimental)**                | Object          | `null`       | Return results based on query keywords and meaning      |
+| **[`vector`](#vector-experimental)**                | Array of numbers          | `null`       | Search using a custom query vector      |
 
 [Learn more about how to use each search parameter](#search-parameters).
 
@@ -252,8 +252,8 @@ This is not necessary when using the `POST` route or one of our [SDKs](/learn/wh
 | **[`matchingStrategy`](#matching-strategy)**            | String           | `last`        | Strategy used to match query terms within documents |
 | **[`showRankingScore`](#ranking-score)**                | Boolean          | `false`       | Display the global ranking score of a document      |
 | **[`attributesToSearchOn`](#customize-attributes-to-search-on-at-search-time)**            | Array of strings           | `["*"]`        | Restrict search to the specified attributes |
-| **[`hybrid`](#hybrid-search)**                | Object          | `null`       | Return results based on query keywords and meaning      |
-| **[`vector`](#vector)**                | Array of numbers          | `null`       | Search using a custom query vector      |
+| **[`hybrid`](#hybrid-search-experimental)**                | Object          | `null`       | Return results based on query keywords and meaning      |
+| **[`vector`](#vector-experimental)**                | Array of numbers          | `null`       | Search using a custom query vector      |
 
 ### Query (q)
 
--- a/reference/api/settings.mdx
+++ b/reference/api/settings.mdx
@@ -165,6 +165,7 @@ If the provided index does not exist, it will be created.
 | **[`stopWords`](#stop-words)**                       | Array of strings | Empty                                                                                            | List of words ignored by Meilisearch when present in search queries              |
 | **[`synonyms`](#synonyms)**                          | Object           | Empty                                                                                            | List of associated words treated similarly                                       |
 | **[`typoTolerance`](#typo-tolerance)**               | Object           | [Default object](#typo-tolerance-object)                                                         | Typo tolerance settings                                                          |
+| **[`embedders`](#embedders-experimental)**               | Object of objects           | [Default object](#embedder-object)                                                         | Embedder required for performing meaning-based search queries                                                          |
 
 #### Example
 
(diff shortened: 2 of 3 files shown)


reference/api/search.mdx
```diff
@@ -1060,3 +1066,51 @@ The following query returns documents whose `overview` includes `"adventure"`:
...
+    "hybrid": {
+      "semanticRatio": 0.9,
+      "embedder": "default"
+    }
+  }'
+```
+
+### Vector (experimental)
+
+Use a custom-generated vector to perform a search query. Must be an array of numbers corresponding to the dimensions of the custom vector.
```
dureuill, line 1104: I feel like we're probably missing some explanations here:

1. `vector` is effectively mandatory for a `userProvided` embedder, as auto-embedding won't work for it
2. `vector` can also be used to punctually override the auto-embedding of the query. If vector is used, then no autoembedding will take place.
3. The custom vector's dimensions must match the dimensions of vectors produced by the embedder, or an error will be returned.

reference/api/settings.mdx
```diff
@@ -1910,3 +1911,189 @@ Reset an index's typo tolerance settings to their [default value](#typo-toleranc
...
 
 You can use the returned `taskUid` to get more details on [the status of the task](/reference/api/tasks#get-one-task).
+
+## Embedders (experimental)
+
+Embedders translate documents and queries into vector embeddings. You must configure at least one embedder to use AI-powered search.
+
+### Embedders object
+
+The embedders object may contain an arbitrary number of embedder objects:
```
dureuill, line 1921: should we specify "with arbitrary names"?

BTW the number is not arbitrary, it must be <256. Up to 256 embedders

reference/api/settings.mdx
```diff
@@ -1910,3 +1911,189 @@ Reset an index's typo tolerance settings to their [default value](#typo-toleranc
...
+| **`pathToEmbeddings`** | Array            | []                                                                       | Path to vector embedding data in embedder response                                                                                                             |
+| **`embeddingObject`**  | String           | {}                                                                       | Name of the embedding object in embedder response                                                                                                              |
+| **`distribution`**     | Object           | Empty                                                                    | Describes the natural distribution of search results. Must contain two fields, `mean` and `sigma`, each containing a numeric value between `0` and `1`         |
+
+Depending on your chosen embedder, some of these fields may be mandatory.
```
dureuill, line 1957: Depending on the chosen source, some of the fields may be mandatory, and only some of the fields are allowed.

I understand this may deviate from the usual presentation, but I would expect the reference to list which fields are allowed for each embedder source, and which fields are mandatory for each embedder source. Could be an addition to the table, or could be one table per source (as the exact meaning of parameters may change depending on the source)

reference/api/settings.mdx
```diff
@@ -1910,3 +1911,189 @@ Reset an index's typo tolerance settings to their [default value](#typo-toleranc
...
+}
+```
+
+These embedder objects may contain the following fields:
+
+| Name                   | Type             | Default Value                                                            | Description                                                                                                                                                    |
+|--:---------------------|--:---------------|--:-----------------------------------------------------------------------|--:-------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| **`source`**           | String           | Empty                                                                    | The third-party tool that will generate embeddings from documents. Must be `openAi`, `huggingFace`, `ollama`, `rest`, or `userProvided`                        |
+| **`url`**              | String           | `http://localhost:11434/api/embeddings`                                  | URL Meilisearch contacts when querying the embedder                                                                                                        |
```
dureuill, line 1944: I understand the detailed meaning of the REST embedder parameters (url, inputField, inputType, query, pathToEmbeddings and embeddingObject) will be specified in the later guide on REST embedder.

reference/api/settings.mdx
```diff
@@ -1910,3 +1911,189 @@ Reset an index's typo tolerance settings to their [default value](#typo-toleranc
...
+| **`inputType`**        | String           | `text`                                                                   | Type of data Meilisearch sends to embedder. Must be `text` or `textArray`                                                                                      |
+| **`query`**            | String           | Object                                                                   | Extra fields Meilisearch adds to queries Meilisearch sends to embedder                                                                                         |
+| **`pathToEmbeddings`** | Array            | []                                                                       | Path to vector embedding data in embedder response                                                                                                             |
+| **`embeddingObject`**  | String           | {}                                                                       | Name of the embedding object in embedder response                                                                                                              |
```
dureuill, line 1954: This is an array, not a String. It represents the path to the location of the embedding in each embedding object of the response.

reference/api/settings.mdx
```diff
@@ -1910,3 +1911,189 @@ Reset an index's typo tolerance settings to their [default value](#typo-toleranc
...
+| **`inputField`**       | Array of strings | []                                                                       | Path to location of document data in query Meilisearch sends to embedder                                                                                       |
+| **`inputType`**        | String           | `text`                                                                   | Type of data Meilisearch sends to embedder. Must be `text` or `textArray`                                                                                      |
+| **`query`**            | String           | Object                                                                   | Extra fields Meilisearch adds to queries Meilisearch sends to embedder                                                                                         |
+| **`pathToEmbeddings`** | Array            | []                                                                       | Path to vector embedding data in embedder response                                                                                                             |
```
dureuill, line 1953: Path to the location of embedding objects in the response. Useful for `inputType = textArray`.

reference/api/settings.mdx
```diff
@@ -1910,3 +1911,189 @@ Reset an index's typo tolerance settings to their [default value](#typo-toleranc
...
+| **`dimensions`**       | Integer          | Empty                                                                    | Number of dimensions in the chosen model. If not supplied, Meilisearch tries to infer this value                                                               |
+| **`revision`**         | String           | Empty                                                                    | Model revision hash                                                                                                                                            |
+| **`inputField`**       | Array of strings | []                                                                       | Path to location of document data in query Meilisearch sends to embedder                                                                                       |
+| **`inputType`**        | String           | `text`                                                                   | Type of data Meilisearch sends to embedder. Must be `text` or `textArray`                                                                                      |
```
dureuill, line 1951: - `text` means a single text will be sent in each embedding request, and a single embedding retrieved in each embedding response.
  - When using `text`, one must specify the path to the embedding in the response in `embeddingObject`.

- `textArray` means multiple texts will be sent in each embedding request, and the same number of embeddings retrieved in each embedding response.
  -  When using `textArray`, one must specify the path to the array of results in `pathToEmbeddings`, then the path from each item of the array to the embedding in `embeddingObject`.
  - When supported by the API of the remote embedder, it is more efficient to use `textArray` as text input, as it will result in fewer requests.

reference/api/settings.mdx
```diff
@@ -1910,3 +1911,189 @@ Reset an index's typo tolerance settings to their [default value](#typo-toleranc
...
+| **`query`**            | String           | Object                                                                   | Extra fields Meilisearch adds to queries Meilisearch sends to embedder                                                                                         |
+| **`pathToEmbeddings`** | Array            | []                                                                       | Path to vector embedding data in embedder response                                                                                                             |
+| **`embeddingObject`**  | String           | {}                                                                       | Name of the embedding object in embedder response                                                                                                              |
+| **`distribution`**     | Object           | Empty                                                                    | Describes the natural distribution of search results. Must contain two fields, `mean` and `sigma`, each containing a numeric value between `0` and `1`         |
```
dureuill, line 1955: I think we have some explanation about the distribution somewhere already. Maybe we should link it from here?

reference/api/settings.mdx
```diff
@@ -1910,3 +1911,189 @@ Reset an index's typo tolerance settings to their [default value](#typo-toleranc
...
+    "dimensions": 1536
+  }
+}
+```
+
+### Update embedder settings
+
+<RouteHighlighter method="PATCH" route="/indexes/{index_uid}/settings/embedders"/>
+
+Partially update the embedder settings for an index. When this setting is updated Meilisearch may reindex all documents and regenerate their embeddings.
```
dureuill, line 1993: I feel like at some point we should specify **when** Meilisearch will reindex everything, but this is currently in flux for v1.9 so we might want to wait a bit to do this (maybe opening an issue in the docs?).

reference/api/settings.mdx
```diff
@@ -1910,3 +1911,189 @@ Reset an index's typo tolerance settings to their [default value](#typo-toleranc
...
+| **`pathToEmbeddings`** | Array            | []                                                                       | Path to vector embedding data in embedder response                                                                                                             |
+| **`embeddingObject`**  | String           | {}                                                                       | Name of the embedding object in embedder response                                                                                                              |
+| **`distribution`**     | Object           | Empty                                                                    | Describes the natural distribution of search results. Must contain two fields, `mean` and `sigma`, each containing a numeric value between `0` and `1`         |
+
+Depending on your chosen embedder, some of these fields may be mandatory.
+
```
dureuill, line 1958: I feel like we should specify somewhere that do remove an embedder and all its associated embeddings, one should set this embedder to `null` in a settings update.

reference/api/settings.mdx
```diff
@@ -1910,3 +1911,189 @@ Reset an index's typo tolerance settings to their [default value](#typo-toleranc
...
+}
+```
+
+You can use the returned `taskUid` to get more details on [the status of the task](/reference/api/tasks#get-one-task).
+
+### Reset embedder settings
+
+<RouteHighlighter method="DELETE" route="/indexes/{index_uid}/settings/embedders"/>
+
+Removes all embedders from your index.
```
dureuill, line 2075: We should maybe specify that to remove a single embbeder, one should send a PATCH with the target embedder set to `null`.

(the question came up)

dureuill (changes requested):
Thank you for putting this reference together, that's a lot of work ☀️  There are some missing pieces, see inline comments

reference/api/settings.mdx
guimachiavelli: You mean something like this?

```sh
  curl \
    -X PATCH 'http://localhost:7700/indexes/kitchenware/settings' \
    --data-binary '{
      "embedders": {
        "default":  null
      }
    }'
```

reference/api/settings.mdx
guimachiavelli: Yeah, the embedders settings are sort of stretching the format. I'll follow your suggestion.

reference/api/settings.mdx
guimachiavelli: Just to be sure I understand this and `embeddingObject` correctly.

If `pathToEmbeddings` is `["A", "B", "C"]`, Meilisearch will look for the embedding objects in `response.A.B.C`, correct?

What happens if someone uses the wrong path? Will we throw an error? Or will the documents' `_vector` field be null/undefined?

reference/api/settings.mdx
guimachiavelli: I'm not sure I understand your statement. Setting aside any other piece of content, is the definition here correct? Or is it correct for ollama/openAI but incorrect for REST embeddders?

reference/api/settings.mdx
dureuill: yes, this will remove the "default" embedder

reference/api/settings.mdx
dureuill: The definitions are not incorrect, but they lack context.

reference/api/settings.mdx
dureuill: If `pathToEmbeddings` is `["A", "B", "C"]` and `embeddingObject` is `["D", "E", "F"]`

1. Meilisearch will look for the embedding objects in `response.A.B.C`
  - If any of `response`, `A` or `B` is not an object, then an error will be thrown
2. If `inputType == "text"`, Meilisearch will look for the embedding in `response.A.B.C.D.E.F`.
  - If any of `C`, `D`, `E` is not an object, then an error will be thrown
  - if `F` is not an array of numbers, then an error will be thrown
3. If `inputType == "arrayOfTexts"`, then Meilisearch will handle `response.A.B.C` as an array of elements, and for each element in elements, will look in `element.D.E.F`
  - If `C` is not an array, then an error will be thrown
  - If any of `element`, `D` or `E` is not an object, then an error will be thrown
  - If `F` is not an array of numbers, then an error will be thrown

guimachiavelli:
Thanks for the review, @dureuill!

I have updated the `embedder` documentation according to your feedback. I did my best for `pathToEmbeddings` and `embeddingObject`, but usage is quite complex due to behaviour changing signficantly depending on the value for `inputType`.

Also, I'm not convinced about the documentation for the `query` field. Could you give me a bit more context on what it's supposed to do and/or expected use cases?

dureuill:
> I did my best for pathToEmbeddings and embeddingObject, but usage is quite complex due to behaviour changing signficantly depending on the value for inputType.

I agree! I'm thinking of a better way of achieving this, here's what I get:

1. Make it so that `inputType` only governs whether an array or single text is sent in the query, as its name indicates.
2. Make it so that `pathToEmbeddings` can be `null`.
3. When `pathToEmbeddings` is `null`, then Meilisearch expects a single embedding in the response, and will look it up at the path described by `embeddingObject` in the response
4. When `pathToEmbeddings` is not `null`, then Meilisearch expects an array of embeddings in the response, and will look up the array of embeddings at the path described by `pathToEmbeddings` in the response, and then will look up each embedding at the path described by `embeddingObject` in each item of the array of embeddings.

Maybe for clarity we could rename these to `pathToEmbeddingArray` and `pathToEmbeddingData` or something. Similarly `inputField` could be `pathToInput`.

Or these could be `EmbeddingArrayPath`, `EmbeddingDataPath` and `InputPath`.

What do you think? Do you think it would make things clearer?

dureuill:
> Could you give me a bit more context on what [`query` is] supposed to do and/or expected use cases?

Sure, `query` is about sending the value of fields other than the texts to embed.

To send the embedding request, Meilisearch performs two steps:

1. Build an object that is equal to the value of the `query` field.
2. Inject the text(s) to embed at the path described by `inputField`.


For example, to implement the [OpenAI embedder API](https://platform.openai.com/docs/api-reference/embeddings/create), the final request needs to be:

```json
{
    "input": "TEXT TO EMBED",
    "model": "text-embedding-ada-002",
    "encoding_format": "float"
}
```

in this example, the `model` and `encoding_format` fields will be the same in all our requests to that embedder. So we have the field `query` be:
```json
{
  "model": "text-embedding-ada-002",
  "encoding_format": "float",
}
```

However the `input` field will differ for each request: this is where we should inject the text that we want to embed. So we would have `"inputField": ["input"]`.

Final embedder configuration would be:

```jsonc
{
  "url": "https://api.openai.com/v1/embeddings",
   "apiKey": "OPENAI_APIKEY",
  "query": {
    "model": "text-embedding-ada-002",
    "encoding_format": "float",
  },
  "inputField": ["input"],
  // omitting "inputType", "pathToEmbeddings" and "embeddingObject"
}
```

Similarly, an ollama request looks like:

```json
{
  "model": "nomic-embed-text",
  "prompt": "TEXT TO EMBED"
}
```

So you'd have `query = { "model": "nomic-embed-text" }` and `inputField = ["prompt"]`

curquiza (approved):
Ok for code samples

guimachiavelli:
@dureuill, Regarding the API, I think `pathToEmbeddingArray`, `pathToEmbeddingData`, and `pathToInput` are pretty good. I think the more we can "tag" fields to signal whether they refer to what Meilisearch sends or to what Meilisearch receives, the best. 

Possibly stupid idea: would we gain anything from creating two main fields, `input` and `response`? So the embedder object would be something like:

```json
{
  "default": {
    "source": "",
    "input": {
      "apiKey": "",
      "model": "",
      "revision": "",
      "dimensions": 123,
      "inputType": "",
      "pathToInput": [],
      "query": {}
    },
    "response": {
      "pathToEmbeddingArray": [],
      "pathToEmbeddingData": [],
      "distribution": {}
    },
  }
}
```

---

I was also thinking about `inputType`. It seems its only function is to determine whether a user should use `pathToEmbeddingArray` or not. Couldn't we cut `inputType` and infer a request is `textArray` from whether or not `pathToEmbeddingArray` has been specified? Or would that be bad API design? Or perhaps return an error if a user sets `pathToEmbeddingArray`, but `inputType` is `text`?

---

Regarding `query`, why do users need to specify `model` inside `query` for `rest` embedders, but as its own thing for `openAi`, `huggingFace`, and `ollama`?

reference/api/settings.mdx
guimachiavelli: You mean specifically for vector search stuff, or in general? Either way, yes, an issue might be the best approach. 

Here you go: #2831

guimachiavelli:
Ok, thanks for the answers, @dureuill. I'm curious on how the API will evolve before we stabilise it, especially if we manage to get more direct feedback from users (perhaps organising a poll or a couple of interviews?). A lot of my concerns might end up being fairly academic and mostly unimportant for the majority of people who are actually using vector search.

In any case, I think this PR is ready for an official review. I don't think you need to re-read everything, just the new section describing each embedder option in more detail: https://github.com/meilisearch/documentation/pull/2824/files#diff-a88efc3f5697059650c8e14b221124b09e9c2eb12aadc2290bb87a71456fd64aR1999-R2193

reference/api/settings.mdx
dureuill: I meant specifically for vector search stuff, but you're right that with the settings diff of v1.9 it would be better to have this in general.

Loving #2831, I think we can resolve this comment and continue the discussion there at a later date

reference/api/settings.mdx
```diff
@@ -1910,3 +1911,333 @@ Reset an index's typo tolerance settings to their [default value](#typo-toleranc
...
+This field is optional for `openAI` and `ollama` embedders. If you don't specify `apiKey`, Meilisearch will attempt to read it from environment variables `OPENAI_API_KEY` and `MEILI_OLLAMA_URL`, respectively.
+
+This field is incompatible with `huggingFace` and `userProvided` embedders.
+
+##### `model`
+
+The model your embedder uses when generating vectors. These are the officially supported models Meilisearch supports:
+
+- `openAi`: `openai-text-embedding-ada-002`, `text-embedding-3-small`, and `text-embedding-3-large`
+- `huggingFace`: `BAAI/bge-base-en-v1.5`
```
dureuill, line 2065: For huggingFace, we should maybe explicitly indicate that other BERT models may work.

reference/api/settings.mdx
```diff
@@ -1910,3 +1911,333 @@ Reset an index's typo tolerance settings to their [default value](#typo-toleranc
...
+`documentTemplate` is a string containing a [Liquid template](https://shopify.github.io/liquid/basics/introduction). Meillisearch interpolates the template for each document and sends the resulting text to the embedder. The embedder then generates document vectors based on this text.
+
+You may use the following context values:
+
+- `{{doc.FIELD}}`: `doc` stands for the document itself. `FIELD` must correspond to an attribute present on all documents value will be replaced by the value of that field in the input document
+- `{{fields}}`: a list of all the `field`s appearing in any document in the index. Each `field` object in this list has two properties: `name` and `value`. If a `field` does not exist in a document, `value` is `nil`
+
+For best results, build short templates that only contain highly relevant data. If working with a long field, consider [truncating it](https://shopify.github.io/liquid/filters/truncatewords/). If you do not manually set it, `documentTemplate` will include all document fields. This may lead to suboptimal performance and relevancy.
+
+This field is optional for all embedders.
```
dureuill, line 2084: We should maybe specify that while optional, it is recommended to fill it because the default template may render to long texts and contain information that is not relevant to the semantics of the document (such as the id of the document)

reference/api/settings.mdx
```diff
@@ -1910,3 +1911,333 @@ Reset an index's typo tolerance settings to their [default value](#typo-toleranc
...
+
+##### `dimensions`
+
+Number of dimensions in the chosen model. If not supplied, Meilisearch tries to infer this value.
+
+In most cases, `dimensions` should be the exact same value of your chosen model. Setting `dimensions` to a value lower than the model may lead to performance improvements and is only supported in the following OpenAI models:
+
+- `openAi`: `text-embedding-3-small`, `text-embedding-3-large`
+
+This field is mandatory for `rest` and `userProvided` embedders.
```
dureuill, line 2094:
```suggestion
This field is mandatory `userProvided` embedders.
```

If missing, `dimensions` are inferred for `rest` embedders with a test request.

reference/api/settings.mdx
```diff
@@ -1910,3 +1911,333 @@ Reset an index's typo tolerance settings to their [default value](#typo-toleranc
...
+
+##### `embeddingObject`
+
+Array indicating the path to the field containing a single set of vector embedding in a `textArray` embedder response.
+
+This field must be an array of strings. Each string should indicate the field containing vector data for a single element when the embedder returns multiple embeddings.
+
+For example, if `pathToEmbeddings` is `["A", "B", "C"]`, the document data should be in `response.A.B.C`, where `C` is an array of objects. For each element in `C`, Meilisearch will look for the path configured in `embeddingObject`. If `embeddingObject` is `["D", "E", "F"]`, vector data should be in `element.D.E.F`.
+
+This field is optional for the `rest` embedder. It is mandatory when `inputType` is `textArray`.
```
dureuill, line 2150:
```suggestion
This field is optional for the `rest` embedder.
```

The field is always optional and defaults to `[]`. 

In the case where `inputType` is `textArray` and `pathToEmbeddings` is `["A", "B", "C"]`, then the document data is in `response.A.B.C`, where `C` is an array of objects.

When `embeddingObject == []` in this situation, then each item of the array `C` is directly an embedding (that is, an array of floats).

reference/api/settings.mdx
```diff
@@ -1910,3 +1911,333 @@ Reset an index's typo tolerance settings to their [default value](#typo-toleranc
...
+`distribution` is an optional field compatible with all embedder sources. It must be an object with two fields:
+
+- `mean`: a number between `0` and `1` indicating the semantic score of "somewhat relevant" hits before using the `distribution` setting
+- `sigma`: a number between `0` and `1` indicating the average absolute difference in `_rankingScore`s between "very relevant" hits and "somewhat relevant" hits, and "somewhat relevant" hits and "irrelevant hits".
+
+Changing `distribution` does not trigger a reindexing operation.
+
+#### Example
+
+<CodeSamples id="update_embedders_1" />
```
dureuill, line 2197: I wonder if for the case of REST embedder we should include a lot of the examples I gave in our discussion from this PR, or if it should be left to a future REST embedder guide.

dureuill:
Ah, I also think it would be nice to have a list of the embedder with the allowed/mandatory parameter per embedder.

reference/api/settings.mdx
guimachiavelli: I want to keep the list restricted to stuff we know and have tested. 

Since that doesn't seem to be the case for all BERT models, I'd rather add it to the paragraph below as a "it may work, but use it at your own peril".

reference/api/settings.mdx
```diff
@@ -1910,3 +1911,333 @@ Reset an index's typo tolerance settings to their [default value](#typo-toleranc
...
+##### `model`
+
+The model your embedder uses when generating vectors. These are the officially supported models Meilisearch supports:
+
+- `openAi`: `openai-text-embedding-ada-002`, `text-embedding-3-small`, and `text-embedding-3-large`
+- `huggingFace`: `BAAI/bge-base-en-v1.5`
+
+Other models, such as those provided by Ollama and REST embedders may also be compatible with Meilisearch.
+
+This field is mandatory for `openAi`, `huggingFace`, and `Ollama` embedders.
```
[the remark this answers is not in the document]
guimachiavelli: What are the default values for `openAi` and `huggingFace`? `text-embedding-3-small` and `BAAI/bge-base-en-v1.5`?

reference/api/settings.mdx
guimachiavelli: Unless we know a significant percentage of users are going the REST route, I think keeping this example simple is better.

reference/api/settings.mdx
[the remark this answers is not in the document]
dureuill: BAAI/bge-base-en-v1.5 for hf and the ada one for openai

dureuill (approved):
Thank you for this huge addition 🎈 🎉

PR #2829, opened 2024-05-22
New "Engine" section

guimachiavelli:
Creates new "Engine" section for explanations on varied aspects of the engine itself.

Information about more specific topics such as relevancy and indexing will be moved to a dedicated section at a later date.


guimachiavelli:
This change was already preapproved outside github. Merging.

PR #2832, opened 2024-05-23
Tenant tokens rework

### LINKED ISSUE #2786: Users are struggling when creating tenant token
A user struggled to understand why the search requests were failing with their tenant token

https://github.com/meilisearch/meilisearch/discussions/4567#discussioncomment-9118613

They wanted to access the full index but without specifying any filter, so they needed to add `indexName: {}` which was not intuitive. Maybe we could detail this?

@guimachiavelli, to be honest, I did not check in the documentation if this explanation is already available, so feel free to close this issue if not useful

However I think it's still useful to work on tenant token experience. To let you know, and maybe we could do something on the product side as well: users often struggle with tenant token usage because they DON'T know why the request is forbidden (i.e. the tenant token is invalid). The message is really vague and does not indicate if the action or the index is forbidden. So maybe we could add some warning of basic errors to check have having issues with tenant token?

guimachiavelli:
This PR breaks the current "Multitenancy and tenant tokens" page into five new pages:

- Multitenancy and tenant tokens: an explanation describing common terms (multitenancy, tenant tokens), differences between tenant tokens and API keys, and general guidance on implementation
- Generate tenant tokens with an official SDK/third-party library/from scratch: short guides outlining the main steps when generating tenant tokens. The three are all very similar and mostly copypasted from the previous tenant tokens page. Generating tokens from scratch is not very thorough, not enough demand to make it better
- Tenant token payload reference: reference info for search rules, API key and UID, and expriy date

Also closes #2786


curquiza (dismissed):
Looks like it solves the original problem, thank you gui for this work!

learn/security/generate_tenant_token_scratch.mdx
```diff
@@ -0,0 +1,73 @@
...
+  }
+}
+```
+
+Consult the [token payload reference](/learn/security/tenant_token_reference) for more information on the requirements for each payload field.
+
+## Encode header and payload
+
+You must then encode both the header and the payload into `base64`, concatenate them, and generate the token by signing it using your chosen encryption algorithm.
+
```
irevoire, line 68: It would be nice to link to this website, which explains the concept of jwt and helps you create your own.
https://jwt.io/

PR #2837, opened 2024-06-03
Add Swedish to list of languages with optimized support

CaroFG:
# Pull Request

Adds Swedish to list of languages with optimized support


guimachiavelli (approved):
Thanks! 🙇

PR #2839, opened 2024-06-04
Update systemd service template

martinnj:
# Pull Request

## What does this PR do?
- It adds automatic restart to the systemd service template in the "Running in Production" section

The motivation is that if the service crashes (like from 6/ABRT errors during memory allocation, something I've seen several times during the last week) it should restart automatically and not stay dead until manual intervention.

## Feedback wanted
- I set it to `Restart=always` because that is what I needed for my usage, but I know some maintainers prefer the more conservative `on-failure` or `on-abort` behavior. I can substitute that if you'd rather have that.
- The fix for the `too_many_open_files` error described in the [docs here](https://www.meilisearch.com/docs/reference/errors/error_codes#too_many_open_files) also requires a fix in this file, basically you have to set `LimitNOFILE=<howevermanyfilesyouwant>` should I add that in this PR as well?

### DIFF
--- a/guides/running_production.mdx
+++ b/guides/running_production.mdx
@@ -112,6 +112,7 @@ WorkingDirectory=/var/lib/meilisearch
 ExecStart=/usr/local/bin/meilisearch --config-file-path /etc/meilisearch.toml
 User=meilisearch
 Group=meilisearch
+Restart=always
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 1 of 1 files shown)


guimachiavelli:
Hello @martinnj—thanks for opening this PR, it's much appreciated.

I would err towards being conservative in this case. So perhaps setting it `on-failure` might be better?

I would also not add the `too_many_open_files` fix to this guide.

@brunoocasali, what do you think?

martinnj:
There we go, changed the restart condition to `on-failure`, sorry about the messy git history, had to update my fork and rebase the branch. :)

guimachiavelli (approved):
Just got confirmation from @brunoocasali and we're good to go!

Also, @brunoocasali and I were discussing and he was really fond of your other point about the `too_many_open_files`. We'll definitely add this info, though possibly to another page where it can be easier to find for users who are not following this specific guide.

Thanks again for your effort, @martinnj!

PR #2841, opened 2024-06-04
Update CI to check links in guides

Strift:
This PR updates the CI to also check for links in the `config/sidebar-guides.json` file.


guimachiavelli (approved):
LGTM

Should I merge this immediately or should we merge something else first?

Just realised `/` in the branch name breaks the preview. Good to know. We should probably avoid it in the future.

PR #2855, opened 2024-06-12
changed useradd command flag '-b' to '-s'

szzd:
Updated the useradd command to use the -s flag instead of -b when creating the meilisearch user

# Pull Request

## What does this PR do?
The flag for "/bin/false" should be "-s" , ensuring the user cannot log in.

### DIFF
--- a/guides/running_production.mdx
+++ b/guides/running_production.mdx
@@ -53,7 +53,7 @@ Meilisearch is now running, but it is not publicly accessible.
 Running applications as root exposes you to unnecessary security risks. To prevent that, create a dedicated user for Meilisearch:
 
 ```sh
-useradd -d /var/lib/meilisearch -b /bin/false -m -r meilisearch
+useradd -d /var/lib/meilisearch -s /bin/false -m -r meilisearch
 ```
 
 ## Step 3: Create a configuration file


guimachiavelli:
Thanks for taking the time to open this PR, @szzd!

This looks ok at first glance, but our policy is to always get a programmer to validate changes like this. @curquiza, can you take a quick look?

curquiza (commented):
Hey @szzd 
i don't really understand why you change `-b` into `-s`. Indeed, -s looks like to set the path the shell, which is not the case here

https://www.man7.org/linux/man-pages/man8/useradd.8.html

szzd:
In the documentation, the command provided is:
`useradd -d /var/lib/meilisearch -b /bin/false -m -r meilisearch`
Flag `-d` and `-b` both set the home directory, `-b` just concatenate the username with provided option. So I don't think we need both.
On the other had second flag use `/bin/false` which is an shell path, generally set with `-s` flag. I thought we are trying to create a user that shouldn't have login access.
Sorry if I got it wrong.

guimachiavelli (approved):
Re-reading this after a brief talk with @curquiza and looking at the `man` pages, I think you're absolutely right, @szzd.

Thanks again for fixing this and taking the time to respond to our comments. 🙇

Approving and merging.

PR #2868, opened 2024-06-17
The `facetDistribution` is also exhaustive when using `page=0`

Kerollmops:
This PR specifies that the `facetDistribution` is also exhaustive when using `page=0`.


guimachiavelli (approved):
Thanks for that addition!

LGTM!

PR #2876, opened 2024-06-18
v1.9: Add new `retrieveVectors` and `_vectors` behaviour

### LINKED ISSUE #2857: v1.9: AI-powered search updates
v1.9 introduces three significant changes to AI-powered search.

1. Meilisearch now interprets empty document `_vectors` as no embedding instead of an embedding with 0 dimensions
2. Document `_vectors` now also accept object values. The object may have two fields, `embeddings` and `regenerate`. If `regenerate` is `true`, the document's embeddings are replaced every time an auto-embedder runs. If `regenerate` is `false`, embeddings remain unchanged and must be manually updated
3. AI-powered search responses no longer include `_vectors` by default. Users must set `retrieveVectors: true`, a new search parameter, to see `_vector` data search result responses

## Tasks
- [ ] Add empty embeddings and `regenerate` information to either tutorial or embedder index settings
- [ ] Update search parameters with `retrieveVectors`

## Reference
- PRD: https://www.notion.so/meilisearch/v1-9-AI-search-changes-e90d6803eca8417aa70a1ac5d0225697
- engine issue: https://github.com/meilisearch/meilisearch/issues/4607
- SME: @dureuill

guimachiavelli:
Closes #2857


reference/api/documents.mdx
```diff
@@ -248,6 +248,45 @@ An array of documents. Each document is represented as a JSON object.
 ]
 ```
 
+#### `_vectors`
+
+`_vectors` is a special document attribute containing vector embeddings for AI-powered search. It accepts different types of values.
+
+`_vectors` may be an array of numbers:
```
dureuill, line 255: No, `_vectors` is always a map, whose keys are the name of embedders (or future embedders).

The value of each key can be an array or an object.

reference/api/documents.mdx
```diff
@@ -248,6 +248,45 @@ An array of documents. Each document is represented as a JSON object.
...
+
+```json
+[
+  {
+    "id": 452,
+    "title": "Female Trouble",
+    "overview": "Delinquent high school student Dawn Davenport runs away from home and embarks upon a life of crime.",
+    "_vectors": [0.0, 0.1, 0.003, 0.75]
```
dureuill, line 263:
```suggestion
    "_vectors": { "default":  [0.0, 0.1, 0.003, 0.75] }
```

reference/api/documents.mdx
```diff
@@ -248,6 +248,45 @@ An array of documents. Each document is represented as a JSON object.
...
+  },
+]
+```
+
+`_vectors` may also be an object with two fields, `embeddings` and `regenerate`:
```
dureuill, line 268:
```suggestion
Each subobject of `_vectors` may also be an object with up to two fields, `embeddings` and `regenerate`:
```

reference/api/documents.mdx
```diff
@@ -248,6 +248,45 @@ An array of documents. Each document is represented as a JSON object.
...
+
+```json
+[
+  {
+    "id": 452,
+    "title": "Female Trouble",
+    "overview": "Delinquent high school student Dawn Davenport runs away from home and embarks upon a life of crime.",
+    "_vectors": {
+      "embeddings": [],
```
dureuill, line 277: It would be nicer to have an actual embedding here, even with only 3 floats.

```suggestion
      "embeddings": [0.1, 0.2, 0.3],
```

reference/api/documents.mdx
```diff
@@ -248,6 +248,45 @@ An array of documents. Each document is represented as a JSON object.
...
+      "regenerate": true
+    }
+  }
+]
+```
+
+`embeddings` must be an array of numbers.
```
dureuill, line 284:
```suggestion
If provided, `embeddings` must be an array of numbers, representing a single embedding for that embedder for that document, or an array of arrays of numbers, representing multiple embeddings for that embedder for that document.

If missing or `null`, the embeddings for that document will remain at their current value, except if `regenerate` is `true` and other fields in the document were modified in the same document update.
```

I suggest we introduce `regenerate` first.

reference/api/documents.mdx
curquiza, line 263: Don't we want to promote this format instead?
`_vectors: { default: [...] } }

@dureuill

curquiza (approved):

curquiza:
Approval for code samples 👆

reference/api/documents.mdx
```diff
@@ -248,6 +248,45 @@ An array of documents. Each document is represented as a JSON object.
...
+
+`regenerate` must be a Boolean value. If `regenerate` is `true`, the supplied embeddings will be overwritten the next time Meilisearch automatically generates document embeddings. If `regenerate` is `false`, the document's embeddings will be automatically replaced by an embedder.
```
dureuill, line 286:
```suggestion
`regenerate` must be a Boolean value. If `regenerate` is `true`, embeddings will be regenerated for that document and embedder when the document is updated. If `regenerate` is `false`, the last value of the embeddings will be kept on document updates.
```

reference/api/documents.mdx
```diff
@@ -248,6 +248,45 @@ An array of documents. Each document is represented as a JSON object.
...
+
+If `_vectors` is empty or `null`, Meilisearch treats the document as not having any embeddings. This document is then ignored during AI-powered searches.
```
dureuill, line 288:
```suggestion
If the key for an embedder inside of `_vectors`  is empty or `null`, Meilisearch treats the document as not having any embeddings for that embedder. This document is then ignored during AI-powered searches.
```

reference/api/search.mdx
```diff
@@ -1118,3 +1122,17 @@ Use a custom vector to perform a search query. Must be an array of numbers corre
...
+
+### Display `_vectors` in response (experimental)
+
+**Parameter**: `retrieveVectors`<br />
+**Expected value**: `true` or `false`<br />
+**Default value**: `false`
+
+Return document vector data in AI-powered searches.
+
+This parameter has no effect on queries that do not include `vector` or `hybrid`.
```
dureuill, line 1134:
```suggestion
```

This is incorrect. If the document contains vector data, setting `retrieveVectors` to true will retrieve them, regardless of the kind of search used.

reference/api/documents.mdx
dureuill: Hello @curquiza 

this format is incorrect and will not work as soon as an embedder config has actually been declared.

See my comments below

reference/api/documents.mdx
guimachiavelli: Could you expand on this? I find the explanation flows better by saying "embeddings defines the vectors for this document" -> "regenerate regenerates the embeddings your previously defined".

reference/api/search.mdx
```diff
@@ -1118,3 +1122,15 @@ Use a custom vector to perform a search query. Must be an array of numbers corre
 <Capsule intent="warning">
 Meilisearch will return an error if you use `vector` before activating your instance's `vectorStore` and [configuring a custom embedder](/reference/api/settings#embedders-experimental).
 </Capsule>
+
+### Display `_vectors` in response (experimental)
+
+**Parameter**: `retrieveVectors`<br />
```
dureuill, line 1128: retrieveVectors is also added to various document fetch routes (get documents) and the similar route

reference/api/documents.mdx
```diff
@@ -248,6 +248,54 @@ An array of documents. Each document is represented as a JSON object.
...
+    }
+  }
+]
+```
+
+`embeddings` must be an array of numbers representing a single embedding for that document. It may also be an array of arrays of numbers representing multiple embeddings for that document. `embeddings` is an optional field.
+
+`regenerate` must be a Boolean value. If `regenerate` is `true`, Meilisearch automatically generates embeddings for that document immediately and every time the document is updated. If `regenerate` is `false`, Meilisearch keeps the last value of the embeddings on document updates.
+
+If `_vectors` is empty or `null`, Meilisearch treats the document as not having any embeddings. This document is then ignored during AI-powered searches.
```
dureuill, line 293: > If `_vectors` is empty or `null`, Meilisearch treats the document as not having any embeddings.

I don't think that's true. I think this paragraph can probably be removed since we have the contents of the capsule already

reference/api/documents.mdx
```diff
@@ -248,6 +248,54 @@ An array of documents. Each document is represented as a JSON object.
...
+
+<Capsule intent="note">
+If the key for an embedder inside `_vectors` is empty or `null`, Meilisearch treats the document as not having any embeddings for that embedder. This document is then ignored during AI-powered searches.
```
dureuill, line 296:
```suggestion
If the key for an embedder inside `_vectors` is empty or `null`, Meilisearch treats the document as not having any embeddings for that embedder. This document is then returned last during AI-powered searches.
```

Ah, sorry for not noticing earlier. The current behavior of the semantic search is same as the geodistance and any  sort: documents that don't fulfill the condition (don't have an embedding for that embedder, don't have a geopoint, don't have the sorted field) are returned after any document that does fulfill the condition in an unspecified order.

reference/api/documents.mdx
dureuill: Sure.

- `embeddings` overrides the current embeddings of the documents
- separately from this, `regenerate` describes the behavior when the document is updated.

So, the flow could be:

By default, Meilisearch will generate embeddings when a document is added and each time the document is updated such that the filling the `documentTemplate` with the document's data produces a different text as before the update.

When setting `regenerate` to `false,` this behavior will be disabled, and the embedding will remain the same regardless of the document's updates.

Additionally, if providing the `embeddings` field, the embeddings for that document and embedder will be overridden. 

The various combinations of `regenerate` and `embeddings` can achieve various use cases:

### 1. Providing embeddings to a `userProvided` embedder

Setting `regenerate` to `false` and the `embeddings` to the target value will achieve this use case.

### 2. Importing embeddings that were created externally using a similar embedder once, then letting regeneration occurs when updating documents.

Import the embeddings in `embeddings`, and set `regenerate` to `true` so that future updates regenerate the embeddings.

### 3. Opting-out of embeddings for a single document.

Similar to use case (1), with `regenerate` set to `false`, but with an `embeddings` equal to `[]`

### 4. "Freezing" the current values of embeddings.

Pass only `regenerate` to `false` without specifying `embeddings`. The document will keep the embeddings currently in its database, regardless of future updates.

guimachiavelli: I'm still digesting this, but a question. Are these two examples equivalent?

```json
{
  "default": [0.1, 0.2, 0.3]
}
```

```json
{
  "default": {
    "embeddings": [0.1, 0.2, 0.3],
    "regenerate" : true
}
```

If they're identical, what is the use case for the simple array format? Wouldn't it be better to remove it and simplify the API, so users don't need to think which format is best for them?

guimachiavelli: Ah, ok, so by default Meilisearch will not regenerate custom embeddings. That makes sense, thanks.

I think you have found the correct framing for this. The new format (with embeddings and regenerate) is the actual format, whilst the the regular array is simply a shorthand. Idk, perhaps it was obvious to you, but I kept seeing them as two separate things.

reference/api/documents.mdx
```diff
@@ -248,6 +251,57 @@ An array of documents. Each document is represented as a JSON object.
...
+"_vectors": {
+  "default": [0.003, 0.1, 0.75]
+}
+```
+
+Vector embeddings added with the shorthand are not replaced when Meilisearch generates new embeddings. The above example is equivalent to:
+
+```json
+"default": {
+  "embeddings": [0.1, 0.2, 0.3],
```
dureuill, line 296: We should probably use the same vector value here, otherwise it is a bit confusing 🤣 
```suggestion
  "embeddings": [0.003, 0.1, 0.75],
```

reference/api/search.mdx
```diff
@@ -1162,3 +1169,15 @@ Use a custom vector to perform a search query. Must be an array of numbers corre
...
+**Expected value**: `true` or `false`<br />
+**Default value**: `false`
+
+Return document vector data with search results.
```
dureuill, line 1179: Do we want to specify the format here?

The document vector data will be injected in documents in a `_vectors` field that follows the same format as for passing the vector data to Meilisearch. Returned vector data always uses the long-form variant of the format and specifies if vectors will be regenerated on document updates.

dureuill (approved):
I have a few final comments but we're definitely converging

PR #2965, opened 2024-08-29
Add a guides for generating embeddings with Mistral, OpenAI, Voyage, and Cloudflare

qdequele:
I created a first guide to use Mistral as an embedder in Meilisearch. Once this is confirmed I would be able to duplicate this content for all the embedders (OpenAI, Voyage, Cloudflare, and others).

PR #2966, opened 2024-09-03
feat: update UTM params on vector & hybrid search content

Strift:
I'm updating the UTMs parameters on links to:
- fix incorrect sources (should be: `utm_source=docs`)
- add `utm_medium` to help differentiate pages (will be more important with #2965)


qdequele (approved):
LGTM

guimachiavelli:
@Strift, please notify me when you merge PRs that don't pass the broken links check. I understand why you ignored the errors, but I got a bit of a scare when I opened a new PR myself just now and got hammered with 30 warnings 😅 

I'll open a ticket with Notum so they can fix it ASAP, since its usefulness is heavily reduced due to how hard it is to separate false positives from actual errors.

PR #2974, opened 2024-09-09
Fix the known limitation around the size of faceted values

irevoire:
The limitation was slightly wrong, as you can see in meilisearch we keep a « safe margin of 32 bytes »:
https://github.com/meilisearch/meilisearch/blob/db0cf3b2ed0ca6cf6549510bf373681e83b775dd/milli/src/lib.rs#L106-L110

Discovered while trying to understand https://github.com/meilisearch/meilisearch/issues/4915


guimachiavelli (approved):
LGTM! Thanks!

PR #2976, opened 2024-09-10
Explicitly state attributes are case-sensitive

### LINKED ISSUE #2808: Specify attributes are case sensitive
It could be worth warning users that attributes are case sensitive, especially in the following guides:
- [Configuring index settings with the Meilisearch Cloud interface](https://www.meilisearch.com/docs/learn/configuration/configuring_index_settings)
- [Configuring index settings with the Meilisearch API](https://www.meilisearch.com/docs/learn/configuration/configuring_index_settings_api)

A free-trial customer was getting 0 search results, because he had copied the searchable attributes from the mini-dashboard, so all uppercase, so maybe worth mentioning something in the search preview page too?

guimachiavelli:
Closes #2808

### DIFF
--- a/learn/getting_started/documents.mdx
+++ b/learn/getting_started/documents.mdx
@@ -24,13 +24,13 @@ Nothing will be shared between two indexes if they contain the exact same docume
 
 ## Fields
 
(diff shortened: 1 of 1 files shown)


CaroFG (approved):
LGTM!

PR #2979, opened 2024-09-11
Move content out of `learn/experimental` section

guimachiavelli:
@meilisearch/integration-team: this PR makes the following changes to code samples:
- adds a handful of new samples:
  - `experimental_get_metrics_1`
  - `experimental_post_logs_stderr_1`
  - `experimental_post_logs_stream_1`
  - `experimental_delete_logs_stream_1`


curquiza (commented):
> add an experimental_ prefix to the following existing code sample ids, without any changes to the samples themselves: 

I have to say I'm not pro this change 😞 
- a lot of changes on our side to do
- features can be stabilized in a future, and it will be the case of AI, so we will have to change it everywhere again...

Can't you manage your code samples without this change? 🙏 would definitely help us on our side.

guimachiavelli:
@macraig, in short, I'm moving everything out of the experimental section and into places users are more likely to find that info naturally (e.g. `CONTAINS` in the filter reference). If you have the time, I'd appreciate if you could give it a quick look just in case there's any info that's clearly out of date.

macraig (dismissed):
The new structure looks good to me!

curquiza (changes requested):
discussed with @guimachiavelli and I would rather we don't do this breaking due to the lot of work coming after this breaking for integrations

macraig:
> discussed with @guimachiavelli and I would rather we don't do this breaking due to the lot of work coming after this breaking for integrations

Oh my bad, as @guimachiavelli 's request for review was after your initial comment @curquiza , I thought you were ok with these changes.

guimachiavelli:
Apologies, I should have clarified that I intended @macraig's review to be on the content of the PR itself so at least that part has been approved. I'll change the current code samples back to their previous to reflect in a future commit, once I've dealt with other priorities.

guimachiavelli:
@curquiza, I have rolled back the changes to existing code samples, instead indicating they're  experimental by adding them to the bottom of the file with a comment explaining it.

New code samples stay the same, but are sort of low priority.

curquiza (approved):
Ok for me 👌

PR #3003, opened 2024-09-26
Clarify that only custom API keys can be reused after migration

### LINKED ISSUE #2905: How to keep keys when restarting Meilisearch
I've seen recently some user wondering about deterministic API keys, they want to know how to keep the same keys when restarting an instance.

> I'm using Meilisearch in Render and every month they update the server manually and so the API keys are updated, which causes problems in my applications because I have to update the API keys.
> 
> Is there any way to keep the API keys fixed even if the server restarts? Can I generate the API keys and pass them on via an environment variable?

Also they seem confused about which keys are deterministic:

> I'm running meilisearch with the same master key (and backed by the same EBS volume) and I'm sometimes getting different 
> admin API keys
> 
> I was under the impression keys were deterministically based on the master key, and would only change when the master key changes?
> Is this assumption invalid or is this a bug (either in my code or Meilisearch)?

guimachiavelli:
Closes #2905


CaroFG (approved):
Sorry I did not see this earlier. LGTM!

PR #3016, opened 2024-10-04
doc(vector): update Ollama endpoint

Guikingone:
Hi 👋🏻 

As stated in the official Ollama documentation, the endpoint is `/api/embed` and not `/api/embeddings`.

Doc: https://github.com/ollama/ollama/blob/main/docs/api.md#generate-embeddings

# Pull Request

## Related issue
Fixes #<issue_number>

## What does this PR do?

Update on the embeddings endpoints for Ollama

## PR checklist

- [X] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [X] Have you read the contributing guidelines?
- [X] Have you made sure that the title is accurate and descriptive of the changes?

### DIFF
--- a/learn/experimental/vector_search.mdx
+++ b/learn/experimental/vector_search.mdx
@@ -116,7 +116,7 @@ curl \
     "embedders": {
       "default": {
         "source": "ollama",
-        "url": "http://localhost:11434/api/embeddings",
+        "url": "http://localhost:11434/api/embed",
         "apiKey": "OLLAMA_API_KEY",
         "model": "MODEL_NAME",
         "documentTemplate": "A document titled {{doc.title}} whose description starts with {{doc.overview|truncatewords: 20}}"


guimachiavelli (approved):
Thanks for the contribution, @Guikingone, it's much appreciated.

Looks good to me as it does match their documentation, but I'm requesting a review from one of our devs just to be sure and in case this is interesting for them to know.

guimachiavelli:
Seems like this was also [reported on the engine repo](https://github.com/meilisearch/meilisearch/pull/4983) and is actually a possible breaking change of Ollama's API.

I'll delay merging until the devs have had a chance to do some research and decide the best course of action.

guimachiavelli:
@dureuill, I saw you made a few steps forward regarding this on the engine repo. From what I understand, on the engine side this is more a question of deciding the best default value for `url`. 

In the case of the docs, I'd rather reflect the official guidance from Ollama. Do you agree?

dureuill:
hey @guimachiavelli :wave:

the issue is that ollama currently offers two endpoints to perform embedding:

1. A deprecated one, that is hidden from docs
2. A supported one, the only documented one, only available on fairly recent ollama versions.

Meilisearch will not remove support for the deprecated endpoint in an attempt to not break existing ollama embedders.

It means we should document that Meilisearch accepts both endpoints. We should document that the deprecated one is deprecated, and the ollama version where the new one appeared.

Due to this change, Meilisearch will no longer accept any URL for the ollama embedder, but only an URL that is suffixed by one of the two supported endpoints.

The documentation for the `url` parameter for the `ollama` source could look thus:

```
Specifies the *full* path to the embedding endpoint of the remote ollama server.

The passed `url` **must** end with one of the following suffixes:

1. `/api/embed`. This is the recommended endpoint for ollama > v0.3.4.
2. `/api/embeddings`. This endpoint has been deprecated on ollama's end. However, it is the only available endpoint for ollama < v0.3.4, so Meilisearch supports it.

Any other suffix will cause Meilisearch to return an error.

If missing, the value of this parameter will be taken from the `MEILI_OLLAMA_URL` environment variable.
If that environment variable is also missing, then the default value of `http://localhost:11434/api/embeddings` will be used.
```

dureuill:
> So, according to some quick testing on a local instance, current behaviour is that url accepts any string for all embedder types that support url.

Exactly, and this was confusing to users who wanted to change the default URL, because ollama only documents the new `embeddings` API, but Meilisearch currently only makes calls in the format of the deprecated one.
So, when changing the default URL, a user would use the documented one, and then get errors from Meilisearch due to incompatibilities between the deprecated and the supported API of ollama.

> Values deviating from those two will not be accepted.

We may allow a parent to the `api/` part, but we'll need the `api/embeddings` or `api/embed` part. We'll use it to decide which API to use when talking to the remote ollama server.

> If that's the case, then I might close this PR and implement the required changes in v1.12.

Sounds like a good plan :+1:

guimachiavelli:
Thanks, @dureuill!

@Guikingone, thank you so much for your bringing this to your attention. Given the information we now have, this is sadly a bit more involved than changing an example URL. I will close this PR and address the issue with the updates for the upcoming release.

Guikingone:
Of course, thanks for the feedback and @dureuill for the extra informations 🙂

PR #3027, opened 2024-10-15
v1.11: Binary quantization usage recommendation

guimachiavelli:
As discussed on [Slack](https://meilisearch.slack.com/archives/C07CLB9J4BH/p1728480901080119) (private link), we should provide clearer recommended usage guidelines for `binaryQuantized`.


curquiza (commented):
Can you also make it visible in the indexing good practice.
Like "if you have indexing perf issue AND you are using vector search, you can enable binary quantization"?

curquiza (approved):

PR #3058, opened 2024-11-28
Add new article on how to choose embedder type

### LINKED ISSUE #3040: Create new “How to choose an embedder” guide
Recent customer feedback indicates users are struggling to move beyond the basic AI-powered search tutorial and implement hybrid search in their own projects.

Main points to address:
- Define what are the main differences between available embedders
- Ideally we should be less “this is the best option for use case X” and more “if your app does Y, choose embedders with high A”
  - pinging @dureuill: do you think it is possible to create general guidelines as described above? What are the things users should look at when trying to decide which embedder to choose?

guimachiavelli:
Closes #3040

### DIFF
--- a/learn/ai_powered_search/choose_an_embedder.mdx
+++ b/learn/ai_powered_search/choose_an_embedder.mdx
@@ -29,4 +29,10 @@ In these cases, you will have to supply your own embeddings.
 
 Although it returns very relevant search results, the Hugging Face embedder must run directly in your server. This may lead to lower performance and extra costs when you are hosting Meilisearch in a service like DigitalOcean or AWS.
 
-That said, Hugging Face can be a good embedder for datasets under 10k documents that you don't plan to update often. Meilisearch Cloud does not support Hugging Face embedders.
+That said, Hugging Face can be a good embedder for datasets under 10k documents that you don't plan to update often. 
+
+<Capsule intent="note">
+Meilisearch Cloud does not support embedders with `{"source": "huggingFace"}`.
+
+To use  Hugging Face in the Cloud,  use [HuggingFace inference points with the REST embedder](/guides/ai/huggingface).
+ </Capsule>
--- a/learn/ai_powered_search/choose_an_embedder.mdx
+++ b/learn/ai_powered_search/choose_an_embedder.mdx
@@ -1,9 +1,9 @@
 ---
-title: How to choose an embedder — Meilisearch documentation
+title: Which embedder should I choose? — Meilisearch documentation
 description: General guidance on how to choose the embedder best suited for projects using AI-powered search.
 ---
 
-# How to choose an embedder
+# Which embedder should I choose?
 
 Meilisearch officially supports many different embedders, such as OpenAI, Hugging Face, and Ollama, as well as the majority of embedding generators with a RESTful API.
 
--- a/learn/ai_powered_search/choose_an_embedder.mdx
+++ b/learn/ai_powered_search/choose_an_embedder.mdx
@@ -23,7 +23,7 @@ If you are already using a specific model from a compatible embedder, choose Mei
 
 Meilisearch does not support searching images, audio, or any other content not presented as text. This limitation applies to both queries and documents. For example, Meilisearch's built-in embedder sources cannot search using an image instead of text. They also cannot use text to search for images without attached textual metadata.
 
-In these cases, you will have to supply your own embedder. Meilisearch Cloud does not support custom embedders.
+In these cases, you will have to supply your own embedder.
 
 ## Only choose Hugging Face when self-hosting small static datasets
 
--- a/learn/ai_powered_search/choose_an_embedder.mdx
+++ b/learn/ai_powered_search/choose_an_embedder.mdx
@@ -1,11 +1,11 @@
 ---
 title: How to choose an embedder — Meilisearch documentation
-description: This article contains general guidance on how to choose the embedder best suited for projects using AI-powered search.
+description: General guidance on how to choose the embedder best suited for projects using AI-powered search.
 ---
 
 # How to choose an embedder
 
-Meilisearch officially supports many different embedders, such as OpenAI, Hugging Face, and Ollama, as well as the majority of embedding generators with a RESTful API. It can be difficult to understand their differences and how to pick one.
+Meilisearch officially supports many different embedders, such as OpenAI, Hugging Face, and Ollama, as well as the majority of embedding generators with a RESTful API.
 
 This article contains general guidance on how to choose the embedder best suited for your project.
 
@@ -13,7 +13,7 @@ This article contains general guidance on how to choose the embedder best suited
 
 OpenAI returns relevant search results across different subjects and datasets. It is suited for the majority of applications and Meilisearch actively supports and improves OpenAI functionality with every new release.
 
-In the majority of cases, and especially if this is your first time working with LLMS, choose OpenAI.
+In the majority of cases, and especially if this is your first time working with LLMs and AI-powered search, choose OpenAI.
 
 ## If you are already using a specific AI service, choose the REST embedder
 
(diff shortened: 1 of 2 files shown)


learn/ai_powered_search/choose_an_embedder.mdx
```diff
@@ -0,0 +1,32 @@
...
+
+## If you are already using a specific AI service, choose the REST embedder
+
+If you are already using a specific model from a compatible embedder, choose Meilisearch's REST embedder. This ensures you continue building upon tooling and workflows already in place with minimal configuration necessary.
+
+## If dealing with non-textual content, choose the user-provided embedder
+
+Meilisearch does not support searching images, audio, or any other content not presented as text. This limitation applies to both queries and documents. For example, Meilisearch's built-in embedder sources cannot search using an image instead of text. They also cannot use text to search for images without attached textual metadata.
+
+In these cases, you will have to supply your own embedder.
```
macraig, line 26:
```suggestion
In these cases, you will have to supply your own embeddings.
```

learn/ai_powered_search/choose_an_embedder.mdx
```diff
@@ -0,0 +1,32 @@
...
+
+## Only choose Hugging Face when self-hosting small static datasets
```
macraig, line 28: This was initially advice for Cloud users using HF embedders because we were generating the embeddings locally (running on our Cloud infra). This is no longer the case, we removed the option on the Cloud and we've replaced it with the [Hugging Face Inference points using the REST embedder](https://www.meilisearch.com/docs/guides/ai/huggingface) option. 

Self-hosted users can still use HuggingFace as an embedder option, as they can tweak their infrastructure to fit their specific needs.

macraig: We can either remove this section, or point users in the direction of how to set a HF embedder using the REST option (for Cloud) and the API reference (for self hosted)

learn/ai_powered_search/choose_an_embedder.mdx
```diff
@@ -0,0 +1,32 @@
...
+
+Meilisearch officially supports many different embedders, such as OpenAI, Hugging Face, and Ollama, as well as the majority of embedding generators with a RESTful API.
+
+This article contains general guidance on how to choose the embedder best suited for your project.
+
+## When in doubt, choose OpenAI
+
+OpenAI returns relevant search results across different subjects and datasets. It is suited for the majority of applications and Meilisearch actively supports and improves OpenAI functionality with every new release.
+
+In the majority of cases, and especially if this is your first time working with LLMs and AI-powered search, choose OpenAI.
```
macraig, line 16: (Feel free to ignore this one as it might be just me) I wonder if this doesn't sound a bit too biased towards OpenAI, making it sound like it's our provider of choice over others. Maybe we can phrase it around "ease of config" or "easiest for beginners" as it only requires pasting the OpenAI key.

learn/ai_powered_search/choose_an_embedder.mdx
```diff
@@ -0,0 +1,32 @@
...
+
+## If you are already using a specific AI service, choose the REST embedder
+
+If you are already using a specific model from a compatible embedder, choose Meilisearch's REST embedder. This ensures you continue building upon tooling and workflows already in place with minimal configuration necessary.
+
+## If dealing with non-textual content, choose the user-provided embedder
```
macraig, line 22:
```suggestion
## If dealing with non-textual content, choose the Custom (user-provided) embedder
```

learn/ai_powered_search/choose_an_embedder.mdx
```diff
@@ -0,0 +1,32 @@
...
+
+Meilisearch does not support searching images, audio, or any other content not presented as text. This limitation applies to both queries and documents. For example, Meilisearch's built-in embedder sources cannot search using an image instead of text. They also cannot use text to search for images without attached textual metadata.
```
macraig, line 24: pinging @dureuill as I'm not 100% sure about this one

macraig (dismissed):

learn/ai_powered_search/choose_an_embedder.mdx
dureuill: yes that is correct. We may want to specify that, *by supplying the embeddings generated using their own embedder*, the user can indeed achieve these use cases.

learn/ai_powered_search/choose_an_embedder.mdx
guimachiavelli: I don't want parentheses in the heading, but I'll include "custom embedder" somewhere in the section body.

learn/ai_powered_search/choose_an_embedder.mdx
guimachiavelli: I think there may be something a bit confusing about how we handle hugging face right now.

We have a `huggingFace` embedder source, but it works very differently from all other built-in embedders in that it must run locally. At the same time, if I understand correctly, there's HF's inference points, which instead work more or less exactly like open AI and ollama. So we end up treating hugging face as a monolithic choice, which may not be ideal for users who must use hugging face but don't want to run it locally.

guimachiavelli: This may be my ignorance speaking, but it almost seems as if our official hugging face source gives the worst of the two available options for existing HF users.

macraig: Yes, I see how it gets a bit confusing. I'll try to describe the options clearly and we can go from there.

Options available on Cloud:
- OpenAI
- Custom (user-provided)
- REST: used to set up HF inference points, ollama, Cloudflare, or any embedder running on a third-party that supports REST format

Optionas available for self-hosted:
- OpenAI
- user-provided
- REST: set up any embedder running on a third-party that supports REST
- huggingface: set up a local HF model of choice that will run locally
- ollama: similar to REST but specific to ollama config

guimachiavelli: Sorry, I don't think I expressed myself very well.

What I meant is that our official HF source (`source: "huggingFace"`) seems to be "worse" than using inference points with REST for the majority of users.

Of course, "worse" is not exactly correct, but if I'm just an average user who wants to use HF, I'll look at the documentation and think "oh, they have official support for hugging face, I'll use that." If they're a cloud user, they'll realise they can't actually use the HF source and need to reach for the generic REST source. If they're self-hosted, they'll have to jump through some hoops to set it everything up locally, and even then they'll not have as much flexibility as e.g. `openAI` or `Ollama`.

IMHO, a dedicated source implies benefits over generic solutions. You can use OpenAI with the REST source, but it's easier if you use the `openAI` source. The opposite seems to be true for the majority of Hugging Face users: you can use the official HF source, but it will require more effort and care than using the HF with the REST source.

Of course, some users will want to run HF locally and it's great they have the option to do so, but as far as I'm aware of they are a minority, and their use case is not super aligned with our current design focus.

macraig: > you can use the official HF source, but it will require more effort and care than using the HF with the REST source.

Do you mean the effort of configuring it or because it requires using local resources? 
In any case, the choice on how to use HF will be on whether they want to set up the embedder locally or not.

learn/ai_powered_search/choose_an_embedder.mdx
```diff
@@ -0,0 +1,32 @@
...
+
+In these cases, you will have to supply your own embeddings.
+
+## Only choose Hugging Face when self-hosting small static datasets
+
+Although it returns very relevant search results, the Hugging Face embedder must run directly in your server. This may lead to lower performance and extra costs when you are hosting Meilisearch in a service like DigitalOcean or AWS.
+
+That said, Hugging Face can be a good embedder for datasets under 10k documents that you don't plan to update often. Meilisearch Cloud does not support Hugging Face embedders.
```
macraig, line 32:
```suggestion
That said, Hugging Face can be a good embedder for datasets under 10k documents that you don't plan to update often. On Meilisearch Cloud we recommend using Hugging Face embedders by combining the [REST embedder with HuggingFace inference points](https://www.meilisearch.com/docs/guides/ai/huggingface#introduction).
```

learn/ai_powered_search/choose_an_embedder.mdx
guimachiavelli: I mean effort in both configuration (running a local embedder is more complex than sending data to an API and having Meilisearch do the rest) and in terms of how much it differs from other embedders. Using OpenAI and ollama is very similar, using the default HF embedder is completely different.

Anyway, this discussion is completely out of scope for this PR by now 😅 

My main point is from the documentation pov there's something unintuitive about our `huggingFace` source. If you feel it's necessary, we can discuss it in the next product sync, but it's up to you. Whether the docs pov corresponds to an actual issue for users that we need to anything about is completely your call 👍

dureuill (dismissed):
Thank you, approved :+1:

PR #3072, opened 2024-12-11
v1.12: New `/batches` route

### LINKED ISSUE #3061: v1.12: New `/batches` API route
v1.12 introduces the `/batches` route. Its main goal is to give more information on how tasks are being processed and helps diagnose in case of failure.

- new `GET` `/batches` 
    - supports a several parameters
    - need defaults from #person/tamo
    - returns a list of `batchObjects`
- new `GET` `/batches/:uid` 
    - does not accept query/body parameters
    - returns a single `batchObject`
- updated `/tasks` and `/tasks/:uid`
    - new parameter `batchUids`, returns all tasks whose `batchUid` match the specified value
- updated task object
    - includes new field `batchUid`
- new error: `InvalidBatchUids` for `/batch` and `/batch/:uid`

## Tasks
- [x] Create new `reference/api/batches.mdx`
- [x] Update `reference/api/tasks.mdx`
- [x] Update `learn/async/filtering_tasks.mdx` -> probably add batches to title and/or a box saying everything there is valid for batches
- [x] Update `learn/async/paginating_tasks.mdx` -> same as above
- [x] Update `learn/async/asynchronous_operations.mdx` -> mention batches

## References
- issue on core: https://github.com/meilisearch/meilisearch/issues/4977
- PRD (private link): https://www.notion.so/meilisearch/Enhance-visibility-on-batched-tasks-1194b06b651f810b8fe0fab5d72846a8?pvs=4
- SME: @irevoire

guimachiavelli:
Closes #3061


curquiza (approved):
ok for code samples

irevoire (changes requested):

reference/api/batches.mdx
```diff
@@ -0,0 +1,244 @@
...
+
+| Name            | Description                                        |
+| :---------------| :------------------------------------------------- |
+| **`name`**      | A string describing the operation                  |
+| **`total`**     | The total number of operations in the step         |
+| **`finished`**  | The number of operations Meilisearch has completed |
+
+#### `percentage`
+
+The progress of the current set of operations, calculated from all current steps and substeps.
```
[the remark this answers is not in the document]
guimachiavelli: Thanks for the suggestion. I'll push an update stating `percentage` does not grow linearly, and added a tip to `steps` advising users to monitor the last step's `finished` in case a batch is taking longer than expected.

reference/api/batches.mdx
```diff
@@ -0,0 +1,244 @@
...
+List with the `types` of tasks contained in the batch.
+
+#### `indexUids`
+
+List of the number of tasks in the batch separated by the indexes they affect.
+
+### `duration`
+
+**Type**: String<br />
+**Description**: The total elapsed time the batch spent in the `processing` state, in [ISO 8601](https://www.ionos.com/digitalguide/websites/web-development/iso-8601/) format
```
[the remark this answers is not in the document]
irevoire: Looks good to me

reference/api/batches.mdx
```diff
@@ -0,0 +1,250 @@
...
+| **`total`**       | The total number of operations in the step         |
+| **`finished`**    | The number of operations Meilisearch has completed |
+
+<Capsule intent="tip" title="">
+If Meilisearch is taking longer than expected to process a batch, monitor the `steps` array. If the `finished` field of the last item in the `steps` array does not update, Meilisearch may be stuck.
+</Capsule>
+
+#### `percentage`
+
+The percentage of completed operations, calculated from all current steps and substeps. This value is a rough estimate and may not always reflect the current state of the batch due to how different tasks are processed more quickly than others.
```
irevoire, line 83: Sorry, I'm bothering you again one last time on this part :confonded: 

```suggestion
The percentage of completed operations, calculated from all current steps and substeps. This value is a rough estimate and may not always reflect the current state of the batch due to how different steps are processed more quickly than others.
```
-----

And for more context, just so we're 100% aligned on the difference between a step and a task:

With Louis, we think we should **really** avoid using the term `task` in this context because we don't want the user to think that a batch is processing tasks in order or something.
We want to be sure that people don't misunderstand this:
- A batch is crafted by grouping multiple tasks together, that's true
- Once it starts processing there is no concept of task anymore, everything is merged together
- All the tasks are processing **together, at the same time**
=> That's why in this context, we really want to use the term « step » instead of task. We're processing the step « extracting word docids » for all the tasks at the same time.

guimachiavelli: Thanks for this, it makes sense.

Just one question to clarify some aspects of the batching process. Is "once meilisearch starts processing the batch, there's no concept of task" a bit of an exaggeration? We still keep track of tasks so we can tell users task A was successful and task B failed, right?

irevoire: Yes, but in reality we know that beforehand.
Once we starts processing the batch, we don't have the information anymore

irevoire (approved):
Awesome :ok_hand:

PR #3075, opened 2024-12-12
v1.12: New `/tasks` `reverse` parameter

### LINKED ISSUE #3062: v1.12: New `reverse` parameter for `GET` `/tasks`
v1.12 adds a new `reverse` parameter to `GET` `/tasks`. When set to `true`, displays results in reverse chronological order (i.e. oldest to newest). This is mostly to improve developer QoL, as seeing the oldest occurrences of a task can be useful during debugging.

## Tasks
- [ ] Update `reference/api/tasks.mdx` -> parameter table at `reference/api/tasks#query-parameters`

## References
- engine issue: https://github.com/meilisearch/meilisearch/issues/5047
- PRD (private link): https://www.notion.so/meilisearch/Reverse-the-order-of-the-task-queue-1364b06b651f801ca127d982289892e5?pvs=4
- SME: @irevoire

guimachiavelli:
Closes #3062

### DIFF
--- a/reference/api/tasks.mdx
+++ b/reference/api/tasks.mdx
@@ -233,7 +233,6 @@ Task results are [paginated](/learn/async/paginating_tasks) and can be [filtered
 | **`afterEnqueuedAt`**  | `*` (all tasks)                | [Filter tasks](/learn/async/filtering_tasks) by their `enqueuedAt` field                                                                 |
 | **`afterStartedAt`**   | `*` (all tasks)                | [Filter tasks](/learn/async/filtering_tasks) by their `startedAt` field                                                                  |
 | **`afterFinishedAt`**  | `*` (all tasks)                | [Filter tasks](/learn/async/filtering_tasks) by their `finishedAt` field                                                                 |
-| **`reverse`** | `false` | If `true`, returns results in the reverse order, from oldest to most recent |
 
 ### Response
 


reference/api/tasks.mdx
```diff
@@ -232,6 +232,7 @@ Task results are [paginated](/learn/async/paginating_tasks) and can be [filtered
 | **`afterEnqueuedAt`**  | `*` (all tasks)                | [Filter tasks](/learn/async/filtering_tasks) by their `enqueuedAt` field                                                                 |
 | **`afterStartedAt`**   | `*` (all tasks)                | [Filter tasks](/learn/async/filtering_tasks) by their `startedAt` field                                                                  |
 | **`afterFinishedAt`**  | `*` (all tasks)                | [Filter tasks](/learn/async/filtering_tasks) by their `finishedAt` field                                                                 |
+| **`reversed`** | `false` | If `true`, returns results in the reverse order, from oldest to most recent |
```
irevoire, line 235:
```suggestion
| **`reverse`** | `false` | If `true`, returns results in the reverse order, from oldest to most recent |
```

Oh, sorry, the PRD was wrong (I saw you already noticed).

Don't you think that should be right after the `limit` and `from` since it's related to the way we navigate / paginate instead of doing filtering?

PR #3079, opened 2024-12-12
v1.12: Update Postman collection

### LINKED ISSUE #3066: v1.12: Update Postman

guimachiavelli:
Closes #3066

### DIFF
--- a/assets/misc/meilisearch-collection-postman.json
+++ b/assets/misc/meilisearch-collection-postman.json
@@ -1000,7 +1000,7 @@
 								},
 								{
 									"key": "reverse",
-									"value": "false",
+									"value": "true",
 									"disabled": true
 								}
 							]
--- a/assets/misc/meilisearch-collection-postman.json
+++ b/assets/misc/meilisearch-collection-postman.json
@@ -1,7 +1,7 @@
 {
 	"info": {
 		"_postman_id": "cc6bb097-033d-4f65-8704-f10e4e4b10d0",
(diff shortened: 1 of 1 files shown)


assets/misc/meilisearch-collection-postman.json
```diff
@@ -711,6 +716,11 @@
 									"key": "from",
 									"value": "10",
 									"disabled": true
+								},
+								{
+									"key": "reverse",
+									"value": "false",
```
curquiza, line 722: I'm not sure you want to put `false` as value here, for the example, since `false` is the default value (if I'm correct). `true` would be more appropriate so that it's an example with immediate visible impact 😊

curquiza (commented):
Good to me, only 1 suggetion 😊

curquiza:
@guimachiavelli
Do you manage to import the file? It fails on my Postman

curquiza (approved):
It's now working, probably an issue on Postman side

guimachiavelli:
After spending some time trying to debug this, I believe it's a temporary issue with postman. I got extremely erratic behavior (e.g. I remove all endpoints from the "Settings" route/subroutes in one go, importing works; I remove each endpoint manually, importing fails), until suddenly everything started working again, including the current collection file in this PR.

Looking through the electron app inspector, I see a lot of 429 errors (too many requests) and failed `Promise`s, so I'm guessing it's something on their servers.

PR #3104, opened 2025-01-15
Update the embedder settings documentation

Kerollmops:
There seems to be an issue in the documentation as it refers to  the `_vector` field instead of the `_vectors` one.


dureuill (approved):

guimachiavelli (approved):
Thanks for fixing this! I've done a quick review and couldn't find any other `_vector` mentions, so it seems to have been an isolated typo.

PR #3107, opened 2025-01-16
Updated `expected values` for limit in search.mdx

### LINKED ISSUE #2774: Clarify accepted values for `/search` `limit`
Documentations currently states the `limit` search parameter accepts "any positive integer".

We should update the `/search` reference so it's clear `limit` also accepts `0` as a value. This will cause Meilisearch to return no search results.

## References
- slack discussion (private): https://meilisearch.slack.com/archives/D022TFFR09G/p1711544852015859

RecreationalMath:
Fixes #2774

Updated `expected values` for limit in `reference/api/search.mdx` file to include zero as a valid input to the limit parameter.

### DIFF
--- a/reference/api/search.mdx
+++ b/reference/api/search.mdx
@@ -373,7 +373,7 @@ If you want to skip the **first** result in a query, set `offset` to `1`:
 ### Limit
 
 **Parameter**: `limit`<br />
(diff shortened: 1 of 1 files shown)


guimachiavelli (approved):
LGTM, thanks!

PR #3112, opened 2025-01-20
Add missing spacing in json blocks

ellnix:
I was reading the Tasks reference, and noticed that a lot of the JSON blocks had some minor missing spaces. Next thing I know I have a vim macro in my hands and here we are.

Sorry if this is a waste of time. If this is superfluous (which it is) feel free to close the PR, no hard feelings :smile:


guimachiavelli (approved):
Not a waste of time at all! These have been bugging me for the longest time, but there's always something with higher priority than white space that looks bad but works just fine. My peace of mind thanks you.

PR #3123, opened 2025-01-25
Correct the location of the dumps folder

ellnix:
While this would be accurate if following the official instructions and running ./meilisearch right after downloading, there's nothing preventing someone from running the binary from a different directory.


guimachiavelli (approved):
Minor wording change, but looks good to me otherwise! Thanks!

PR #3138, opened 2025-01-30
v1.13: Dumpless upgrades

### LINKED ISSUE #3135: v1.13: Dumpless upgrades
v1.13 introduces a new experimental CLI flag: `--experimental-dumpless-upgrade`. Using it simplifies the upgrading process, allowing people to simply update the binary and launch Meilisearch instead of having to generate dumps.

This feature is only publicly relevant for self-hosted users. It may also be useful for the cloud, but this would be an internal change without any visibility for customers.

New update process: 
1. download new binary
2. interrupt `meilisearch` 
3. overwrite old binary 
4. relaunch with `./meilisearch --experimental-dumpless-upgrade`

> [!WARNING]  
> The dumpless upgrade is not currently atomic, so it's possible some processes fail and Meilisearch still finalizes the upgrade. If this happens, it may corrupt the database. Users should take a snapshot of their db before performing a dumpless upgrade.

Cloud updates remain the same: press update button.

## Tasks
- [ ] update `learn/self_hosted/configure_meilisearch_at_launch.mdx`
- [ ] update `learn/update_and_migration/updating.mdx`

## References
- SME: @irevoire 
- Engine issue: https://github.com/meilisearch/meilisearch/issues/5162
- Changes summary: https://meilisearch.notion.site/Dumpless-upgrade-fff4b06b651f81f1acafe24d4687b3f7

guimachiavelli:
Closes #3135

### DIFF
--- a/learn/update_and_migration/updating.mdx
+++ b/learn/update_and_migration/updating.mdx
@@ -55,7 +55,7 @@ curl \
   -X POST 'MEILISEARCH_URL/snapshots'
 ```
 
-Once the snapshot is available in your [snapshots folder](/learn/self_hosted/configure_meilisearch_at_launch#snapshot-destination), proceed to the next step.
+Meilisearch will respond with a partial task object. Use its `taskUid` to monitor the snapshot creation status. Once the task is completed, proceed to the next step.
 
 ### Step 2: Stop the Meilisearch instance
 
--- a/learn/update_and_migration/updating.mdx
+++ b/learn/update_and_migration/updating.mdx
@@ -36,13 +36,124 @@ Once the project has been successfully updated, you will receive an email confir
 
 ## Updating a self-hosted Meilisearch instance
 
+You may update a self-hosted instance in one of two ways: with or without a dump.
+
 <Capsule intent="warning">
 This guide only works for v0.15 and above. If you are using an older Meilisearch release, please [contact support](https://discord.meilisearch.com) for more information.
 </Capsule>
 
-### Step 1: Export data
+### Dumpless upgrade <NoticeTag type="experimental" label="experimental" />
 
-#### Verify your database version
+Dumpless upgrades are available when upgrading from Meilisearch >=v1.12 to Meilisearch >=v1.1
+
+#### Step 1: Make a backup
+
+Dumpless upgrades are an experimental feature. Because of that, it may in rare occasions partially fail and result in a corrupted database. To prevent data loss, create a snapshot of your instance:
+
+```sh
+curl \
+  -X POST 'MEILISEARCH_URL/snapshots'
+```
+
+Once the snapshot is available in your snapshots folder, proceed to the next step.
+
+### Step 2: Stop the Meilisearch instance
+
+Next, stop your Meilisearch instance.
+
+<Tabs.Container labels={['Local installation', 'Cloud platforms']}>
+
+<Tabs.Content label="Local installation">
+
+If you're running Meilisearch locally, you can stop the program with `Ctrl + c`.
+
+</Tabs.Content>
+
+<Tabs.Content label="Cloud platforms">
+
+If you're running Meilisearch as a `systemctl` service, connect via SSH to your cloud instance and execute the following command to stop Meilisearch:
+
+```bash
+systemctl stop meilisearch
+```
+
+You may need to prefix the above command with `sudo` if you are not connected as root.
+
+</Tabs.Content>
+</Tabs.Container>
+
+#### Step 3: Install the new Meilisearch binary
+
+Install the latest version of Meilisearch using:
+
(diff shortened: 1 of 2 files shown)


learn/update_and_migration/updating.mdx
```diff
@@ -36,13 +36,124 @@ Once the project has been successfully updated, you will receive an email confir
...
+#### Step 1: Make a backup
+
+Dumpless upgrades are an experimental feature. Because of that, it may in rare occasions partially fail and result in a corrupted database. To prevent data loss, create a snapshot of your instance:
+
+```sh
+curl \
+  -X POST 'MEILISEARCH_URL/snapshots'
+```
+
+Once the snapshot is available in your [snapshots folder](/learn/self_hosted/configure_meilisearch_at_launch#snapshot-destination), proceed to the next step.
```
irevoire, line 58: Shouldn't we tell people to look at their tasks instead of looking at the directory?

learn/update_and_migration/updating.mdx
```diff
@@ -36,13 +36,124 @@ Once the project has been successfully updated, you will receive an email confir
...
+<Tabs.Content label="Cloud platforms">
+
+```sh
+meilisearch --experimental-dumpless-upgrade 
+```
+
+</Tabs.Content>
+</Tabs.Container>
+
+Meilisearch should launch normally and immediately create a new `UpgradeDatabase` task. This task is processed immediately and cannot be canceled. You may follow its progress by using the `GET /tasks` endpoint to obtain its `taskUid`, then querying `GET /tasks/TASK_UID`.
```
irevoire, line 144:
```suggestion
Meilisearch should launch normally and immediately create a new `UpgradeDatabase` task. This task is processed immediately and cannot be canceled. You may follow its progress by using the `GET /tasks?types=UpgradeDatabase` endpoint to obtain its `taskUid`, then querying `GET /tasks/TASK_UID`.
```

This way even if there are a lot of enqueued tasks you'll see it.

We could also tell people to use the new batch route since this route will also give users the `progress`ion of the upgrade process:
```
GET /batches?types=UpgradeDatabase
```

learn/update_and_migration/updating.mdx
```diff
@@ -36,13 +36,124 @@ Once the project has been successfully updated, you will receive an email confir
...
+
+While the task is processing, you may continue making search queries. You may also enqueue new tasks. Meilisearch will only process new tasks once `UpgradeDatabase` is completed.
+
+<Capsule intent="warning">
+If after the upgrade is completed and the task status is set to `failed` or Meilisearch returns internal error messages to your queries, [restart your instance from the snapshot](/learn/data_backup/snapshots#starting-from-a-snapshot) you generated during step 1.
```
irevoire, line 149:
```suggestion
If after the upgrade is completed and the task status is set to `failed` or Meilisearch returns internal error messages to your queries, [restart your instance from the snapshot](/learn/data_backup/snapshots#starting-from-a-snapshot) you generated during step 1 and open an issue to the [Meilisearch repository](https://github.com/meilisearch/meilisearch).
```

Maybe we could also propose to upgrade via a dump? Or it's not needed since it's the category immediately below

irevoire (changes requested):

irevoire (approved):
I left one last comment just in case you want to change the tone of a sentence, but overall, it’s good to me 👍

learn/update_and_migration/updating.mdx
```diff
@@ -36,13 +36,124 @@ Once the project has been successfully updated, you will receive an email confir
...
 </Capsule>
 
-### Step 1: Export data
+### Dumpless upgrade <NoticeTag type="experimental" label="experimental" />
+
+Dumpless upgrades are available when upgrading from Meilisearch >=v1.12 to Meilisearch >=v1.13
+
+#### Step 1: Make a backup
+
+Dumpless upgrades are an experimental feature. Because of that, it may in rare occasions partially fail and result in a corrupted database. To prevent data loss, create a snapshot of your instance:
```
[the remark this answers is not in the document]
guimachiavelli: hmmm…if I understand you correctly, the difference sounds like a technicality.

As in, if something goes wrong, we're fairly sure the issue will lie with the engine, not the upgrade process itself. Whatever the actual cause, though, the fact we might have them is more or less the reason why this is experimental, right?

macraig (approved):

learn/update_and_migration/updating.mdx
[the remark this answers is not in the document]
irevoire: Huum I would say it’s experimental mainly because it’s not ACID, the day it’s ACID, and we can make something that handle unexpected internal error correctly then we’ll be good.

But yeah it’s mainly a technicality

PR #3164, opened 2025-02-18
REST embedder guide

### LINKED ISSUE #3129: New guide: How to populate REST embedder `request` and `response` settings
We have plenty of dedicated guides for different embedders with `source: rest`. These are helpful for users using that exact embeddings provider, but not so much for people implementing other services.

Instead of explicitly covering each and every provide in existence, it's a good idea to have a more general set of instructions so users have the tools to do what they want without direct guidance.

## Tasks
This new guide should:
- [ ] state the requirements for a provider to be compatible with AI-powered search
- [ ] give a high-level description of the process of finding a provider's request/response schema
- [ ] two examples in which we fill in the `request` field for a provider
- [ ] two examples in which we fill in the `response` field for a provider
- [ ] summarize the process in the conclusion
- [ ] end guide with links to dedicated mistral/cloudflare/etc

## References
- AI-powered search GA planning: https://www.notion.so/meilisearch/Path-to-AI-search-GA-54544e3ae8f345ebb2dbf4dbeecc6f28?pvs=4#17a4b06b651f806e80e9da3721d15166
- SME: @dureuill, @gmourier

guimachiavelli:
Fixes #3129


learn/ai_powered_search/rest_response_request.mdx
```diff
@@ -0,0 +1,254 @@
...
+
+Now the embedder object is complete, update your index settings:
+
+```sh
+curl \
+  -X PATCH 'MEILISEARCH_URL/indexes/INDEX_NAME/settings/embedders' \
+  -H 'Content-Type: application/json' \
+  --data-binary '{
+    "EMBEDDER_NAME": {
+      "source": "rest",
```
dureuill, line 225: We are missing the `url` parameter, which is mandatory for `rest` embedders.

learn/ai_powered_search/rest_response_request.mdx
```diff
@@ -0,0 +1,254 @@
...
+}
+```
+
+In Cloudflare's case, the model is part of the API route itself and doesn't need to be specified in your `request`.
+
+### The embedding prompt
+
+The prompt corresponds to the data that the provider will use to generate your document embeddings. Its specific name changes depending on the provider you chose. In Mistral, this is the `input` field. In Cloudflare, it's called `text`.
+
+Most providers accept either a string or an array of strings. A single string will generate one request per document in your database. An array of strings allows Meilisearch to send up to 10 documents in one request.
```
dureuill, line 59:
```suggestion
Most providers accept either a string or an array of strings. A single string will generate one request per document in your database. An array of strings allows Meilisearch to send up to 10 documents in one request, reducing the number of calls to the provider when indexing documents.
```

learn/ai_powered_search/rest_response_request.mdx
```diff
@@ -0,0 +1,254 @@
...
+```json
+{
+  "EMBEDDER_NAME": {
+    "source": "rest",
+    "request": {
+      "text": "{{text}}"
+    },
+    "response": {
+      "result": {
+        "data": "{{embedding}}"
```
dureuill, line 151: Cloudflare seems to always return the data as an array of embeddings, even when `text` is a single string and not an array.

So you should have 
```suggestion
        "data": "[{{embedding}}]"
```

learn/ai_powered_search/rest_response_request.mdx
```diff
@@ -0,0 +1,254 @@
...
+        "{{..}}"
+      ]
+    },
+    "apiKey": "PROVIDER_API_KEY",
+    "dimensions": MODEL_DIMENSIONS,
+    "documentTemplate": "SHORT_AND_RELEVANT_DOCUMENT_TEMPLATE"
+  }
+}
+```
+
```
dureuill, line 214: `rest` embedders also support the [`headers`](https://www.meilisearch.com/docs/reference/api/settings#headers) parameter, to send custom headers when sending the request.

This enables custom authentication schemes such as the [`api-key` header](https://learn.microsoft.com/en-us/azure/ai-services/openai/reference#request-header-1) required by Azure OpenAI.

learn/ai_powered_search/rest_response_request.mdx
```diff
@@ -0,0 +1,341 @@
+---
+title: Configure a REST embedder
+description: Create Meilisearch embedders using any provider with a REST API
+---
+
+# Configure a REST embedder
+
+You can integrate any embedding generator with Meilisearch if your chosen provider offers a public REST API.
```
gmourier, line 8: Only text embedding generator

learn/ai_powered_search/rest_response_request.mdx
```diff
@@ -0,0 +1,341 @@
...
+  "EMBEDDER_NAME": {
+    "source": "rest",
+    "url": "PROVIDER_URL",
+    "request": {
+      "model": "MODEL_NAME",
+      "input": "{{text}}"
+    },
+    "response": {
+      "data": {
+        "text": "{{embedding}}"
```
gmourier, line 163:
```suggestion
        "embedding": "{{embedding}}"
```

If we use Mistral example here

learn/ai_powered_search/rest_response_request.mdx
```diff
@@ -0,0 +1,341 @@
...
+{
+  "EMBEDDER_NAME": {
+    "source": "rest",
+    "url": "PROVIDER_URL",
+    "request": {
+      "text": "{{text}}"
+    },
+    "response": {
+      "result": {
+        "data": "[{{embedding}}]"
```
gmourier, line 207: A bit confusing to me there, would it works?

learn/ai_powered_search/rest_response_request.mdx
```diff
@@ -0,0 +1,341 @@
...
+        },
+        "{{..}}"
+      ]
+    },
+    "apiKey": "PROVIDER_API_KEY",
+  }
+}
+```
+
+You should also set a `documentTemplate`. Good templates are short and include only highly relevant document data:
```
gmourier, line 275: Why? A link to documentTemplate could be cool here!

learn/ai_powered_search/rest_response_request.mdx
```diff
@@ -155,18 +174,51 @@ Cloudflare's embeddings are instead located in `response.result.data`:
...
 
+<Capsule intent="note">
+The prompt data type does not necessarily match the response data type. For example, Cloudflare always returns an array of embeddings, even if the prompt in your request was a string.
+</Capsule>
+
 Meilisearch silently ignores fields not pointing to an `{{embedding}}` value.
 
+## The embedding header
+
+Your provider might also request you to add specific headers to your request. For example, Azure's AI services require a `api-key` header containing an API key.
```
dureuill, line 220: I think the Azure service is called Azure OpenAI

learn/ai_powered_search/rest_response_request.mdx
```diff
@@ -0,0 +1,341 @@
...
+{
+  "EMBEDDER_NAME": {
+    "source": "rest",
+    "url": "PROVIDER_URL",
+    "request": {
+      "text": "{{text}}"
+    },
+    "response": {
+      "result": {
+        "data": "[{{embedding}}]"
```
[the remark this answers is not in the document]
dureuill: Hmm, looks like my previous suggestion was wrong 🤔 

```suggestion
        "data": ["{{embedding}}"]
```

The placeholder string must be exactly `"{{embedding}}"`. It so happens that the response is an array contained in the `result.data` key, whose first element is the embedding we want.

Alternatively, as I see the current situation might spur confusion, it might be better to find an alternative provider that actually returns the embedding outside of an array, for illustration purposes.

learn/ai_powered_search/rest_response_request.mdx
```diff
@@ -155,18 +174,51 @@ Cloudflare's embeddings are instead located in `response.result.data`:
...
+      }
+    },
+    "header": {
+      "FIELD_NAME": "FIELD_VALUE"
+    }
+  }
+}
+```
+
+By default, Meilisearch includes a `Content-Type` header. It may also include an authorization bearer token, if you have supplied an API key.
```
dureuill, line 244: Should we specify the value of the Content-Type header?

- `Content-Type: application/json`

learn/ai_powered_search/rest_response_request.mdx
```diff
@@ -155,18 +174,51 @@ Cloudflare's embeddings are instead located in `response.result.data`:
...
+    "url": "PROVIDER_URL",
+    "request": {
+      "text": "{{text}}"
+    },
+    "response": {
+      "result": {
+        "data": "[{{embedding}}]"
+      }
+    },
+    "header": {
```
dureuill: I'm pretty sure the configuration key is `headers` with an `s` https://www.meilisearch.com/docs/reference/api/settings#headers

learn/ai_powered_search/rest_response_request.mdx
guimachiavelli: It will come as soon as we have the document template best practices ready 😉

learn/ai_powered_search/rest_response_request.mdx
guimachiavelli: I don't think it's super necessary. The point of the guide is to show people how to use `request`/`response`/`headers` to match the requirements from their provider. It's relevant that people know there are two default headers, but they can consult the reference if they need exhaustive information.

learn/ai_powered_search/rest_response_request.mdx
guimachiavelli: They seem to have a few different AI products, which their documentation refers to as "Azure AI services". I think I prefer to use the more generic term here, since it's likely to remain relevant even if the name changes.

learn/ai_powered_search/rest_response_request.mdx
guimachiavelli: I decided to use Cloudflare and a generic scenario for this section since I couldn't find an embedder that did not return data in an array.

learn/ai_powered_search/rest_response_request.mdx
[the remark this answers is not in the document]
dureuill: The `/api/embeddings` endpoint of ollama works that way:

```
{
  "request": {
    "model": <MODEL>,
    "prompt": "{{text}}"
  },
  "response": {
    "embedding": "{{embedding}}"
  }
}
```

Of note however that this endpoint is _deprecated_ in favor of the more modern `/api/embed` that does return and array of embeddings.

Sounds like returning an array of embeddings is the current consensus.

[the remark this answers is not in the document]
guimachiavelli: Completely forgot about ollama (tried Mistral, Cohere, Cloudflare, Voyage), but I really don't want to point users towards deprecated APIs.

gmourier (dismissed):
Awesome guide, it will be so helpful

learn/ai_powered_search/rest_response_request.mdx
```diff
@@ -0,0 +1,337 @@
...
+If the response contains multiple embeddings, use `"{{..}}"` as its second value:
+
+```json
+{
+  "EMBEDDER_NAME": {
+    "source": "rest",
+    "url": "PROVIDER_URL",
+    "request": {
+      "model": "MODEL_NAME",
+      "input": "{{text}}"
```
dureuill, line 176:
```suggestion
      "input": ["{{text}}", "{{..}}"]
```

for the response to contains multiple embeddings, the request needs to contain multiple embeddings, so both the request and response must use the `"{{..}}"` repeat placeholder syntax.

Meilisearch will actually enforce this and return an error if only one of the request or response contains the repeat placeholder syntax.

learn/ai_powered_search/rest_response_request.mdx
```diff
@@ -0,0 +1,337 @@
...
+      }
+    }
+  }
+}
+```
+
+The prompt data type does not necessarily match the response data type. For example, Cloudflare always returns an array of embeddings, even if the prompt in your request was a string.
+
+Meilisearch silently ignores `response` fields not pointing to an `"{{embedding}}"` value.
+
```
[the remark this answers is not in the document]
guimachiavelli: That's interesting, but isn't this scenario covered by the general instruction of following the path to the embeddings even if said path is very short?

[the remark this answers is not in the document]
dureuill: it is, just had this particular question come up in support once, so I thought that the guide could maybe have some kind of "btw, this specific scenario is indeed covered by the general instruction".

You may or may not add it though, do as you think is best.

learn/ai_powered_search/rest_response_request.mdx
guimachiavelli: Oh, I didn't know we returned an error for mismatched placeholder existence across request/response. Thanks for letting me know.

learn/ai_powered_search/rest_response_request.mdx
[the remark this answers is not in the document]
guimachiavelli: Ok, if it's come up in support, it's a solid argument to make it very explicit. I'll update the PR,  thanks!

learn/ai_powered_search/rest_response_request.mdx
dureuill: Yes, sample error messages:

- Error while generating embeddings: user error: in `response`: `response` has multiple embeddings, but `request` has only one text to embed
- Error while generating embeddings: user error: in `response`: `response` has a single embedding, but `request` has multiple texts to embed

dureuill (approved):
thank you

learn/ai_powered_search/configure_rest_embedder.mdx
```diff
@@ -0,0 +1,367 @@
+---
+title: Configure a REST embedder
+description: Create Meilisearch embedders using any provider with a REST API
+---
+
+# Configure a REST embedder
+
+You can integrate any text embedding generator with Meilisearch if your chosen provider offers a public REST API.
+
+Integrating a REST embedder with Meilisearch varies depending on the provider and the way it structures its data. This guide shows you where to find the information you need, then walks you through configuring your Meilisearch embedder based on the information you found.
```
guimachiavelli, line 10:
```suggestion
The process of integrating a REST embedder with Meilisearch varies depending on the provider and the way it structures its data. This guide shows you where to find the information you need, then walks you through configuring your Meilisearch embedder based on the information you found.
```

learn/ai_powered_search/configure_rest_embedder.mdx
```diff
@@ -0,0 +1,367 @@
...
+        "{{..}}"
+      ]
+    }
+  }
+}
+```
+
+When using array prompts, the first item must be `{{text}}`. If you want to send multiple documents in a single request, the second array item must be `{{..}}`. When using `"{{..}}"`, it must be present in both `request` and `response`.
+
+When using other tools, `input` might be called something else, like `text` or `prompt`:
```
guimachiavelli, line 112:
```suggestion
When using other embedding providers, `input` might be called something else, like `text` or `prompt`:
```

learn/ai_powered_search/configure_rest_embedder.mdx
```diff
@@ -0,0 +1,367 @@
...
+      "model": "MODEL_NAME",
+      "text": "{{text}}"
+    }
+  }
+}
+```
+
+### Provide other request fields
+
+You may add as many fields to the `request` object as you need. Meilisearch will send them as configured when it queries the embeddings provider.
```
guimachiavelli, line 129:
```suggestion
You may add as many fields to the `request` object as you need. Meilisearch will include them when querying the embeddings provider.
```

learn/ai_powered_search/configure_rest_embedder.mdx
```diff
@@ -0,0 +1,367 @@
...
+}
+```
+
+The prompt data type does not necessarily match the response data type. For example, Cloudflare always returns an array of embeddings, even if the prompt in your request was a string.
+
+Meilisearch silently ignores `response` fields not pointing to an `"{{embedding}}"` value.
+
+## The embedding header
+
+Your provider might also request you to add specific headers to your request. For example, Azure's AI services require a `api-key` header containing an API key.
```
guimachiavelli, line 246:
```suggestion
Your provider might also request you to add specific headers to your request. For example, Azure's AI services require an `api-key` header containing an API key.
```

PR #3175, opened 2025-03-05
fix typo

tacman:
batch, not patch


guimachiavelli (approved):
Thanks for your contribution @tacman, it's much appreciated!

LGTM!

PR #3205, opened 2025-03-26
Update the env variable for the experimental logs mode

irevoire:
Someone noticed that the env variable was not working when updating the logs mode, and that’s the fix 🌟 

Context: https://github.com/meilisearch/meilisearch/issues/5459


guimachiavelli (approved):

guimachiavelli:
Thanks for the fix!

PR #3210, opened 2025-03-31
v1.14: Experimental feature: Composite embedders

### LINKED ISSUE #3188: v1.14: Experimental feature: Composite embedders
v1.14 introduces a new embedder `source`, `composite`. Composite embedders are built out of two embedders, `searchEmbedder` and `indexingEmbedder`, between which Meilisearch switches depending on the task at hand.

`searchEmbedder` and `indexingEmbedder` are mandatory when using `composite`. Each must be an embedder object following the same rules as any other embeddings, with the following exceptions:
- setting `binaryQuantized` and `distribution` is not allowed. Both settings can only be configured in the outer embedder
- `documentTemplate` and `documentTemplateMaxBytes`: invalid in `searchEmbedder`, mandatory in `indexingEmbedder`

## Tasks
- [ ] update `learn/resources/experimental_features_overview.mdx`
- [ ] update `reference/api/settings#embedders`

## References
- SME: @dureuill 
- PRD: https://meilisearch.notion.site/Composite-embedder-usage-14a4b06b651f81859dc3df21e8cd02a0
- engine issue: https://github.com/meilisearch/meilisearch/issues/5343

guimachiavelli:
Closes #3188

### DIFF
--- a/reference/api/settings.mdx
+++ b/reference/api/settings.mdx
@@ -2380,20 +2380,22 @@ The embedders object may contain up to 256 embedder objects. Each embedder objec
 
 These embedder objects may contain the following fields:
 
-| Name                   | Type             | Default Value                                                            | Description                                                                                                                                                    |
-|  :---------------------|  :---------------|  :-----------------------------------------------------------------------|  :-------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| **`source`**           | String           | Empty                                                                    | The third-party tool that will generate embeddings from documents. Must be `openAi`, `huggingFace`, `ollama`, `rest`, or `userProvided`                        |
-| **`url`**              | String           | `http://localhost:11434/api/embeddings`                                  | The URL Meilisearch contacts when querying the embedder                                                                                                            |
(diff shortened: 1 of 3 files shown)


reference/api/settings.mdx
```diff
@@ -2695,6 +2719,26 @@ This option can be useful when working with large Meilisearch projects. Consider
...
 
+##### `indexingEmbedder` and `searchEmbedder` <NoticeTag type="experimental" label="experimental" />
+
+When using a [composite embedder](#composite-embedders), configure separate embedders Meilisearch should use when vectorizing documents and search queries.
+
+`indexingEmbedder` often benefits from the higher bandwidth and speed of remote providers so it can vectorize large batches of documents quickly. `searchEmbedder` may often benefits from the lower latency of processing queries locally.
+
+Both fields must be an object and accept the same fields as a regular embedder, with the following exceptions:
+
+- `indexingEmbedder` and `searchEmbedder` must use the same `model`
```
dureuill, line 2730: this is not strictly true. At the abstract level both embedders should use the same model, but regarding the `model` parameter, if one sub-embedder is a `huggingFace` (local) embedder, and the other is a `rest` embedder  that calls a remote providers, `model` will not even be an available parameter for the REST embedder.

dureuill (approved):
Still :+1:

PR #3212, opened 2025-04-01
v1.14: New `pooling` parameter for Hugging Face embedders

### LINKED ISSUE #3189: v1.14: New `pooling` parameter for `huggingFace` embedders
Embedders with `source: huggingFace` now accept a new optional parameter, `pooling`.

Pooling refers to the last step in Hugging Face embedding generation, in which the multiple embeddings for individual tokens (e.g. words) are merged into a single embedding. Most models indicate which pooling method they expect.

`pooling` must be one of the following strings:
- `useModel`: fetch pooling method from model configuration, default for hugging face embedders created on >=v1.14
- `forceMean`: use mean pooling, default for embedders in <=v1.13
- `forceCls`: use CLS pooling

`pooling` is optional for hugging face embedders. `pooling` is invalid for all other embedder sources.

## Tasks
- [ ] Update `reference/api/settings#embedders`

## Reference
- SME: @dureuill 
- PRD: https://meilisearch.notion.site/Composite-embedder-usage-14a4b06b651f81859dc3df21e8cd02a0#19f4b06b651f80439972fbbdbc891764
- engine issue: https://github.com/meilisearch/meilisearch/issues/5343

guimachiavelli:
Closes #3189

### DIFF
--- a/reference/api/settings.mdx
+++ b/reference/api/settings.mdx
@@ -2394,6 +2394,7 @@ These embedder objects may contain the following fields:
 | **`request`**          | Object           | Empty                                                                    | A JSON value representing the request Meilisearch makes to the remote embedder                                                                                 |
 | **`response`**         | Object           | Empty                                                                    | A JSON value representing the request Meilisearch expects from the remote embedder                                                                             |
 | **`binaryQuantized`**         | Boolean           | Empty                                                                    | Once set to `true`, irreversibly converts all vector dimensions to 1-bit values                                                                             |
+| **`pooling`**         | String           | `"useModel"`                                                                    | Pooling method for Hugging Face embedders                                                                                                                                            |
 
 ### Get embedder settings
 
@@ -2457,7 +2458,8 @@ Partially update the embedder settings for an index. When this setting is update
     "request": { … },
     "response": { … },
     "headers": { … },
-    "binaryQuantized": <Boolean>
+    "binaryQuantized": <Boolean>,
+    "pooling": <String>
   }
 }
 ```
@@ -2542,7 +2544,6 @@ The maximum size of a rendered document template. Longer texts are truncated to 
 This field is incompatible with `userProvided` embedders.
 
 This field is optional for all other embedders.
-
 
 ##### `dimensions`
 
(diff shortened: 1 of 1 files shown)


reference/api/settings.mdx
```diff
@@ -2695,6 +2696,20 @@ This option can be useful when working with large Meilisearch projects. Consider
...
 
+##### `pooling`
+
+Configure how Meilisearch should merge individual tokens into a single embedding.
+
+`pooling` must be one of the following strings:
+
+- `"useModel"`: Meilisearch will fetch the pooling method from the model configuration, default value for new embedders
+- `"forceMean"`: always use mean pooling, default value for embedders created in Meilisearch \<=v1.13
+- `"forceCls"`: always use CLS pooling
```
dureuill, line 2707: not sure if we can find an authoritative explanation for CLS pooling to add as a link

dureuill (approved):
thank you

reference/api/settings.mdx
guimachiavelli: Yeah, I was looking for resources earlier, but most of what I could find was either not authoritative/trustworthy enough, or it was just a section in an article about something else (I even found a LinkedIn post about, I kid you not, BERT/CLS-related drama).

The only thing that could sort of fit is this article on NLP's BERT: https://www.geeksforgeeks.org/explanation-of-bert-model-nlp/, but even then I'm not super convinced.

dureuill: that's super sad :'( 

Without explaining what these means, maybe we could expand on the reason for this parameter to exist?

- In general, users want `useModel`
- The other options were introduced for compatibility with embedders created in Meilisearch <=v1.13 and for compatibility with some remote embedders. 
- The prime example is a composite embedder composed of a local embedder and of a remote embedder that complains about a similarity issue despite using the same model: try changing the `pooling` parameter of the local embedder to the other values and see if it fixes the issue.

Not sure how that fits in the

guimachiavelli: I added a short line explaining the expected usage of the `force` options.

The bit about composite embedders is interesting, but I prefer to have it in the reference for composite embedders.

dureuill (approved):

PR #3228, opened 2025-04-10
Fix localized-attributes samples

ellnix:


guimachiavelli (approved):
Brilliant, thanks once again, @ellnix!

PR #3282, opened 2025-06-25
Fix incorrect statement regarding apiKey and env vars for ollama

### LINKED ISSUE #3140: Documentation claims apiKey is inferred from `MEILI_OLLAMA_URL`
> [apiKey](https://www.meilisearch.com/docs/reference/api/settings#apikey)
> Authentication token Meilisearch should send with each request to the embedder.
> This field is mandatory if using a protected rest embedder.
> This field is optional for openAI and ollama embedders. If you don't specify apiKey, Meilisearch will attempt to read it from environment variables OPENAI_API_KEY **and MEILI_OLLAMA_URL**, respectively.
(emphasis mine)

I checked the source code, that env variable appears in
```rust
fn get_ollama_path() -> String {
    // Important: Hostname not enough, has to be entire path to embeddings endpoint
    std::env::var("MEILI_OLLAMA_URL").unwrap_or("http://localhost:11434/api/embeddings".to_string()) }
```

It only gets called once:
```rust
        let url = self.url.unwrap_or_else(get_ollama_path);
```

I cannot see how it interacts with the api key, if at all. This probably belongs in the documentation for the `url` parameter.

I don't use AI in anything, so I thought I'd post here just in case I'm misinterpreting something.

guimachiavelli:
Fixes #3140

### DIFF
--- a/reference/api/settings.mdx
+++ b/reference/api/settings.mdx
@@ -2635,7 +2635,7 @@ Authentication token Meilisearch should send with each request to the embedder.
 
 This field is mandatory if using a protected `rest` embedder.
 
-This field is optional for `openAI` and `ollama` embedders. If you don't specify `apiKey`, Meilisearch will attempt to read it from environment variables `OPENAI_API_KEY` and `MEILI_OLLAMA_URL`, respectively.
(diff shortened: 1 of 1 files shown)


dureuill (approved):
thank you :+1:

PR #3312, opened 2025-07-16
v1.16: Multimodal search

### LINKED ISSUE #3298: v1.16: Experimental: Multimodal search
v1.16 introduces a new experimental feature, `multimodal`. When activated, it enables two new embedder settings, `indexingFragments` and `searchFragments`, and one new search parameter, `media`. Used together, they make it possible to generate embeddings based on image instead of text. They also allow users to perform searches based on base64 images, image URLs, or text.

## Tasks
- [x] Update experimental feature overview table
- [x] Update error codes reference
- [x] Update Settings reference > Embedders with `indexingFragments` and `searchFragments`
- [x] Update Search reference with `media`

## References
- SME: @dureuill 
- Engine issue: https://github.com/meilisearch/meilisearch/issues/5501
- PRD: https://www.notion.so/meilisearch/Multimodal-search-usage-1c14b06b651f80c1bf9effe56dbeef54

guimachiavelli:
Fixes #3298

### DIFF
--- a/reference/api/settings.mdx
+++ b/reference/api/settings.mdx
@@ -3039,22 +3039,24 @@ To remove a single embedder, use the [update embedder settings endpoint](#update
 
 You can use the returned `taskUid` to get more details on [the status of the task](/reference/api/tasks#get-one-task).
 
-## Conversation <NoticeTag type="experimental" label="experimental" />
+## Chat <NoticeTag type="experimental" label="experimental" />
 
 <Note>
 This is an experimental feature. Use the Meilisearch Cloud UI or the experimental features endpoint to activate it:
 
 ```sh
 curl \
-  -X PATCH 'MEILISEARCH_URL/experimental-features/' \
+  -X PATCH 'http://localhost:7700/experimental-features/' \
+  -H 'Authorization: Bearer <master-key>' \
   -H 'Content-Type: application/json' \
   --data-binary '{
     "chatCompletions": true
   }'
 ```
+
 </Note>
 
-Conversational querying allows a more human-like interaction with your search engine. It enables you to ask questions and receive relevant answers based on the content of your documents.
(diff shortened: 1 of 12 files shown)


reference/api/settings.mdx
```diff
@@ -2875,6 +2877,118 @@ Both fields must be an object and accept the same fields as a regular embedder,
...
+  "FRAGMENT_NAME": {
+    "value": {
+      …
+    }
+  }
+```
+
+`FRAGMENT_NAME` can be any valid string. It must contain a single field, `value`. `value` must then follow your chosen model's specifications.
+
+For example, for Voyage AI's text and image embedder, `value` must be an array containing a single object with a `type` field. Depending on `type`'s value, you must include either `text`, `image_url`, or `image_base64`:
```
dureuill:
```suggestion
For example, for [VoyageAI's multimodal embedding route](https://docs.voyageai.com/reference/multimodal-embeddings-api), `value` must be an object containing a `content` field. `content` itself must contain an array of objects with a `type` field. Depending on `type`'s value, you must include either `text`, `image_url`, or `image_base64`:
```

snippets/samples/code_samples_add_or_replace_documents_1.mdx
```diff
@@ -80,7 +80,7 @@ documents := []map[string]interface{}{
     "release_date": "2019-03-23",
   },
 }
-client.Index("movies").AddDocuments(documents)
+client.Index("movies").AddDocuments(documents, nil)
```
dureuill: I'm not sure what this change is about 🤔

guimachiavelli:
Hmmm…that's not a bad idea, but after some thought I don't see a good way of integrating a multimodal section that doesn't make the current guide worse. at least not without rewriting significant chunks, at which point we're not really saving any time.

I spoke about this with Guillaume and I'll make the time to write a dedicated bit of content for this after the engine release, but before the marketing launch in September.

PR #3315, opened 2025-07-17
v1.16: `/export` route

### LINKED ISSUE #3302: v1.16: `/export` route
v1.16 introduces a new API route, `/export`. This allows users to migrate between Meilisearch instances without using dumps or snapshots.

## Tasks
- [ ] Create new API reference page, Export
- [ ] Update error code reference

## References

- SME: @Kerollmops 
- Engine issue: https://github.com/meilisearch/meilisearch/issues/5706
- PRD: https://www.notion.so/meilisearch/Export-Transfert-documents-between-instances-21e4b06b651f80e49005d61bd3a42761

guimachiavelli:
Fixes #3302

### DIFF
--- a/reference/api/export.mdx
+++ b/reference/api/export.mdx
@@ -15,6 +15,8 @@ Use the `/export` route to transfer data from your origin instance to a remote t
 
 Migrate data from the origin instance to a target instance.
 
+Migration is an additive operation. If the exported indexes already exist in the target instance, Meilisearch keeps the existing documents intact and adds the new data to the index.
+
 ### Body
 
 | Name                | Type   | Default value | Description                                                                                               |
@@ -32,7 +34,7 @@ This field is required.
 
 #### `apiKey`
 
-A security key with full admin permissions to a secured Meilisearch instance.
+A security key with `index.create`, `settings.update`, and `documents.add` permissions to a secured Meilisearch instance.
 
 #### `payloadSize`
 
@@ -42,10 +44,12 @@ The maximum size of each single data payload in a human-readable format such as 
 
 A set of objects whose keys correspond to patterns matching the indexes you want to export. By default, Meilisearch exports all documents across all indexes.
 
+Meilisearch does not override any index settings by default. If the target instance contains an index with the same name as an index you're exporting, Meilisearch uses the settings in the target instance. If the index does not already exist in the target instance, Meilisearch uses the settings in the origin instance.
+
 Each index object accepts the following fields:
 
 - `filter`: a [filter expression](/learn/filtering_and_sorting/filter_expression_reference) defining the subset of documents to export. Optional, defaults to `null`
-- `overrideSettings`: if `false`, uses the index settings as configured in the origin instance. if `true`, uses the target instance settings. Optional, defaults to `false`.
+- `overrideSettings`: if `true`, configures indexes in the target instance with the origin instance settings. Optional, defaults to `false`
 
 ### Example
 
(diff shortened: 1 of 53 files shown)


reference/api/export.mdx
```diff
@@ -0,0 +1,64 @@
+---
+title: Export
+description: "Migrate between instances with the `/export` route"
+---
+
+import { RouteHighlighter } from '/snippets/route_highlighter.mdx'
+
+import CodeSamplesExperimentalDeleteLogsStream1 from '/snippets/samples/code_samples_experimental_delete_logs_stream_1.mdx';
+
+Use the `/export` route to transfer data from your origin instace to a remote target instance. This is particularly useful when migrating from your local development environment to a Meilisearch Cloud instance.
```
Kerollmops:
```suggestion
Use the `/export` route to transfer data from your origin instance to a remote target instance. This is particularly useful when migrating from your local development environment to a Meilisearch Cloud instance.
```

reference/api/export.mdx
```diff
@@ -0,0 +1,64 @@
...
+
+#### `url`
+
+A string pointing to a remote Meilisearch instance, including its port if necessary.
+
+This field is required.
+
+#### `apiKey`
+
+A security key with full admin permissions to a secured Meilisearch instance.
```
Kerollmops: It requires the right to create indexes, update settings and upload documents, at least for now.

reference/api/export.mdx
```diff
@@ -0,0 +1,64 @@
...
+The maximum size of each single data payload in a human-readable format such as `"100MiB"`. Larger payloads are generally more efficient, but require significantly more powerful machines.
+
+#### `indexes`
+
+A set of objects whose keys correspond to patterns matching the indexes you want to export. By default, Meilisearch exports all documents across all indexes.
+
+Each index object accepts the following fields:
+
+- `filter`: a [filter expression](/learn/filtering_and_sorting/filter_expression_reference) defining the subset of documents to export. Optional, defaults to `null`
+- `overrideSettings`: if `false`, uses the index settings as configured in the origin instance. if `true`, uses the target instance settings. Optional, defaults to `false`.
```
Kerollmops:
```suggestion
- `overrideSettings`: if `false`, uses the index settings as configured in the target instance. if `true`, uses the origin instance settings. Optional, defaults to `false`.
```

Am I wrong, or is it a bad idea, and should it be the other way around? I mean in terms of API the default should be `true` 🤔

Kerollmops (changes requested):

reference/api/export.mdx
Kerollmops: I'll discuss with the team to see if the default shouldn't be `true` instead 👀

Kerollmops: Hey @guimachiavelli 👋
We found out with @Mubelotix that the parameter and default value are valid. The parameter defines that if the index already exists, we do not send the settings, and if it doesn't exist, we send them. If you set it to `true`, it will override the settings.

reference/api/export.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+Use the `/export` route to transfer data from your origin instance to a remote target instance. This is particularly useful when migrating from your local development environment to a Meilisearch Cloud instance.
+
+## Migrate database between instances
+
+<RouteHighlighter method="POST" path="/export" />
+
+Migrate data from the origin instance to a target instance.
+
+Migration is an additive operation. If the exported indexes already exist in the target instance, Meilisearch keeps the existing documents intact and adds the new data to the index.
```
Kerollmops:
```suggestion
Migration is an additive operation. If the exported indexes already exist in the target instance, Meilisearch keeps the existing documents intact and adds the new data to the index. If the same document is present in both the target and origin, Meilisearch replaces the target documents with the new data.
```

Kerollmops (changes requested):

PR #3322, opened 2025-07-23
v1.16: `rankingScoreThreshold` no longer affects `totalHits` accuracy

### LINKED ISSUE #3309: v1.16: `page` and `hitsPerPage` return correct values when used with `rankingScoreThreshold`
Prior to v1.16, when using the `rankingScoreThreshold`, Meilisearch stops calculating the total number of hits once it reaches the number of hits configured by `limit` and `hitsPerPage`, whether subsequent search results are above or below the score threshold. This causes the number of total hits and the facet distribution to always be incorrect, no matter if users are paginating results with `page`/`hitsPerPage` or `limit`/`offset`.

With v1.16, when processing search requests with `hitsPerPage`, Meilisearch computes whether all documents are below or above the ranking score threshold. This negatively impacts performance, but ensures `totalHits` is accurate when using `page`/`hitsPerPage`.

## Tasks
- [ ] Update `rankingScoreThreshold` warning box text

## References
- SME: @Mubelotix 
- Engine PR: https://github.com/meilisearch/meilisearch/pull/5725

guimachiavelli:
Fixes #3309

### DIFF
--- a/reference/api/search.mdx
+++ b/reference/api/search.mdx
@@ -1141,7 +1141,9 @@ Excludes results below the specified ranking score.
 Excluded results do not count towards `estimatedTotalHits`, `totalHits`, and facet distribution.
 
 <Warning>
-For performance reasons, if the number of documents above `rankingScoreThreshold` is higher than `limit`, Meilisearch does not evaluate the ranking score of the remaining documents. Results ranking below the threshold are not immediately removed from the set of candidates. In this case, Meilisearch may overestimate the count of `estimatedTotalHits`, `totalHits` and facet distribution.
+When using `rankingScoreThreshold`, queries paginating results with  `limit` and `offset` may exaggerate the count of `estimatedTotalHits`. For performance reasons, if the number of documents above `rankingScoreThreshold` is higher than `limit`, Meilisearch does not evaluate the ranking score of the remaining documents. Results ranking below the threshold are not immediately removed from the set of candidates.
+
+Queries using `page` and `hitsPerPage` are unnaffected by this issue and `totalHits` should always display the correct total number of search results.
 </Warning>
 
 #### Example
--- a/snippets/samples/code_samples_add_movies_json_1.mdx
+++ b/snippets/samples/code_samples_add_movies_json_1.mdx
@@ -59,7 +59,7 @@ file, _ := os.ReadFile("movies.json")
 var movies interface{}
 json.Unmarshal([]byte(file), &movies)
 
-client.Index("movies").AddDocuments(&movies)
+client.Index("movies").AddDocuments(&movies, nil)
 ```
 
 ```csharp C#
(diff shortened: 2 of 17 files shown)


reference/api/search.mdx
```diff
@@ -1141,7 +1141,9 @@ Excludes results below the specified ranking score.
 Excluded results do not count towards `estimatedTotalHits`, `totalHits`, and facet distribution.
 
 <Warning>
-For performance reasons, if the number of documents above `rankingScoreThreshold` is higher than `limit`, Meilisearch does not evaluate the ranking score of the remaining documents. Results ranking below the threshold are not immediately removed from the set of candidates. In this case, Meilisearch may overestimate the count of `estimatedTotalHits`, `totalHits` and facet distribution.
+When using `rankingScoreThreshold`, queries paginating results with  `limit` and `offset` may exaggerate the count of `estimatedTotalHits`. For performance reasons, if the number of documents above `rankingScoreThreshold` is higher than `limit`, Meilisearch does not evaluate the ranking score of the remaining documents. Results ranking below the threshold are not immediately removed from the set of candidates.
+
+Queries using `page` and `hitsPerPage` are unnaffected by this issue and `totalHits` should always display the correct total number of search results.
```
Mubelotix: Actually the whole bug is fixed and the warning can be removed. We could still warn users that the route may be slower than it was before.

```suggestion
Using `rankingScoreThreshold` with `limit` and `offset` forces Meilisearch to evaluate the ranking score of all matching documents to return an accurate `totalHits`, which can impact performance.

In contrast, queries using `page` and `hitsPerPage` avoid this overhead and may benefit from significantly faster response times.
```

guimachiavelli: Thanks for clarifying that bit. I updated your suggestion a bit, since, AFAIK, `limit`/`offset` searches always return `estimatedTotalHits`. 

I also avoided stating `page`/`hitsPerPage` can result in better performance because it sort of contradicts other parts of the docs where we state that `page`/`hitsPerPage` offers generally worse performance due to us having to calculate a precise `totalHits` value.

```suggestion
Using `rankingScoreThreshold` with `limit` and `offset` forces Meilisearch to evaluate the ranking score of all matching documents to return an accurate `estimatedTotalHits`. This may negatively impact search performance.

Queries with `page` and `hitsPerPage` avoid this overhead when using `rankingScoreThreshold`.
```

Mubelotix: You are right because I confused the two and it's actually the opposite. `page` and `hitsPerPage` are the ones forcing complete search. 

```suggestion
Using `rankingScoreThreshold` with `page` and `hitsPerPage` forces Meilisearch to evaluate the ranking score of all matching documents to return an accurate `totalHits`. This may negatively impact search performance.

Queries with `limit` and `offset` avoid this overhead when using `rankingScoreThreshold`.
```

guimachiavelli: Looks good to me, thanks!

PR #3333, opened 2025-08-05
Remove experimental tags from STARTS WITH

guimachiavelli:
## Reference

- linear: https://linear.app/meilisearch/issue/EXP-152/stabilize-starts-with
- engine: https://github.com/meilisearch/meilisearch/issues/5801


irevoire (approved):
perfecto 👌

PR #3343, opened 2025-08-19
Remove AWS and DO usage and installation

curquiza:
Because dropping of the service: https://github.com/meilisearch/cloud-providers/


tpayet (approved):
lgtm 👍🏽

PR #3355, opened 2025-09-11
Experimental feature: Vector store setting

guimachiavelli:
Fixes EXP-413

### DIFF
--- a/reference/api/settings.mdx
+++ b/reference/api/settings.mdx
@@ -130,7 +130,8 @@ By default, the settings object looks like this. All fields are modifiable.
   "prefixSearch": "indexingTime",
   "searchCutoffMs": null,
   "embedders": {},
-  "chat": {}
+  "chat": {},
+  "vectorStore": "stable"
 }
 ```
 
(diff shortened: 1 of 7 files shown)


reference/api/settings.mdx
```diff
@@ -3185,3 +3187,115 @@ curl \
 ```
 
 You can use the returned `taskUid` to get more details on [the status of the task](/reference/api/tasks#get-one-task).
+
+## Vector store <NoticeTag type="experimental" label="experimental" />
+
+<Note>
+This is an experimental feature. Use the Meilisearch Cloud UI or the experimental features endpoint to activate it:
```
dureuill: So, we won't show the feature in the UI for this week.

Can we update this to say that one should contact support on the Cloud to enable it?

```suggestion
This is an experimental feature. On Meilisearch Cloud, contact support to activate it. On OSS, use the experimental features endpoint:
```

or something better :-)

gmourier: I would also mention its accessible only on v1.21 version and later so interested people may be proactive in upgrading before asking support to do the upgrade for them

PR #3360, opened 2025-09-15
Add  `_geoPolygon` and `_geojson` documentation

guimachiavelli:
Fixes EXP-246

### DIFF
--- a/learn/filtering_and_sorting/geosearch.mdx
+++ b/learn/filtering_and_sorting/geosearch.mdx
@@ -176,7 +176,7 @@ Use the [`filter` search parameter](/reference/api/search#filter) along with `_g
 ### `_geoRadius`
 
 ```
-_geoRadius(lat, lng, distance_in_meters)
+_geoRadius(lat, lng, distance_in_meters, resolution)
 ```
 
 ### `_geoBoundingBox`
--- a/learn/filtering_and_sorting/geosearch.mdx
+++ b/learn/filtering_and_sorting/geosearch.mdx
@@ -21,9 +21,9 @@ Due to Meilisearch allowing malformed `_geo` fields in the following versions (v
 
 ## Preparing documents for location-based search
 
-In order to start filtering and sorting documents based on their geographic location, you must make sure they contain a valid `_geo` field.
+In order to start filtering and sorting documents based on their geographic location, you must make sure they contain a valid `_geo` or `_geojson` field.
 
-`_geo` is a reserved field. If you include it in your documents, Meilisearch expects its value to conform to a specific format.
+`_geo` and `_geojson` are reserved fields. If you include one of them in your documents, Meilisearch expects its value to conform to a specific format.
 
 When using JSON and NDJSON, `_geo` must contain an object with two keys: `lat` and `lng`. Both fields must contain either a floating point number or a string indicating, respectively, latitude and longitude:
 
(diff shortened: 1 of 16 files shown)


guimachiavelli:
I was honestly not aware of its existence and can't find any references to it on Notion/specs. Can you point me to a resource mentioning it?

learn/filtering_and_sorting/geosearch.mdx
```diff
@@ -21,9 +21,9 @@ Due to Meilisearch allowing malformed `_geo` fields in the following versions (v
 
 ## Preparing documents for location-based search
 
-In order to start filtering and sorting documents based on their geographic location, you must make sure they contain a valid `_geo` field.
+In order to start filtering and sorting documents based on their geographic location, you must make sure they contain a valid `_geo` or `_geojson` field.
```
irevoire: This sentence kinda makes it feel like we can sort with GeoJSON when we can't in reality. Not sure how we could reformulate 🤔

learn/filtering_and_sorting/geosearch.mdx
```diff
@@ -37,6 +37,37 @@ When using JSON and NDJSON, `_geo` must contain an object with two keys: `lat` a
...
+    "type": "Feature",
+    "geometry": {
+      "type": "Point",
+      "coordinates": [0.0, 0.0]
+    }
+  }
+}
+```
+
+Meilisearch does not support transmeridian shapes. If your document includes a transmeridian shape, split it into two separate shapes grouped as a `MultiPolygon` or `MultiLine`. Transmeridian shapes are polygons that cross the 180th or antimeridian.
```
irevoire:
```suggestion
Meilisearch does not support transmeridian shapes. If your document includes a transmeridian shape, split it into two separate shapes grouped as a `MultiPolygon` or `MultiLine`. Transmeridian shapes are polygons or lines that cross the 180th or antimeridian.
```

Don't know if it's interesting or not but we could also refer to the RFC that state you **SHOULD** split your shape:
> Any geometry that crosses the antimeridian SHOULD be represented by
   cutting it in two such that neither part's representation crosses the
   antimeridian.

[source](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.9)

Also, FYI and just in case you're not familiar with RFC: a **SHOULD** is not a **MUST**, you have the right to ignore it, it's just a recommendation:
> SHOULD   This word, or the adjective "RECOMMENDED", mean that there
   may exist valid reasons in particular circumstances to ignore a
   particular item, but the full implications must be understood and
   carefully weighed before choosing a different course.

[source](https://datatracker.ietf.org/doc/html/rfc2119#section-3)

So, here, meilisearch is being more strict than the RFC. In exchange, we gain two things:
1. It's way easier to code
2. People don't have to follow any winding order when sending polygon like they **SHOULD** in the RFC

learn/filtering_and_sorting/geosearch.mdx
```diff
@@ -37,6 +37,37 @@ When using JSON and NDJSON, `_geo` must contain an object with two keys: `lat` a
...
+
+Meilisearch does not support polygons with holes. If your polygon consists of an external ring and an inner empty space, Meilisearch ignores the hole and treats the polygon as a solid shape.
```
irevoire: Maybe we should put that in a warning or anything that stands out. What do you think? 🤔 
I'm afraid it could be a complex to understand source of bug 🤔

learn/filtering_and_sorting/geosearch.mdx
```diff
@@ -138,18 +171,24 @@ Meilisearch will rebuild your index whenever you update `filterableAttributes`.
...
 ### Usage
 
-Use the [`filter` search parameter](/reference/api/search#filter) along with `_geoRadius` or `_geoBoundingBox`. These are special filter rules that ensure Meilisearch only returns results located within a specific geographic area.
+Use the [`filter` search parameter](/reference/api/search#filter) along with `_geoRadius` and `_geoBoundingBox`. These are special filter rules that ensure Meilisearch only returns results located within a specific geographic area. If you are using GeoJSON for your documents, you may also filter results with `_geoPolygon`.
 
 ### `_geoRadius`
 
 ```
-_geoRadius(lat, lng, distance_in_meters)
+_geoRadius(lat, lng, distance_in_meters, resolution)
```
irevoire, line 179: The resolution is optional here, I don't know if we have a way to indicate that 🤔 
In command line application, they would put like that: `[resolution]`
But since we're already using brackets below, I don't think it's a good idea

guimachiavelli: We don't differentiate mandatory from optional parameters anywhere else in the docs, so there isn't a convention for this at the moment. Doesn't mean we can't create one, but I'd have to investigate alternatives—as you said, both curly and square brackets would create confusion due to them already being used elsewhere.

learn/filtering_and_sorting/geosearch.mdx
guimachiavelli: While it's good for us to know why we're doing this internally, I think describing differences between the RFC and our implementation is unnecessary as long as it's clear that in Meilisearch separating the transmeridian shape is mandatory.

PR #3413, opened 2025-11-27
Add further indexing optimization information

guimachiavelli:
Fixes EXP-709

### DIFF
--- a/snippets/samples/code_samples_add_or_update_documents_1.mdx
+++ b/snippets/samples/code_samples_add_or_update_documents_1.mdx
@@ -82,7 +82,8 @@ await index.UpdateDocumentsAsync(movie);
 #[derive(Serialize, Deserialize)]
 struct IncompleteMovie {
   id: usize,
-  title: String
+  title: String,
+  genres: String
 }
 
 let task: TaskInfo = client
@@ -90,7 +91,8 @@ let task: TaskInfo = client
   .add_or_update(&[
     IncompleteMovie {
       id: 287947,
-      title: "Shazam ⚡️".to_string()
+      title: "Shazam ⚡️".to_string(),
+      genres: "comedy".to_string()
     }
   ], None)
   .await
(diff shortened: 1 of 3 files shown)


learn/indexing/optimize_indexing_performance.mdx
```diff
@@ -80,7 +80,8 @@ Focus on the **longest-running steps** and investigate which index settings or d
 
 | Trace key | Description | Optimization |
 |------------|--------------|--------------|
-| `writing embeddings to database` | Time spent saving vector embeddings. | Use embedding vectors with fewer dimensions. <br/>- [Disabling embedding regeneration on document update](/reference/api/documents#vectors). <br/>- Consider enabling [binary quantization](/reference/api/settings#binaryquantized). |
+| `writing embeddings to database` | Time spent saving vector embeddings. | Use embedding vectors with fewer dimensions. <br/>- [Disabling embedding regeneration on document update](/reference/api/documents#vectors). <br/>- Consider enabling [binary quantization](/reference/api/settings#binaryquantized). <br/>- Consider enabling the [experimental vector store](/reference/api/settings#vector-store).|
```
ManyTheFish: I'd suggest the following order:
- enable new vector store
- enable binary quantization
- ~Disabling embedding regeneration on document update~: this will skip the `extracting embeddings` part, but we still need to write the vectors in the database. We should move it to the section below.



> FYI in v1.29.0: the new vector store will be enabled by default on any newly created index, old index will keep the legacy vector store (related to: https://github.com/meilisearch/meilisearch/pull/6004)

ManyTheFish (changes requested):
Small Changes :)

learn/indexing/optimize_indexing_performance.mdx
guimachiavelli: If the new vector store will be enabled by default in the next version, I think it may be best to remove the recommendation altogether. I updated the text based on your feedback, let me know if this looks good to you!

PR #3439, opened 2026-01-14
Clarify users cannot partially update object values

guimachiavelli:
Fixes EXP-819


bentxi13:
ty @guimachiavelli !! 🫶

PR #3659, opened 2026-08-18
Document federation.distinct in the federated search tutorial

qdequele:
## Description

v1.40 added `distinct` to the `federation` object of `/multi-search`, applying a global cross-index and cross-remote distinct computation. It was in the OpenAPI spec but had no prose, and the federated search tutorial documented only `weight`, `federation.page` and `federation.hitsPerPage`.

Added a "Deduplicate results across indexes" section that uses the tutorial's existing CRM datasets, so one client collapses to a single entry instead of appearing once per source. It includes the prerequisite step of making the attribute filterable in all three indexes, which is easy to miss and produces a confusing error otherwise.

Behavior is taken from the `Federation` schema in the spec plus the v1.40 changelog entry:

- applies across every query regardless of index or remote
- overrides the index-level distinct attribute
- setting `distinct` at both query and federation level returns HTTP 400
- facet distribution accuracy is not guaranteed once remotes are involved

The conclusion now mentions `distinct` alongside the other parameters the tutorial walks through.

## Checklist

- [x] ⚠️ I updated the code samples according to our [internal guidelines](https://www.notion.so/meilisearch/Updating-docs-for-engineers-3034b06b651f808da3c6c4f5e34914fc?source=copy_link#3034b06b651f8026bd63cfa294dfa0c6) (inline cURL, matching the surrounding examples on this page; happy to move these into `.code-samples.meilisearch.yaml` if you would prefer generated snippets here)

For external maintainers

### DIFF
--- a/capabilities/multi_search/getting_started/federated_search.mdx
+++ b/capabilities/multi_search/getting_started/federated_search.mdx
@@ -149,9 +149,60 @@ The response includes `page`, `hitsPerPage`, and `totalPages` instead of `offset
 
 This makes it straightforward to build paginated UIs that display merged results from multiple indexes.
 
+## Deduplicate results across indexes
+
+The same client may appear in a profile, a chat log, and a support ticket at once. Use the `distinct` property of the `federation` parameter to keep at most one result per value of a given attribute, computed across every query in the request.
+
+`distinct` requires the attribute to be filterable in each participating index, so add `client_name` to `filterableAttributes` for all three indexes first:
+
+<CodeGroup>
+
+```sh
+curl -X PUT 'MEILISEARCH_URL/indexes/profiles/settings/filterable-attributes' -H 'Content-Type: application/json' -H 'Authorization: Bearer MEILISEARCH_KEY' --data-binary '["client_name"]' &&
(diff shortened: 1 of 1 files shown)


qdequele:
@dureuill you added `distinct` to federated search in [meilisearch#6214](https://github.com/meilisearch/meilisearch/pull/6214), so tagging you for review.

Two claims worth your eye. I wrote that the attribute must be filterable in every participating index, and that setting `distinct` at both query and federation level returns HTTP 400. The first comes from the `Federation` schema wording ("can be set as distinct in all the participating indexes") plus the changelog; I want to be sure filterable is the precise requirement rather than something narrower.

CaroFG (approved):

dureuill:
the requirement is the same as regular `distinct`, so we could also write that: https://www.meilisearch.com/docs/capabilities/full_text_search/how_to/configure_distinct_attribute#setting-a-distinct-attribute-at-search-time

In the new filterableAttributes detailed syntax, it means that the attribute must have `features.filter.equality` OR `features.filter.comparison` set to `true`.

dureuill:
> and that setting distinct at both query and federation level returns HTTP 400

that is correct

PR #3662, opened 2026-08-18
Document wildcard patterns in the facets search parameter

qdequele:
## Description

v1.50 ([engine PR #6497](https://github.com/meilisearch/meilisearch/pull/6497)) extended the `facets` search parameter from the single `"*"` wildcard to any pattern containing `*`, matched with the same rules as `filterableAttributes.attributePatterns`. So `specs.*` requests a distribution for every filterable attribute under that prefix.

Neither the plain `"*"` form nor the new patterns appeared anywhere in the docs, so the only documented way to request facets was an explicit list.

Added a subsection to Step 2 of the faceted navigation guide, cross-linked to the existing wildcard patterns section for filterable attributes. It carries a warning about cost, since a wildcard can silently request many more distributions than the interface renders, and the neighbouring performance guide is explicit that each requested facet costs a computation.

Semantics were taken from the v1.50.0 release notes rather than the changelog entry, which only gives the `dogs.*` example. The OpenAPI spec has no description for `facets` on `SearchQuery`, so there was nothing to draw on there.

## Checklist

- [x] ⚠️ I updated the code samples according to our [internal guidelines](https://www.notion.so/meilisearch/Updating-docs-for-engineers-3034b06b651f808da3c6c4f5e34914fc?source=copy_link#3034b06b651f8026bd63cfa294dfa0c6) (illustrative JSON body, no new cURL sample)

For external maintainers

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for requesting multiple facet distributions using wildcard patterns.
  * Documented support for all filterable attributes and prefix-based attribute matching.
  * Added performance considerations and recommendations for choosing between wildcards and explicit facet lists.

### DIFF
--- a/capabilities/filtering_sorting_faceting/how_to/build_faceted_navigation.mdx
+++ b/capabilities/filtering_sorting_faceting/how_to/build_faceted_navigation.mdx
@@ -92,6 +92,23 @@ The response includes a `facetDistribution` object showing every value for each 
 
 The `facetDistribution` tells you exactly which values exist and how many documents match each one. The `facetStats` object provides minimum and maximum values for numeric facets, useful for building range sliders.
 
+### Request several facets with a wildcard
+
+`facets` also accepts wildcard patterns, which is convenient when related facets share a prefix and you do not want to list them one by one. Pass `"*"` to request every filterable attribute, or a pattern containing `*` to request the subset that matches it:
+
+```json
+{
+  "q": "classic",
+  "facets": ["specs.*"]
+}
+```
+
+Given filterable attributes `specs.language`, `specs.binding`, and `specs.edition`, that single entry returns a distribution for all three. Patterns follow the same matching rules as [`filterableAttributes.attributePatterns`](/capabilities/filtering_sorting_faceting/how_to/configure_granular_filters#wildcard-patterns), and only filterable attributes are ever matched.
+
+<Warning>
+A wildcard makes it easy to request far more facets than your interface displays, and every matched attribute costs a distribution computation. Prefer an explicit list for a page whose facets you know, and keep wildcards for cases where the attribute names are genuinely dynamic. See [Optimize facet performance](/capabilities/filtering_sorting_faceting/advanced/optimize_facet_performance) for the cost model.
+</Warning>
+
 <Note>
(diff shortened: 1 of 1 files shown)


qdequele:
@Kerollmops you added partial wildcard support for `facets` in [meilisearch#6497](https://github.com/meilisearch/meilisearch/pull/6497), so tagging you for review.

I documented both forms, since neither the plain `"*"` nor the new patterns appeared anywhere in the docs before. I described patterns as matching the same way as `filterableAttributes.attributePatterns` and as only ever matching filterable attributes, which is how your release note puts it. Please confirm the second half, that a pattern can never pull in a non-filterable field.

capabilities/filtering_sorting_faceting/how_to/build_faceted_navigation.mdx
```diff
@@ -92,6 +92,23 @@ The response includes a `facetDistribution` object showing every value for each
...
+`facets` also accepts wildcard patterns, which is convenient when related facets share a prefix and you do not want to list them one by one. Pass `"*"` to request every filterable attribute, or a pattern containing `*` to request the subset that matches it:
+
+```json
+{
+  "q": "classic",
+  "facets": ["specs.*"]
+}
+```
+
+Given filterable attributes `specs.language`, `specs.binding`, and `specs.edition`, that single entry returns a distribution for all three. Patterns follow the same matching rules as [`filterableAttributes.attributePatterns`](/capabilities/filtering_sorting_faceting/how_to/configure_granular_filters#wildcard-patterns), and only filterable attributes are ever matched.
```
CaroFG:
```suggestion
Given filterable attributes `specs.language`, `specs.binding`, and `specs.edition`, that single entry returns a distribution for all three. Only filterable attributes are ever matched.
```

Removing this because the page's "Wildcard patterns" section linked only demonstrates the bare "*" (match-everything) case for setting default features. It contains zero examples or stated rules for partial patterns like specs. or price*. It could be worth adding some info on that here or updating the other page

CaroFG (changes requested):

CaroFG (approved):

PR #3656, opened 2026-08-18
Document MEILI_EXPERIMENTAL_TASK_QUEUE_MAX_READERS from v1.53.1

qdequele:
## Description

v1.53.1 ([release notes](https://github.com/meilisearch/meilisearch/releases/tag/v1.53.1), [engine PR](https://github.com/meilisearch/meilisearch/pull/6587)) added `MEILI_EXPERIMENTAL_TASK_QUEUE_MAX_READERS`, which controls the maximum number of simultaneous LMDB read transactions on the task queue database.

#3649 covered v1.53.0 only, since it was opened on 2026-08-10 and v1.53.1 shipped on 2026-08-13, so this option was not documented anywhere.

**Changes**

- `changelog/changelog.mdx`: folded into the existing `v1.53` entry rather than adding a `v1.53.1` block, following the v1.43.1 precedent. The changelog has no patch-level `<Update>` blocks.
- `resources/self_hosting/configuration/reference.mdx`: new option, placed with the other task queue options, after "Maximum batch payload size".
- `resources/help/experimental_features_overview.mdx`: new table row.

**Default value** is `1024`, read from the engine source in PR #6587 rather than inferred. The variable is environment-only, with no CLI flag, and Meilisearch panics at startup on a non-integer value, which the reference entry notes.

`npm run check-broken-links` passes, including the new anchor link into the configuration reference.

## Checklist

For internal Meilisearch team member only:
- [x] I checked and followed our [internal guidelines](https://www.notion.so/meilisearch/Updating-docs-for-engineers-3034b06b651f808da3c6c4f5e34914fc?source=copy_link)
- [x] ⚠️ I updated the code samples according to our [internal guidelines](https://www.notion.so/meilisearch/Updating-docs-for-engineers-3034b06b651f808da3c6c4f5e34914fc?source=copy_link#3034b06b651f8026bd63cfa294dfa0c6) (no code sample changes needed)

For external maintainers
- [x] Did you use any AI tool while implementing this PR? Yes. Claude Code found the gap while auditing the docs against the engine changelog, read the default value from the engine source, and drafted these edits.
- [x] Have you made sure that the title is accurate and descriptive of the changes?

## Summary by CodeRabbit

* **New Features**
  * Added an experimental setting to control the maximum number of simultaneous task-queue database readers.
  * The setting defaults to `1024` and can be configured through the environment.
  * Invalid or non-positive values prevent startup.

* **Documentation**
  * Added configuration guidance and an entry to the experimental features overview.
[...]

PR #3663, opened 2026-08-18
Document the task and batch streaming routes

qdequele:
## Description

v1.52 added `GET /tasks/stream` and `GET /batches/stream` behind the `tasksStreamingRoute` experimental flag. Both were rendered in the API reference from the OpenAPI spec, but there was no prose explaining they exist, and `tasksStreamingRoute` was missing from the experimental features table, which otherwise lists every current flag.

Added a "Streaming task updates instead of polling" section to the task monitoring guide, since that page's whole premise is polling `GET /tasks/{task_uid}` and telling readers to "wait a few moments and query the database once again". It covers enabling the flag, holding the connection open, the difference between the two routes (batches also emit on progress, which suits long indexing runs), and that streams are live only, so a reconnecting client should resynchronise with a normal query.

Also added the corresponding experimental features table row.

Behavior is taken from the route descriptions in the spec. Note the spec is internally inconsistent about the response media type: the route descriptions say SSE while the 200 response declares `application/x-ndjson`. I deliberately described the behavior without asserting a `Content-Type`, but that looks worth fixing on the engine side.

## Checklist

- [x] ⚠️ I updated the code samples according to our [internal guidelines](https://www.notion.so/meilisearch/Updating-docs-for-engineers-3034b06b651f808da3c6c4f5e34914fc?source=copy_link#3034b06b651f8026bd63cfa294dfa0c6) (inline cURL; `.code-samples.meilisearch.yaml` already gained `get_tasks_stream_1` and `get_batches_stream_1` in #3649, so say the word if you would rather these prose examples used those snippets)

For external maintainers

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for receiving real-time task and batch status updates through experimental streaming endpoints.
  * Documented feature-flag requirements, usage examples, event behavior, reconnect handling, and the lack of historical event replay.
  * Added task and batch streams to the experimental features overview.


qdequele:
Heads up on merge order: this PR and #3656 both append a row to the table in `resources/help/experimental_features_overview.mdx`, at the same position. Each is mergeable against `main` on its own, but whichever lands second will report a one-line "both added" conflict there.

The resolution is to keep both rows, in either order. Nothing else in the two PRs overlaps.

They also both touch `resources/self_hosting/configuration/reference.mdx` (#3656 and #3664), but those insertions are far apart in the file and merge cleanly.

qdequele:
@Kerollmops you added the SSE routes in [meilisearch#6533](https://github.com/meilisearch/meilisearch/pull/6533), so tagging you for review.

Three things to check:

1. I wrote that neither route replays history, so a reconnecting client should resynchronise with a normal query. That is an inference from "sent any time their status changes" rather than something I verified.
2. I described `/batches/stream` as also emitting on progress changes, making it the better choice for following a long indexing run.
3. The spec is inconsistent about the media type: the route descriptions say SSE while the 200 response declares `application/x-ndjson`. I deliberately avoided stating a `Content-Type` in the prose, but that looks like an engine-side fix worth making.

Separately, for your awareness rather than this PR: the revert in v1.52.2 is not an ancestor of the v1.53 line, so these routes are live in v1.52.0/.1 and v1.53+ but absent from the latest v1.52 patches. Quentin confirmed those patches were test releases, so I have documented the routes as current.

capabilities/indexing/tasks_and_batches/monitor_tasks.mdx
```diff
@@ -95,6 +95,41 @@ When `status` changes to `succeeded`, Meilisearch has finished processing your r
...
+
+Both routes are experimental. Enable the `tasksStreamingRoute` flag before using them:
+
+<CodeGroup>
+
+```bash
+curl \
+  -X PATCH 'MEILISEARCH_URL/experimental-features' \
+  -H 'Content-Type: application/json' \
+  -H 'Authorization: Bearer MEILISEARCH_KEY' \
```
CaroFG:
```suggestion
  -H 'Authorization: Bearer API_KEY' \
```

CaroFG (approved):

PR #3665, opened 2026-08-18
Document the remote status field on the network route

qdequele:
## Description

v1.42 ([engine PR #6306](https://github.com/meilisearch/meilisearch/pull/6306)) added remote availability tracking and exposed a `status` field for each remote on `/network`. The replication guide already described the resulting fallback behavior but never mentioned the field, so there was no documented way for an operator to see which remotes Meilisearch currently considers reachable.

Added an "Inspect remote availability" subsection to the existing "Remote availability" section, covering the two values, the default, that Meilisearch maintains the field, and that reading `GET /network` distinguishes an unreachable instance from a shard misconfiguration.

## Worth flagging: the field is missing from the OpenAPI spec

Unlike most gaps I have been filling, this one is absent from the API reference too. The `Remote` schema in `assets/open-api/meilisearch-openapi.json` has only `url`, `searchApiKey` and `writeApiKey`.

The cause looks structural rather than accidental. In `crates/meilisearch-types/src/network.rs` at v1.53.1 the field is declared as:

```rust
#[serde(skip_deserializing)]
pub status: route::Status,
```

`Remote` serves as both the request and response shape, and `status` is response-only, so the generated schema documents the input and drops it. Documenting it as read-only in prose is correct either way, but the spec probably wants a fix so the API reference shows it. Happy to open an engine issue if that is useful.

## Checklist

- [x] ⚠️ I updated the code samples according to our [internal guidelines](https://www.notion.so/meilisearch/Updating-docs-for-engineers-3034b06b651f808da3c6c4f5e34914fc?source=copy_link#3034b06b651f8026bd63cfa294dfa0c6) (illustrative JSON response, no new cURL sample)

For external maintainers
- [x] Did you use any AI tool while implementing this PR? Yes. Claude Code found the gap auditing docs against the changelog, confirmed the field and its read-only semantics against the engine source at v1.53.1, and drafted the section.

### DIFF
--- a/resources/self_hosting/sharding/configure_replication.mdx
+++ b/resources/self_hosting/sharding/configure_replication.mdx
@@ -102,14 +102,14 @@ Meilisearch tracks which remotes are reachable and exposes that judgement on the
   "remotes": {
     "ms-01": {
       "url": "http://ms-01.example.com:7700",
-      "searchApiKey": "REPLACE_ME",
-      "writeApiKey": "REPLACE_ME",
+      "searchApiKey": "SEARCH_KEY_01",
+      "writeApiKey": "WRITE_KEY_01",
       "status": "available"
     },
     "ms-02": {
       "url": "http://ms-02.example.com:7700",
-      "searchApiKey": "REPLACE_ME",
-      "writeApiKey": "REPLACE_ME",
+      "searchApiKey": "SEARCH_KEY_02",
+      "writeApiKey": "WRITE_KEY_02",
       "status": "unavailable"
     }
   }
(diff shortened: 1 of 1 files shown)


qdequele:
@Kerollmops you added remote availability tracking in [meilisearch#6306](https://github.com/meilisearch/meilisearch/pull/6306), so tagging you for review.

I documented the `status` field as read-only with values `available` and `unavailable`, defaulting to `available`, based on `crates/meilisearch-types/src/network.rs` at v1.53.1 where it is declared `#[serde(skip_deserializing)]`.

Worth your attention: the field is missing from the `Remote` schema in the OpenAPI spec, so the API reference cannot show it. `Remote` is used for both request and response, and `status` is response-only, so the generated schema documents the input shape and drops it. Happy to open an engine issue if you would like.

I also noticed `StatusChangeForRemote` in the `NetworkChange` messages, which appears to let an operator change a remote's accessibility directly. That is out of scope here, but it is undocumented too. Tell me if it is worth a page.

qdequele:
Updated the key placeholders in the example. I used `SEARCH_KEY_01`/`WRITE_KEY_01` and `SEARCH_KEY_02`/`WRITE_KEY_02` rather than a single `API_KEY`, because this file already uses that per-remote pattern a few blocks further down (`SEARCH_KEY_03`/`WRITE_KEY_03`), and the example shows two remotes each with a distinct search and write key. Collapsing all four to `API_KEY` would hide that distinction in what is a response body.

Happy to switch to a plain `API_KEY` if you would rather be uniform with the cURL samples.

CaroFG (approved):

PR #3666, opened 2026-08-18
Document that searchCutoffMs bounds remote embedder calls

qdequele:
## Description

v1.44 made the search-time timeout for external REST embedders derive from the index's `searchCutoffMs` rather than a fixed value ([engine PR #6377](https://github.com/meilisearch/meilisearch/pull/6377)). It was listed as a breaking change, with the guidance that missing semantic results or HTTP 500s on pure semantic search after upgrading mean the cutoff needs raising.

This was undocumented in both directions: `configure_search_cutoff.mdx` never mentioned embedders, and no page under `capabilities/hybrid_search/` mentioned `searchCutoffMs`.

It matters because the cutoff page actively steers readers toward low values. It demonstrates a 150 ms cutoff and suggests starting between 100 ms and 500 ms, which is reasonable advice for lexical search and can silently break semantic search against a hosted provider.

Added a "Search cutoff and remote embedders" section covering that the cutoff must absorb the provider round trip, what failure looks like (degraded results for hybrid, HTTP 500 for pure semantic), the fix, and that local embedders are unaffected.

## Reviewer check requested

This is the one item from my audit where I would like an engineer to confirm the wording rather than take my reading of the changelog and release notes. Specifically:

1. Does the derived timeout apply to every embedder that makes a search-time HTTP call, or only to `source: rest`? I wrote "an embedder that calls an external API at search time" to stay accurate either way, but naming the sources precisely would be better.
2. Is HTTP 500 still the observed failure mode for pure semantic search on a timeout, or has that since become a more specific error?

Happy to tighten the section once you confirm.

## Checklist

- [x] ⚠️ I updated the code samples according to our [internal guidelines](https://www.notion.so/meilisearch/Updating-docs-for-engineers-3034b06b651f808da3c6c4f5e34914fc?source=copy_link#3034b06b651f8026bd63cfa294dfa0c6) (no code sample changes needed)

For external maintainers

### DIFF
--- a/capabilities/full_text_search/how_to/configure_search_cutoff.mdx
+++ b/capabilities/full_text_search/how_to/configure_search_cutoff.mdx
@@ -53,6 +53,21 @@ As a general recommendation, avoid setting the cutoff below **500ms**. This prov
 The cutoff is most useful as a safety net, not as a performance tuning knob. If your searches are consistently slow, address the root cause with the optimizations below rather than lowering the cutoff.
 </Tip>
 
+## Search cutoff and remote embedders
+
+If your index uses an embedder that calls an external API at search time, the cutoff also bounds that call. Meilisearch derives the embedding request timeout from `searchCutoffMs` rather than applying a fixed timeout of its own, so the cutoff has to cover the provider round trip on top of the search itself.
+
+This matters when choosing a value:
+
+- A cutoff that is comfortable for keyword search can be too short once a provider round trip is included. A 150 ms cutoff is fine for a purely lexical query and will usually not survive a call to a hosted embedding API.
+- If the embedder does not answer in time, the query loses its semantic results. For a [pure semantic search](/capabilities/hybrid_search/advanced/semantic_vs_hybrid), where there are no keyword results to fall back on, the request can fail with an HTTP 500 instead.
+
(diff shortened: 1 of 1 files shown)


qdequele:
@dureuill you made the embedder timeout follow `searchCutoffMs` in [meilisearch#6377](https://github.com/meilisearch/meilisearch/pull/6377), so tagging you for review.

This is the one PR from my audit I would not merge on my own wording. Two open questions, also in the PR body:

1. Does the derived timeout apply to every embedder that makes a search-time HTTP call, or only `source: rest`? I wrote "an embedder that calls an external API at search time" to stay accurate either way, but naming the sources precisely would be better.
2. Is HTTP 500 still the failure mode for pure semantic search on a timeout, or has that become a more specific error since v1.44?

Context for why I think this matters: the cutoff page demonstrates a 150 ms cutoff and suggests starting between 100 ms and 500 ms, which is sensible for lexical search and looks likely to break semantic search against a hosted provider.

CaroFG (approved):

dureuill:
> Is HTTP 500 still the failure mode for pure semantic search on a timeout, or has that become a more specific error since v1.44?

It looks that way. It kind of makes sense:

1. it is not a user error so not a HTTP 4xx
2. We could feasibly return a 503, but we have no guarantee that retrying later will improve the situation (case where the embedder config is wrong for example)

dureuill:
FYI default cutoff is 1500ms which is suitable for semantic search in a lot of documents

PR #3680, opened 2026-09-08
docs: clarify synonym term length limitation with warning and example table

### LINKED ISSUE #2676: Synonym limitation
Hello,
After speaking with @dureuill, we forgot to mention a limitation about synonyms.
The part [Maximum number of synonyms per term](https://www.meilisearch.com/docs/learn/configuration/synonyms#maximum-number-of-synonyms-per-term) is correct, and we don't want to change it.
However, there is a limitation on the `term` itself. Meilisearch only fetches the synonyms for the 1 to 3-word terms and not more.
This means that if you add a synonym `the lord of the Rings` -> `lotr`, you will never find a document containing `lotr` by searching `the lord of the Rings`. 
It only the case for the term but not for their synonyms, which means that if you add a synonym `lotr` -> `the lord of the Rings`, you will find a document containing `the lord of the Rings` by searching `lotr`.

We may add a section explaining this limitation.

Don't hesitate to ask more questions if you need them.

Abdellox:
## What does this PR do?

Improves the **Synonym term length limitation** section in \capabilities/full_text_search/relevancy/synonyms.mdx\.

The existing section described the 1–3 word limit correctly but did not make clear that the limitation is **asymmetric**: it applies to the search term only, not to the synonym value. This means that even with a mutual association configured, a 4+ word search term will never resolve to its short synonym — which is non-obvious and a common source of confusion.

## Changes

- Added a \<Warning>\ callout making explicit that the limit applies to the search term, not the synonym value
- Added a concrete mutual association example using \lotr\ / \lord of the rings\
- Added a comparison table showing which search direction works and why
- Added a recommendation for working around the limitation

## Related issue

Closes #2676

---

> ⚠️ AI disclosure: This contribution was assisted by an AI tool (Kiro). The content has been reviewed for accuracy against the existing Meilisearch behavior documented in the issue.

## Summary by CodeRabbit

- **Documentation**
  - Clarified that the 1–3 word limit applies to the user’s search term, not the synonym value.
  - Documented how short search terms can resolve to longer synonyms, while longer terms do not resolve to shorter ones.
  - Added a structured example showing synonym lookup behavior and recommending shorter search terms.

### DIFF
--- a/capabilities/full_text_search/relevancy/synonyms.mdx
+++ b/capabilities/full_text_search/relevancy/synonyms.mdx
@@ -179,7 +179,7 @@ However, depending on the ranking rules, they might be considered less [relevant
 Meilisearch only fetches synonyms for search terms that are between 1 and 3 words. Terms with 4 or more words will not trigger synonym lookup.
 
 <Warning>
-This limitation applies to the **search term** (what the user types), not to the synonym value. A short search term can still resolve to a long synonym, but a long search term will never resolve to a short synonym — even if both directions are configured.
+This limitation applies to the **search term** (what the user types), not to the synonym value. A short search term can still resolve to a long synonym, but a long search term will never resolve to a short synonym, even if both directions are configured.
 </Warning>
 
 ### Example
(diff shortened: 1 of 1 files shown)


Abdellox:
Addressed the CodeRabbit review feedback in 88b20fa:
- Replaced the em dashes with commas (warning on line 182 and the example table rows).
- Corrected the wording so synonym resolution is described as working only from the (shorter) search term, not in both directions.

